PR #5, opened 2020-01-15
Edit gitignore to exclude .idea/

Stallind:


Alumniminium:
Adding @Almantask to do the final merge

PR #28, opened 2020-01-17
homework1

kaiomv:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,59 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            int age;
```
Almantask: Declare variable as close as possible to where you use them.

kaiomv:
Variables now declared as closed as used by the first time, the second bmi calculation now uses the same variable as the first calculation

PR #39, opened 2020-01-18
Homework 1 and 2 put into csinn remote repository

calvareztx:
Asked user to input values for each variable, saved them as strings, then converted measurements. Created new variables to store converted values in order to determine BMI.
Wrote those converted values to output window concatenated with string.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,7 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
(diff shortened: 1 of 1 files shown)


calvareztx:
I apologize for leaving out that requirement. Changes have been made. Thank you all!

calvareztx:
I do see how my code is repetitive. Is there another way to structure this so I don't have two blocks of the same code that calculates a person's BMI?

ellmaster:
> I do see how my code is repetitive. Is there another way to structure this so I don't have two blocks of the same code that calculates a person's BMI?

Yeah sure. As that being part of the homework 3 you may want to think a bit more about how you could achieve that. :)

PR #109, opened 2020-01-26
Chapter1/exterminator

ghost:
A new PR where I solved hopefully the branch problem


ellmaster (approved):
Very well. Some points are not quite accurate but in summation this is very good work!

PR #110, opened 2020-01-26
Homework #5 from Aurora

Aurora2500:


Src/BootCamp.Chapter1/Program.cs
```diff
@@ -5,5 +5,16 @@ class Program
         static void Main(string[] args)
         {
         }
+
+        static void PrintArray(int[] array)
```
ellmaster: Take a look at string.Join :)

Almantask:
You don't have to make a new file for all those function implementations- just call it directly :)

Almantask:
Expired

PR #134, opened 2020-01-30
HW6

RoelofWobben:
Way to solved part1 

set a breakpoint on 19
Did  F5 and then step into till I see the value and then continue to enter the value and that for all 4.

For part2 I set a breakpoint on 49 and saw a text appear in my windows console. 
So I copy and paste that int the question and saw a poem.

I hope I did it alllright.


RoelofWobben:
yep, From every lesson I learn something

PR #138, opened 2020-01-31
Homework 6 - Debugging

chatieb:
Fix for the code is as follows:
Increased the Task Delay in the For Loop for "Aysnchromaguss.Mumble()" from 10 to 100.
A task delay of 10 from original code gives the incorrect passcode of "A maius minore a ad solis usque ab ortu disce ad occasum uno omnes".
The newly changed task delay of 100 gives the correct passcode of "A minore ad maius a solis ortu usque ad occasum ab uno disce omnes".
I got confirmatiion from @Antero360 in classroom Discord that the problem with the Task Delay was intended to be found for the homework, and that the problem was to be solved through trial and error (just as I had done by guessing). There does not seem to be a way do debug for this problem.

For Part 1, the return values are as follows:
- First return value is "Debugger!"
- Second return value is "51"
- Third return value is "79"
- Last return value is "77"

For Part 2, the passcode is as follows:
- "A minore ad maius a solis ortu usque ad occasum ab uno disce omnes"

Result is as follows:
  YOU MAY PROCEED!
  -------------------------------
  From the smaller to the greater,
  From sunrise to sunset,
  From one, learn all.
  -------------------------------

Side note: Thank you for teaching us.  There is so much to learn from all of you.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -38,7 +38,7 @@ namespace BootCamp.Chapter
             for (var index = 0; index < 14; index++)
             {
                 Aysnchromaguss.Mumble();
-                await Task.Delay(10);
+                await Task.Delay(100);
             }
 
             await Task.Delay(1000);


ellmaster:
Obviously done.
@Almantask 
I close so the solution is a bit more "hidden" (though I could edit it out.)

@chatieb 
You can reopen if you take the challenge and go for optional part of hw 6. 👍 
Good work!

Almantask:
Yup. this is totally good. Thanks!

PR #143, opened 2020-02-01
Hard reset changes during lesson and added code from homework 1

MrPepino:
Homework 1 put to remote repository

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,7 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
+            // input person 1
             Console.Write("Enter a first name: ");
             string firstName = Console.ReadLine();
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,38 @@
 ﻿using System;
+using System.Globalization;
 
 namespace BootCamp.Chapter
 {
     class Program
     {
         static void Main(string[] args)
         {
```
Almantask: You need to check it for 2 people, not just one.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,38 @@
...
-            Console.WriteLine("Hello World!");
+            string firstName;
```
Almantask: Try to declare variables as close to their usage as possible.

PR #167, opened 2020-02-12
Homework 9:

chatieb:
Part 1
Make a program which has 2 conversion modes:
From decimal to binary
From binary to decimal

Part 2
Use your WASD (both capital and non capital) keys to print arrows
W- ↥
A- ↤
S- ↧
D- ↦

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -1,12 +1,209 @@
 ﻿using System;
+using System.Text;
 
 namespace BootCamp.Chapter
 {
     class Program
     {
+        private const string LeftArrow = "\u2190";
+        private const string UpwardArrow = "\u2191";
+        private const string RightArrow = "\u2192";
+        private const string DownArrow = "\u2193";
+
+        private const char LeftKey = 'A';
+        private const char UpwardKey = 'W';
+        private const char RightKey = 'D';
+        private const char DownKey = 'S';
+
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            //Part 1 - Convert with 2 conversion modes: From decimal to binary, From binary to decimal.
+            ConvertNumberSystem();
+
+            //Part 2 - Use your WASD (both capital and non capital) keys to print arrows.
+            ConvertLettersToArrows();
+        }
+
+        private static void ConvertNumberSystem()
+        {
+            var mode = GetConversionMode();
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,209 @@
...
+
+        private const char LeftKey = 'A';
+        private const char UpwardKey = 'W';
+        private const char RightKey = 'D';
+        private const char DownKey = 'S';
+
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            //Part 1 - Convert with 2 conversion modes: From decimal to binary, From binary to decimal.
```
Almantask: This could be moved as XML docs of method.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,209 @@
...
+            ConvertNumberSystem();
+
+            //Part 2 - Use your WASD (both capital and non capital) keys to print arrows.
```
Almantask: This could be moved as XML docs of method.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,209 @@
...
+            ConvertLettersToArrows();
+        }
+
+        private static void ConvertNumberSystem()
+        {
+            var mode = GetConversionMode();
+            var number = GetNumber(mode);
+            var convertedNumber = (mode == 1) ? ConvertDecimalToBinary(int.Parse(number)) : ConvertBinaryToDecimal(number).ToString(); ;
```
ellmaster: 1 serves a bit as a magic number here. We will tackle this later on but you could have resolved that by using another constant variable. This is a minor issue though. :)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,209 @@
...
+            var mode = 0;
+            do
+            {
+                Console.WriteLine("Choose 1 or 2:");
+                isValidMode = int.TryParse(Console.ReadLine(), out mode);
+
+                if (!isValidMode && (mode != 1 && mode != 2))
+                {
+                    isValidMode = false;
+                }
```
ellmaster: Here we can clearly see what happens if you choose to ever use other methods/values. That will need a lot of changes. That is why I did point out using constants (for now) could be helpful.

PR #193, opened 2020-02-24
Chapter 2 Homework 2 From Aurora

Aurora2500:


Src/BootCamp.Chapter/Armor.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            return _defence;
+        }
+
+        public Armor(string name, decimal price, float weight, float defence):base(name, price, weight)
+        {
+            _defence = defence;
+        }
+
+        public static Armor NullArmor()
```
ellmaster: This feels really bad. Null should be null and nothing else. Don't do this. If you think you need this you probably have design issues.

Src/BootCamp.Chapter/AsymetricalArmor.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public class AsymetricalArmor : Armor
```
ellmaster: While being possible in the current scope this still is nothing I would recommend. You do not need the extra logic at all.

Aurora2500: The same logic used here would just be repeated in Armpiece and Shoulder piece class. There is no extra logic being added, just making a base class instead of having the same code copied all over the solution.

ellmaster: But you do not use it once. `GetIsLeft()` was never used. Your justification would be fine if you make use of your logic you coded in, but you didn't. Because of that you made something more complex than needed. Also you did not make real use of any constructor that needs isLeft. Actually it doesn't matter if you set the "same" Armor to the left and the right or two different ones.

Aurora2500: I might have misunderstood the task. Do Armpiece and Shoulderpiece not have a specified side, and instead the player decides if to use one of those on the left or the right?

Src/BootCamp.Chapter/Armor.cs
Aurora2500: I removed NullArmor and NullItem

Src/BootCamp.Chapter/AsymetricalArmor.cs
Aurora2500: Well I removed them now, gonna leave the conversation open for now just in case.

ellmaster: This is rather a UI-interaction problem than a problem that should be solved through inheritance IMO. It could be possible to solve that through it, though. However, your code did lack the additional logic needed to justify the left and right side differenciation. That is the main point. Your old version didn't fail on trying to put a right side item on the left for example. :)

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -70,15 +73,29 @@ public float GetTotalWeight()
...
+
+            foreach (var item in items)
+            {
+                if (item == null)
+                {
+                    currentWeight += 0f;
+                    continue;
+                }
+                currentWeight += item.GetWeight();
+            }
```
ellmaster: In my opinion this is much more clean now!

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -15,12 +15,29 @@ public Inventory()
...
+            Item[] wantedItems = new Item[0];
+            for (int i = 0; i < _items.Length; i++)
+            {
+                Item currentItem = _items[i];
+                if (currentItem.GetName() == name)
+                {
+                    wantedItems = Item.Append(wantedItems, currentItem);
+                }
+            }
+            return wantedItems;
```
ellmaster: Highly inefficient!!

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+        {
+            var newArray = new Item[1 + array.Length];
+            for (int i = 0; i < array.Length; i++)
+            {
+                newArray[i] = array[i];
+            }
+            newArray[^1] = toAppend;
+            return newArray;
+
+        }
```
ellmaster: You should remove this completely. That's not of a good use case here. Also - while not being specified - the inventory should be of fixed size. I asked Kai for clarification and he told so.

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+
+        public static bool ItemInArray(Item[] array, Item searchItem)
+        {
+            bool itemInArray = false;
+            foreach (Item item in array)
+            {
+                if (item == searchItem)
+                {
+                    itemInArray = true;
```
ellmaster: You can return true here.

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+            {
+                if (array[i] == toRemove)
+                {
+                    itemRemoved = true;
+                }
+
+                newArray[i] = itemRemoved ? array[i + 1] : array[i];
+            }
+            return newArray;
+        }
```
ellmaster: Inefficient, but encapsulated. Thought you should think about making it better!

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+
+        public static bool NameInArray(Item[] array, string name)
+        {
+            bool inArray = false;
+            foreach (Item item in array)
+            {
+                if (item.GetName() == name)
+                {
+                    inArray = true;
```
ellmaster: You can return true here.

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                if (array[i].GetName() == name)
+                {
+                    itemRemoved = true;
+                }
+                newArray[i] = itemRemoved ? array[i] : array[i + 1];
+            }
+            return newArray;
+        }
```
ellmaster: Also pretty inefficient!

Src/BootCamp.Chapter/Inventory.cs
Almantask: Yes, because internally Append recreates arrays, so you create arrays on every iteration.
However at this stage I don't know if you can do it any better. So I guess it's fine @ellmaster what did you have in mind?

Src/BootCamp.Chapter/Item.cs
```diff
@@ -15,12 +15,98 @@ public decimal GetPrice()
...
+                {
+                    itemRemoved = true;
+                }
+
+                newArray[i] = itemRemoved ? array[i + 1] : array[i];
+            }
+            return newArray;
+        }
+
+        public static bool NameInArray(Item[] array, string name)
```
Almantask: By convention, bool methods/variables tend to start with Is. There are exceptions, but for general methods like this one it's better to stick to conventions.

Src/BootCamp.Chapter/Inventory.cs
Almantask: Oh I think I see what you meant. indeed, there is something you can do to make this more efficient with the given knowledge.

Src/BootCamp.Chapter/Item.cs
Aurora2500: What size is that?

ellmaster: That is up to you I guess. Think about a valid size for an inventory. I chose 100 in my implementation but that might be a bit big.

Src/BootCamp.Chapter/Inventory.cs
ellmaster: Yeah I wanted to point out that Item.Append is really bad performance-wise. There are multiple ways to tackle this with the current knowledge IMO @Almantask :)

PR #195, opened 2020-02-25
Chapter2/homework/2

RoelofWobben:
Everyting solved except the extra challenge. I have to think how I can know how to exclude the weapon from the GetTotalDefense() method and only counts the weapons on the GetTotalDefense() method.BUt I made a PR because that part is a extra challenge

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -1,6 +1,4 @@
-﻿using System;
-
-namespace BootCamp.Chapter
+﻿namespace BootCamp.Chapter
 {
     public class Inventory
     {
@@ -23,7 +21,7 @@ namespace BootCamp.Chapter
             {
                 if (item.GetName() == name)
                 {
-                    foundItems = item.AddItem(foundItems,item); 
+                    foundItems = Add(item);
                 }
             }
 
@@ -32,28 +30,62 @@ namespace BootCamp.Chapter
 
         public void AddItem(Item item)
         {
-            _items = item.AddItem(_items, item);
+            _items = Add(item);
         }
 
-        internal bool SearchItem(string name)
+        public Item[] Add(Item item)
+        {
+            var newArray = new Item[_items.Length + 1];
+            for (int i = 0; i <= _items.Length - 1; i++)
+            {
+                newArray[i] = _items[i];
+            }
+            newArray[^1] = item;
+            return newArray;
+        }
+
+        public bool SearchItem(string name)
         {
             foreach (var item in _items)
             {
                 if (item.GetName() == name)
                 {
-                    return true; 
+                    return true;
                 }
             }
 
-            return false; 
+            return false;
         }
 
-        internal void RemoveByName(string name)
+        public void RemoveByName(string name)
         {
-            _items = Item.RemoveByBame(_items, name); 
+            _items = RemoveByBame(name);
         }
 
+        public Item[] RemoveByBame(string name)
+        {
+            var newArray = new Item[_items.Length + 1];
+            var found = false;
+            for (int i = 0; i < _items.Length; i++)
+            {
+                if (_items[i].GetName() == name)
+                {
+                    found = true;
+                    continue;
+                }
 
+                if (found)
+                {
+                    newArray[i - 1] = _items[i];
+                }
+                else
+                {
+                    newArray[i] = _items[i];
+                }
+            }
+
+            return newArray;
+        }
 
         /// <summary>
         /// Removes item matching criteria by item.
@@ -61,9 +93,37 @@ namespace BootCamp.Chapter
         /// </summary>
         public void RemoveItem(Item item)
         {
-            _items = Item.RemoveFromArray(_items, item);
+            _items = RemoveFromArray(item);
         }
 
-       
+        public Item[] RemoveFromArray(Item item)
+        {
+            if (_items.Length == 0)
+            {
+                return _items;
+            }
+
+            var newArray = new Item[_items.Length - 1];
+            var found = false;
+            for (int i = 0; i < _items.Length - 1; i++)
+            {
+                if (_items[i] == item)
+                {
+                    found = true;
+                    continue;
+                }
+
+                if (found)
+                {
+                    newArray[i - 1] = _items[i];
+                }
+                else
+                {
+                    newArray[i] = _items[i];
+                }
+            }
+
+            return newArray;
+        }
     }
 }
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -18,12 +18,12 @@ namespace BootCamp.Chapter
 
         public Item[] GetItems(string name)
         {
-            var foundItems = new Item[1];
+            var foundItems = new Item[0];
             foreach (var item in _items)
             {
                 if (item.GetName() == name)
                 {
-                    foundItems[0] = item;
+                    foundItems = item.AddItem(foundItems,item); 
                 }
             }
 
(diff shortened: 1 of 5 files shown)


RoelofWobben:
I could make a class Armour and a class Weapon to add things to the Equiment class but for the Armour I do not see a way to know what armour it is so which setxxx method to use

ellmaster:
> I could make a class Armour and a class Weapon to add things to the Equiment class but for the Armour I do not see a way to know what armour it is so which setxxx method to use

I do not really understand this remark.

RoelofWobben:
Oke, 

Lets say I make a class Armour. 
and make Glove and a Armpiece a subclass of Armour
Now I can make a function to add a Armour to the Equipment 
but as far as I know which Item Im adding a glove or a armpiece. 
so I cannot call or SetGlove or the SetArmpiece to add them to the right part.
That is what I try to say
but maybe again a extra challenge is a step to far for me

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -29,7 +65,31 @@ public void AddItem(Item item)
...
+            {
+                if (i < indexItem)
+                {
+                    newItemArray[i] = _items[i];
+                }
+                else if (i > indexItem)
+                {
+                    newItemArray[i - 1] = _items[i];
+                }
+            }
```
ellmaster: Using Array.IndexOf() here is waste of performance. You can do that in your own for loop.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -12,12 +12,13 @@ public decimal GetMoney()
 
         public Shop()
         {
-
+
```
ellmaster: This should also set _inventory so it isn't null anymore.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -15,12 +18,45 @@ public Inventory()
...
+            foreach (var item in _items)
+            {
+                if (item.GetName() == name)
+                {
+                    foundItems[0] = item;
+                }
+            }
+
+            return foundItems;
         }
```
ellmaster: This should return all instances, not only the last one found. :)

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -29,7 +65,31 @@ public void AddItem(Item item)
         /// </summary>
         public void RemoveItem(Item item)
         {
+            _items = RemoveFromArray(item);
+        }
+
+        private Item[] RemoveFromArray(Item item)
```
ellmaster: Your array operations look neat when looking at them first but I think that they not only add unneccessary complexity but also do more work than you need to do.
I do not see a real benefit here. This is true for all other ones, too, if I am not mistaken.
Also I think - while your basic idea is good - that the results differ from what you were planning to achieve.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -15,12 +18,45 @@ public Inventory()
...
+        {
+            foreach (var item in _items)
+            {
+                if (item.GetName() == name)
+                {
+                    return item; 
+                }
+            }
+
+            return new Item("", 0.00M, 00F);
```
ellmaster: This looks evil and error prone to me.

ellmaster:
> Oke,
> 
> Lets say I make a class Armour.
> and make Glove and a Armpiece a subclass of Armour
> Now I can make a function to add a Armour to the Equipment
> but as far as I know which Item Im adding a glove or a armpiece.
> so I cannot call or SetGlove or the SetArmpiece to add them to the right part.
> That is what I try to say
> but maybe again a extra challenge is a step to far for me

This is true as far as I can think of what you are able to achieve in the current scope, yet.
Though, even with your still very limited scope there are ways to provide a more suitable functionality than what you have, yet.
**NOTE!!!!**
I do not request this change because it will probably lead into you trying "stupid" stuff just to make it work. You will most likely try something very unusual and complicated whereas the solution I have in mind is very easy and not hard to achieve. I have my solution coded and I will PR it at some point in the future. :)

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: solved

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: I have done that because we learned we cannot add remove items from a array. So we need to make a new one. Copy alll the data we want and add or skip the item we want to add or remove.
but because remove is void and  I need to return the new array I need to make another method/function that does that.  
That was my plan and I do not see how it differ from what I was planning to achieve so if you can telll me  where im wrong im happy to change

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: oke, here I have to think about because right now item[] does not have a method/function to add things and in my oponion there is also no logical place to place such a function.

Src/BootCamp.Chapter/Shop.cs
RoelofWobben: solved

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: why. 
It iterates through the collection and if it's find it  , it returns the object otherwise I return a sort of empty item because I need to return a Item

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: I think I found a solution for this.

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: or do you mean I can make one function for adding and one for removing and re-use that for the shop and the player ?

Src/BootCamp.Chapter/Inventory.cs
RoelofWobben: Is there then a better way to remove a item if iI only know the name.
I only can find the remove method in the array class but that one needs a item and not a name,
So can you give me  some hints to get out of this.  Did the whole morning a lot of googling and trying but not nearby a better solution then I use now.

RoelofWobben: I could do something like : 
``` 
for (var i = 0;  i > collection.Length -1; i ++) 
{
   if (collection[i].GetName() == name 
   {
   }
```  

but how do I know if I can copy the item at the same index or do I have to copy it a index before it.

RoelofWobben: Nope, cannot be working  Inventory is not a string[] so it does not have a Length property.

RoelofWobben:
Hello, 

I hope I did all the changes you requested otherwise I hope you or Ki will explain why my code is complex and do more then expected

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -15,21 +18,52 @@ public Inventory()
...
+                    foundItems = item.AddItem(foundItems,item); 
+                }
+            }
+
+            return foundItems;
         }
 
         public void AddItem(Item item)
         {
+            _items = item.AddItem(_items, item);
```
Almantask: You add item to item? Something isn't right, please move the function outside of item.

RoelofWobben:
Can this one stil be open.  As soon as I solved HW1 I will make the changes here.
I do not like to work with 2 challenges at the same time.

Src/BootCamp.Chapter/Item.cs
```diff
@@ -22,5 +24,11 @@ public Item(string name, decimal price, float weight)
             _price = price;
             _weight = weight;
         }
+
```
Almantask: Remove thosee blank lines

RoelofWobben: solved

Almantask:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- Src/BootCamp.Chapter/Shop.cs  1
- Src/BootCamp.Chapter/Inventory.cs  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Almantask/CSharp-From-Zero-To-Hero/pullRequest?prid=5068129&bid=16675405)

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -15,12 +16,75 @@ public Inventory()
...
+
+            return false;
+        }
+
+        public void RemoveByName(string name)
+        {
+            _items = RemoveByBame(name);
+        }
+
+        public Item[] RemoveByBame(string name)
```
ellmaster: :D

RoelofWobben: sorry, I do not understand what you mean with the emonicon.
But the homework is approved so im happy

ellmaster: Remove by bame is a funny typo

RoelofWobben: oke, never noticed that one

PR #223, opened 2020-03-14
Max Chapter3/homework/1

HellHunterMax:
Finished main task. might do Extra when i get the time. (made a start though).

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -7,59 +7,15 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
-        private Weapon _weapon;
-        public void SetWeapon(Weapon weapon)
-        {
-
-        }
-
-        private Headpiece _head;
-        public void SetHead(Headpiece head)
-        {
-
-        }
-
-        private Chestpiece _chest;
-        public void SetChest(Chestpiece chestpiece)
-        {
-
-        }
-
-        private Shoulderpiece _leftShoulder;
-        public void SetLeftShoulder(Shoulderpiece should)
-        {
-
-        }
-
-        private Shoulderpiece _rightShoulder;
-        public void SetRightShoulder(Shoulderpiece shoulder)
-        {
-
-        }
-
-        private Legspiece _legs;
-        public void SetLeg(Legspiece legs)
-        {
-
-        }
-
-        private Armpiece _leftArm;
-        public void SetLeftArmp(Armpiece arm)
-        {
-
-        }
-
-        private Armpiece _rightArm;
-        public void SetRightArm(Armpiece arm)
-        {
-
-        }
-
-        private Gloves _gloves;
-        public void SetGloves(Gloves gloves)
-        {
-
-        }
+        public Weapon Weapon { get; set; }
+        public Headpiece Head { get; set; }
+        public Chestpiece Chest { get; set; }
+        public Shoulderpiece LeftShoulder { get; set; }
+        public Shoulderpiece RightShoulder { get; set; }
+        public Legspiece Legs { get; set; }
+        public Armpiece LeftArm { get; set; }
+        public Armpiece RightArm { get; set; }
+        public Gloves Gloves { get; set; }
 
         /// <summary>
         /// Gets total weight of armour.
@@ -67,6 +23,7 @@ namespace BootCamp.Chapter
         /// <returns></returns>
         public float GetTotalWeight()
         {
+
             return 0;
         }
 
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -23,7 +23,6 @@ namespace BootCamp.Chapter
         /// <returns></returns>
         public float GetTotalWeight()
         {
-
             return 0;
         }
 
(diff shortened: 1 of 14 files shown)


Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,59 @@
...
+            {
+                if (item.Name == name)
+                {
+                    return item;
+                }
+            }
+            return null;
         }
 
         public Item[] GetItems(string name)
```
Matt500b: The main goal of this homework is to refactor out the arrays and replace with Lists. Will need to change the return type and amend the code inside :)

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -29,7 +62,14 @@ public void AddItem(Item item)
         /// </summary>
         public void RemoveItem(Item item)
```
Matt500b: Since you already have the item object, you can call Items.Remove(item) rather than looping to find based on name. Let the list handle that for you :)

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,59 @@
...
+            Items = new List<Item>();
         }
 
-        public Inventory()
+        /// <summary>
+        /// Returns the item it found based on the name you gave, if no item is found returns null
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public Item Contains(string name)
```
Matt500b: Just from a naming perspective this should probably be GetItem, the singular version of your `GetItems()`.

Most implementations of `Contains` usually returns a bool to say "yes this collection contains this item/No this collection does not contain this item".

More of an FYI than a "you must fix" comment :)

Src/BootCamp.Chapter/Player.cs
```diff
@@ -12,50 +12,35 @@ namespace BootCamp.Chapter
...
-
         public Player()
         {
+            _inventory = new Inventory();
         }
 
         /// <summary>
         /// Gets all items from player's inventory
         /// </summary>
         public Item[] GetItems()
```
Matt500b: Similar comment; refactor away the arrays to use Lists. Then you wont need the `.ToArray()`

Src/BootCamp.Chapter/Player.cs
```diff
@@ -73,34 +58,67 @@ public Item[] GetItems(string name)
...
+        /// <summary>
+        /// Each point of strength allows extra 10 kg to carry.
+        /// </summary>
+        private int Strenght { get; set; }
+
+
+        /// <summary>
+        /// Needed only for the extra task.
+        /// </summary>
+        private Equipment Equipment { get; set; }
```
Matt500b: I'm not sure if you're 100% trying to complete the challenge or not since you've not implemented the attack/weight checks however Equipment is never initialised so trying to call .Equip will raise a `NullReferenceException`

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -2,27 +2,20 @@
...
-        }
 
         public Shop(decimal money)
         {
-            _money = money;
+            Money = money;
+            _inventory = new Inventory();
         }
 
         public Item[] GetItems()
```
Matt500b: Similar, refactor out arrays

Src/BootCamp.Chapter/Inventory.cs
HellHunterMax: That would result in Removing all objects that are sold even if i only want to sell 1?

Src/BootCamp.Chapter/Player.cs
HellHunterMax: No not going for the challenge just yet. i just made a start. To much work to do.

Src/BootCamp.Chapter/Inventory.cs
Matt500b: the `.Remove(T item)` method will only remove the first found instance. if there are duplicates it will not.

however, on a deeper level we should never really have duplicated objects. We might have "two swords" but they are not the _same_ sword.

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,59 +7,15 @@ namespace BootCamp.Chapter
...
-        {
-
-        }
-
-        private Gloves _gloves;
-        public void SetGloves(Gloves gloves)
-        {
-
-        }
+        public Weapon Weapon { get; set; }
```
Almantask: getter in equipment are not needed, if we refactor to EXACTLY what was before.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,35 +1,74 @@
...
+            Items = new List<Item>();
         }
 
-        public Item[] GetItems(string name)
+        /// <summary>
+        /// Returns the item it found based on the name you gave, if no item is found returns null
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public Item GetItem(string name)
```
Almantask: Could make an indexer :)

Src/BootCamp.Chapter/Item.cs
```diff
@@ -2,25 +2,15 @@
...
-        }
-
-        private decimal _price;
-        public decimal GetPrice()
-        {
-            return _price;
-        }
-
-        private float _weight;
+        public string Name { get; set; }
```
Almantask: No need setters. If you set soemething only  in ctor, setters can be removed.

PR #224, opened 2020-03-14
Chapter3/homework/1

chatieb:
Took Homework 2 of Chapter 2 and refactored using lists and properties.

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -1,71 +1,43 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
+        private List<Item> _items;
+        public List<Item> Items
         {
-            return _items;
+            get => _items;
         }
 
         public Inventory()
         {
-            _items = new Item[0];
+            _items = new List<Item>();
         }
 
-        public Item[] GetItems(string name)
+        public List<Item> GetItems(string name)
         {
-            var numberOfMatchName = GetMatchingNamesCount(name);
-            if (numberOfMatchName == 0)
-            {
-                return new Item[0];
-            }
-
-            var tempItems = new Item[numberOfMatchName];
-            var i = 0;
+            var tempItems = new List<Item>();
 
             foreach (var item in _items)
             {
-                if (item.GetName() == name)
+                if (item.Name == name)
                 {
-                    tempItems[i++] = item;
+                    tempItems.Add(item);
                 }
             }
 
             return tempItems;
         }
 
-        private int GetMatchingNamesCount(string name)
-        {
-            var matchName = 0;
-
-            foreach (var item in _items)
-            {
-                if (item.GetName() == name)
-                {
-                    matchName++;
-                }
-            }
-
-            return matchName;
-        }
-
         public void AddItem(Item item)
         {
-            var tempItems = new Item[_items.Length + 1];
-
-            for (var i = 0; i < _items.Length; i++)
-            {
-                tempItems[i] = _items[i];
-            }
-
-            tempItems[_items.Length] = item;
-            _items = tempItems;
+            _items.Add(item);
         }
 
         public void RemoveItem(Item item)
         {
-            RemoveItem(item.GetName());
+            _items.Remove(item);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Src/BootCamp.Chapter/Item.cs
```diff
@@ -3,15 +3,19 @@
     public class Item
     {
         private string _name;
```
Matt500b: you can remove the private backing field and just use auto-implemented properties

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,35 +1,58 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
+        private List<Item> _items;
```
Matt500b: can remove the private backing field and just use an auto-property

Src/BootCamp.Chapter/Item.cs
```diff
@@ -3,15 +3,19 @@
...
+        public string Name
+        {
+            get => _name;
+        }
         public string GetName()
```
Matt500b: Can remove these `Get` methods now we have a property

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,28 +1,30 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Shop
     {
         private decimal _money;
```
Matt500b: can use auto-implemented properties again

Src/BootCamp.Chapter/Item.cs
chatieb: Request help because confused here.  Cannot get rid of GetName().  Test will fail, because the instructor's test expect item.GetName() in https://github.com/csinn/CSharp-From-Zero-To-Hero/blob/Chapter3/Homework/1/Tests/BootCamp.Chapter.Tests/PlayerTests.cs.

If I remove it, it will fail to pass and not be accepted.  If I do not change it, you will want it changed.  Infinite loop.  Request guidance.

Here are screenshots of the Test code that uses the GetName() method:

image

image

Matt500b: Yep it was an accident in the homework - the tests should be using properties, so `item.GetName()` becomes `item.Name`

chatieb: Thank you for the reply.  I thought it was testing us and figuring out how to use multiple methods.  Also, your reply answered a question that I had about GitHub.

Src/BootCamp.Chapter/Item.cs
chatieb: As recommended, I have removed the private backing field. and am using the auto-implemented properities.

Src/BootCamp.Chapter/Inventory.cs
chatieb: As recommended, I have removed the private backing field. and am using the auto-implemented properities.

chatieb:
Ready for review.  As recommended, I removed the private backing field and just use auto-implemented properties   I had left comment in my resolves but unsure if I am also suppose to leave comment in this box which I am typing right now.  I am guessing that you've been too busy to review and that's why there's been no feedback.  Not sure if I am supposed to be typing something here.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,35 +1,54 @@
...
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
-        {
-            return new Item[0];
-        }
+        public List<Item> Items { get; private set; }
```
Almantask: private set can be removed. If a property is get only it can still be set, but only in ctor. This is exactly what you want.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,35 +1,54 @@
...
 
         public Inventory()
         {
-            _items = new Item[0];
+            Items = new List<Item>();
         }
 
-        public Item[] GetItems(string name)
+        public List<Item> GetItems(string name)
```
Almantask: Could be an indexer :)

Src/BootCamp.Chapter/Item.cs
```diff
@@ -2,24 +2,16 @@
 {
     public class Item
     {
-        private string _name;
-        public string GetName()
-        {
-            return _name;
-        }
+        public string Name { get; private set; }
```
Almantask: set can be removed.

Almantask:
I would say the PR is pretty good, you can just removed those private setters because if you set just in a ctor, they are not needed.

PR #225, opened 2020-03-15
C3H1 from Aurora

Aurora2500:


Src/BootCamp.Chapter/Player.cs
```diff
@@ -75,32 +79,46 @@ public Item[] GetItems(string name)
         // Implement equiping logic and total defense/attack calculation.
         public void Equip(Headpiece head)
         {
-
+            _equipment.Head = head;
```
Matt500b: You've done most of the challenge except the last piece. 

"Player cannot equip items that exceed total weight allowed. The total max weight player has is determined by their strength status (30 + strength * 10 (kg))."

Meaning you need to do a check to see if you can equip new items or not based on current total weight.

Src/BootCamp.Chapter/Items/Item.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter.Items
+{
+    public class Item
```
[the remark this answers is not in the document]
Aurora2500: I tried making it one, but some of the tests use it instead of the more concrete child classes.
So I left it like that to make it pass the tests.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -64,7 +74,18 @@ public decimal Buy(Item item)
...
         {
-            return null;
+            var itemList = _inventory.GetItems(item);
+
+            if (itemList.Count == 0)
+            {
+                return null;
+            }
+
+            // itemList should be a singleton since a shop can't contain multiple items of the same name
```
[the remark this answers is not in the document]
Aurora2500: Yeah, in this case I was using the [mathematic definition](https://en.wikipedia.org/wiki/Singleton_(mathematics)).

Src/BootCamp.Chapter/Items/Item.cs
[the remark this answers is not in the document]
Matt500b: I see it's due to the tests, ignore this

Src/BootCamp.Chapter/Player.cs
Aurora2500: I managed to come up with a fix for it, however it is very verbose and I'm wondering if there is a better way to do so so I'll leave this open for now.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,36 +1,36 @@
...
-        }
+        public List<Item> Items { get => _inventory.GetItems(); }
 
         /// <summary>
         /// Adds item to the stock.
         /// If item of same name exists, does nothing.
         /// </summary>
         public void Add(Item item)
         {
+            var items = _inventory.GetItems(item.Name);
```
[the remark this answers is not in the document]
Aurora2500: It was meant to check that such item didn't already exist, added the missing logic.

Src/BootCamp.Chapter/Player.cs
```diff
@@ -75,32 +79,46 @@ public Item[] GetItems(string name)
...
         }
 
-        public void Equip(Chestpiece head)
+        public void Equip(Chestpiece chest)
         {
```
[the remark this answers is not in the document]
Aurora2500: now it is :D

Src/BootCamp.Chapter/Player.cs
```diff
@@ -17,25 +18,27 @@ public class Player
         /// </summary>
         private const int baseCarryWeight = 30;
 
-        private string _name;
-        private int _hp;
+        public string Name { get; }
+        public int Hp { get; private set; }
```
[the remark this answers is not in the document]
Aurora2500: Changed it all to explicit, Kai mentioned that if it was implicit it would be set in the ctor and I might have gotten the wrong impression that it would only be then.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,40 @@
...
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
+        private List<Item> _items;
+
+        public List<Item> GetItems()
```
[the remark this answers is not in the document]
Aurora2500: Method removed and made it into a property

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,40 @@
...
         {
-            _items = new Item[0];
+            _items = new List<Item>();
         }
 
-        public Item[] GetItems(string name)
+        public List<Item> GetItems(string name)
         {
-            return new Item[0];
+            var items = GetItems();
```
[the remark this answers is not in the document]
Aurora2500: Changed all the uses of Item to access it directly instead of using the method.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,35 @@
...
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
-        {
-            return new Item[0];
-        }
+        private List<Item> _items;
+        public List<Item> Items { get { return new List<Item>(_items); } private set { _items = value; } }
```
Matt500b: almost. you can remove the private `_items` and just use auto-implemented properties `{ get; set; }` on the public one.

also, note that since _items is a `List<Item>` if you were to keep this format, you wouldn't need `return new List<Item>(_items);` just `return _items;` but you wont need to keep it, more of an FYI :)

Src/BootCamp.Chapter/Player.cs
```diff
@@ -73,34 +77,116 @@ public Item[] GetItems(string name)
...
+            }
+            if (weightIfEquipped <= EquipCapacity)
+            {
+                _equipment.Gloves = gloves;
+            }
+        }
 
+        float WeightIfAdded(Item item)
+        {
+            return _equipment.GetTotalWeight() + item.Weight;
```
Matt500b: If you're equipping a null item (which pretty much means unequipping!) then this would throw a `NullReferenceExecption`

Src/BootCamp.Chapter/Player.cs
```diff
@@ -73,34 +77,116 @@ public Item[] GetItems(string name)
         // When a slot is equiped, it contributes to total defense
         // and total attack.
         // Implement equiping logic and total defense/attack calculation.
+        private float EquipCapacity { get { return baseCarryWeight + 10 * Strength; } }
         public void Equip(Headpiece head)
         {
-
+            var weightIfEquipped = WeightIfAdded(head) ;
+            if (_equipment.Head != null)
```
Matt500b: this check is being repeated, probably could move it into the `WeightIfAdded` method

Matt500b: You could even make the method return a bool to really slim down all the equip methods! Then all the logic is in one place, less prone to copy/paste mistakes etc

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,67 +7,54 @@ namespace BootCamp.Chapter
...
-        {
-
-        }
+        public Gloves Gloves { get; set; }
 
         /// <summary>
         /// Gets total weight of armour.
         /// </summary>
         /// <returns></returns>
         public float GetTotalWeight()
```
Almantask: This can be a property

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,67 +7,54 @@ namespace BootCamp.Chapter
...
-        {
-
-        }
-
-        private Gloves _gloves;
-        public void SetGloves(Gloves gloves)
-        {
-
-        }
+        public Gloves Gloves { get; set; }
```
Almantask: All your get;set; before were just SetX. Please amend getters, as they are not needed here.

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -76,7 +63,29 @@ public float GetTotalWeight()
         /// <returns></returns>
         public float GetTotalDefense()
```
Almantask: Can be a property

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -85,7 +94,11 @@ public float GetTotalDefense()
         /// <returns></returns>
         public float GetTotalAttack()
```
Almantask: Can be a property

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,35 @@
...
 
         public Inventory()
         {
-            _items = new Item[0];
+            _items = new List<Item>();
         }
 
-        public Item[] GetItems(string name)
+        public List<Item> GetItems(string name)
```
Almantask: Can be an indexer ;)

Src/BootCamp.Chapter/Items/Armor.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter.Items
+{
+    public abstract class Armor : Item
```
Almantask: Well done for using abstract

Src/BootCamp.Chapter/Player.cs
```diff
@@ -17,25 +18,27 @@ public class Player
...
-        private string _name;
-        private int _hp;
+        public string Name { get; private set; }
+        public int Hp { get; private set; }
 
         /// <summary>
         /// Each point of strength allows extra 10 kg to carry.
         /// </summary>
-        private int _strenght;
+        public int Strength { get; private set; }
```
Almantask: Where do yo uset the stength?

Almantask:
> _No description provided._

Please don't be lazy and add descriptions to your PRs.

Src/BootCamp.Chapter/Inventory.cs
Aurora2500: Wouldn't it be passing by reference then? That way something else can change the internal items list

Almantask: That's a good point. The solution to this is to work with a readonly list. So private have List<Item>, publicly return ReadonlyList. It;s inneficient to create a list every time you need to return

Src/BootCamp.Chapter/Player.cs
Aurora2500: I am having a problem with this part, since it's all checking for different parts, I basically would need a switch case for types

Matt500b: if you change the method signature to something like `Item currentItem, Item newItem` then you can check it without any case.

```
if (CanEquipItem(_equipment.Head, head))
    _equipment.Head = head;
```

PR #226, opened 2020-03-16
Add a solution for Homework 7 Chapter 1

tzurowski:
Merge is only for update changes cuz I had older version

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -1,4 +1,6 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+
+namespace BootCamp.Chapter
 {
     public static class BalanceStats
     {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,234 @@ public static class BalanceStats
...
-            return "";
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0)
+            {
+                return "N/A.";
+            }
+
+            var splitArray = peopleAndBalances[0].Split(',');
+            var floatArray = ConvertStringArrayToFloat(splitArray[1..]);
+            var names = new[] { splitArray[0] };
+            var heighest = FindHighest(floatArray);
```
Almantask: Highest what? Don't be too lazy to add the extra bits to the method nmes, because it makes the intention more clear, without having to know all the context.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,234 @@ public static class BalanceStats
...
+            {
+                return currentNames[0];
+            }
+            var formatArray = String.Join(", ", currentNames[..^1]);
+            formatArray += $" and {currentNames[^1]}";
+
+            return formatArray;
+        }
+
+        private static string[] AddStrings(string[] currentNames, string v)
```
Almantask: A method such as this would expect 2 strings for parameters. However, this has array of string and a string. I recommend calling this AddString instead, because you add just one string.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,234 @@ public static class BalanceStats
...
+            {
+                if (floatArray[i] > currentMax)
+                {
+                    currentMax = floatArray[i];
+                }
+            }
+            return currentMax;
+        }
+
+        private static float[] ConvertStringArrayToFloat(string[] v)
```
Almantask: ..ToFloatArray

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,234 @@ public static class BalanceStats
...
+                else if (biggestLoss == currentBiggestLoss)
+                {
+                    AddStrings(names, currentSplitArray[0]);
+                }
+            }
+
+            return $"{FormatStrings(names)} lost the most money. -¤{biggestLoss}.";
+        }
+
+        private static float FindLowest(float[] floatArray)
```
Almantask: Lowest what? It should be lowest balance

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +11,11 @@ static void Main(string[] args)
             // - FindPersonWithBiggestLoss
             // - FindRichestPerson
             // - FindMostPoorPerson
+            Console.WriteLine(TextTable.Build(BalanceStats.FindHighestBalanceEver(PeoplesBalances.Balances),3));
+            Console.WriteLine(TextTable.Build(BalanceStats.FindPersonWithBiggestLoss(PeoplesBalances.Balances),3));
+            Console.WriteLine(TextTable.Build(BalanceStats.FindRichestPerson(PeoplesBalances.Balances),3));
+            Console.WriteLine(TextTable.Build(BalanceStats.FindMostPoorPerson(PeoplesBalances.Balances),3));
```
Almantask: I would put 0 blank lines in beween those text tables- because they are the same/similar thing.
However, I would put a blank line after Console.ReadKey() as a way to separate it from the rest, indicating that it's a completely different thing/intention

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -26,6 +29,9 @@ public static class TextTable
            +-------+
 
          */
+        private static string _corner = "+";
```
Almantask: instead of private static use const

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +40,75 @@ public static class TextTable
...
+            }
+
+            var wordArray = message.Split(Environment.NewLine);
+
+            int maxLength = MaxLength(wordArray);
+            int messageWidth =  maxLength + 2 * padding;
+
+            StringBuilder textTable = new StringBuilder();
+
+            AddHorizontalLine(textTable, messageWidth);
```
Almantask: Well done splitting this function into steps.

Src/BootCamp.Chapter/BalanceStats.cs
tzurowski: Okay, sorry, changed.

Src/BootCamp.Chapter/BalanceStats.cs
tzurowski: Good point, changed.

Src/BootCamp.Chapter/BalanceStats.cs
tzurowski: Changed, FindHeighest changed to FindFighestBalance too

Src/BootCamp.Chapter/Program.cs
tzurowski: You're, code should be clean and more readable.

Src/BootCamp.Chapter/TextTable.cs
tzurowski: Okay, my bad

PR #228, opened 2020-03-17
Add a solution for Homework 9 Chapter 1

tzurowski:


Src/BootCamp.Chapter/ArrowMovement.cs
```diff
@@ -18,7 +18,22 @@ public static class ArrowMovement
         /// <returns>One of the arrow characters. '↥' by default.</returns>
         public static char GetIndicator(char symbol)
         {
-            return '-';
+            switch (symbol)
```
Almantask: Well done- this is clear and simple.

Src/BootCamp.Chapter/ArrowMovement.cs
```diff
@@ -18,7 +18,22 @@ public static class ArrowMovement
...
+                case 'a':
+                    return '↤';
+                case 'S':
+                case 's':
+                    return '↧';
+                case 'D':
+                case 'd':
+                    return '↦';
+            }
+            return '↥';
```
Almantask: this should go to ``default:`` inside the switch

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -8,12 +8,50 @@ public static class BinaryConverter
...
+                    return false;
+                }
+            }
+            return true;
         }
 
         public static string ToBinary(long number)
         {
-            return "";
+            Stack<int> myBinaryStack = new Stack<int>();
```
Almantask: You don't need a stack here- you can simply append the number to string right away.

Src/BootCamp.Chapter/ArrowMovement.cs
tzurowski: You're right, changed.

Src/BootCamp.Chapter/BinaryConverter.cs
tzurowski: Changed, but my way can be little to mess :/

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -37,20 +37,22 @@ private static bool IsBinary(string binary)
 
         public static string ToBinary(long number)
         {
-            Stack<int> myBinaryStack = new Stack<int>();
+            string binaryWord = "";
```
Almantask: Prefer to use ``StringBuilder`` instead when concatinating a string in a loop. It is much faster (like a hundred times faster)

PR #230, opened 2020-03-18
Chapter 3 Homework 2

chatieb:
Coded an insecure credentials manager, which works supports register and login functions.

Coded as per Discord bc-discussion with Kai:
1) Program.cs should just call what is already implemented.
2) Save and ready credentials from/to file
3) Can save multiple credentials but only one user can be logged on at a time

Coded Register works like this:
1) User inputs name and password
2) Password gets encoded to Unicode
3) Username (not encoded) and password (encoded) are persisted in a file

Coded Login works like this:
1) Reads the credentials file
2) User inputs their credentials
3) Encodes password
4) Encoded password is searched for in credentials field
5) Comparison between encoded passwords is made
6) Is password, login returns true, else returns false.

Coded Extra #1: credentials parsing from the file should be done applying TryParse pattern. 
Coded Extra #2: inputting password should be hidden with smile emoji character.


Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace BootCamp.Chapter
+{
+    internal struct Credentials
+    {
+        public readonly string Username;
+        public readonly string Password;
+
+        public Credentials(string username, string password)
+        {
+            Username = username;
```
Almantask: It would be nice to validate against null or empty

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+{
+    internal class CredentialsManager
+    {
+        private const string CredentialsPath = @"..\..\..\Database\Credentials.txt";
+
+        public CredentialsManager()
+        {
+            if (!File.Exists(CredentialsPath))
+            {
+                File.WriteAllText(CredentialsPath, "");
```
Almantask: use this instead:
```
File.Create(filename).Dispose();
```
It does just what you need- create a new file and close it.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+
+        public void Register()
```
Almantask: Register should accept user. Getting user input should come from outside- the console part. Don't couple console and business logic.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                Console.WriteLine("You have successfully registered.\n");
+            }
+            else
+            {
+                Console.WriteLine("Sorry, your registration was unsuccessful.  The username is already taken.\n");
+            }
+        }
+
+        private bool RegisterUser()
```
Almantask: Split persistence of credentials and getting input

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            string password = "";
+
+            do
+            {
+                password = Utilities.PromptPassword("Enter password:");
+            } while (string.IsNullOrEmpty(password.Trim()));
+
+            return new Credentials(username, password);
+        }
+
```
Almantask: A bit too many blank lines

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                Console.WriteLine("You have sucessfuly logged in.\n");
+                return true;
+            }
+
+            Console.WriteLine("Sorry, we don't recognize you.\n");
+            return false;
+        }
+
+        private Credentials GetNameAndPassword()
```
Almantask: You don't get credentials. You ReadCredentials (from console)

Almantask: And ofc- instead of many words, use your domain language. PasswordAndUsername = Credentials.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            var encodedPassword = new Cryptography().Encode(user.Password);
+            File.AppendAllText(CredentialsPath, user.Username);
+            File.AppendAllText(CredentialsPath, Delimiter);
+            File.AppendAllText(CredentialsPath, encodedPassword);
+            File.AppendAllText(CredentialsPath, CarriageReturn);
+
+            return true;
+        }
+
+        public bool Login()
```
Almantask: Login should take a user.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            var user = GetNameAndPassword();
```
Almantask: Login should be decoupled from getting input.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            if (storedUser.Username == null || storedUser.Password != new Cryptography().Encode(user.Password))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        private Credentials GetUser(string username)
```
Almantask: You get Credentials. Maybe user credentials. But not user itself.

Src/BootCamp.Chapter/Credentials.cs
chatieb: As requested, made change to validate username and password against null or empty.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: As requested, I am using the File.Create(filename).Dispose() command in place of the code that I was using.  The effect seems to be the same when testing.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: Changed code for Register function to accept credentials.   The getting of the input was separated from the console part.  This was due to a misunderstanding that "Program.cs should just call what is already implemented" in a reply thinking that mean everything needs to be together inside.  Logic was separated to the way it was originally designed.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: This is probably referring to the input and business logic being bunched together.  That change has been undone.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: Removed the extra blank lines.  You have eagle eyes!

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: Changed the wording to use Credentials instead of PasswordAndUsername for function name.  I appreciate very much whenever you point this out.  Thank you.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: I have separated the input and the business logic.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: Changes made when the code was coded saying getting user information, such as "GetUser" name.  When I switched to using the term "Credential", I did not change the all the places in the code where was still referring as "User".  Thank you for catching this with your sharp eyes.  It showed me that I have to be more careful when making global name changes.

Src/BootCamp.Chapter/CredentialsManager.cs
chatieb: Changed Login function to take credentials.

chatieb:
Please review when convenient as I value your feedback.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -2,13 +2,40 @@
...
         {
-            Username = username;
-            Password = password;
+            get
+            {
+                return _username;
+            }
+            set
+            {
+                if (value == null || value.Trim() == "")
```
Almantask: I like the validation inside a property. However, you should have used a built-in function for it:
```string.IsNullOrWhitespace(value)```

File: Src/BootCamp.Chapter/Credentials.cs [hunk 11]
```diff
@@ -2,13 +2,40 @@
...
+
+        public string Password
+        {
+            get
+            {
+                return _password;
+            }
+            set
+            {
+                if (value == null || value.Trim() == "")
```
Almantask: string.IsNullOrWhiteSpace instead

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -12,157 +12,106 @@ public CredentialsManager()
...
-                users.Add(user);
+                credentialsList.Add(parsedCredentials);
             }
 
-            return users;
+            return credentialsList;
         }
 
-        private bool TryCredentialsParse(string credential, out Credentials users)
+        private bool TryCredentialsParse(string credentials, out Credentials parsedCredentials)
```
Almantask: This should be inside the Credentials class. Remember the TryParse pattern?
It shoudl be Credentials.TryParse(...)

PR #232, opened 2020-03-18
Chapter3/homework/2

mihailfox:
I implemented Login, Register, TryParse and password hiding logic. Added a simple console user interface for demonstration purpose. I did some exception implementation but I will remove them when the tests come in if it is not needed.


Src/BootCamp.Chapter/StringOps.cs
```diff
@@ -0,0 +1,23 @@
...
+            foreach (var character in temp)
+            {
+                encodedPassword.Append(character);
+            }
+            return encodedPassword.ToString();
+        }
+
+        public static bool IsValid(string input)
+        {
+            return !string.IsNullOrWhiteSpace(input) || !string.IsNullOrEmpty(input);
```
Matt500b: probably want's to be `&&`? not null, not whitespace and not empty?

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            if (!StringOps.IsValid(input))
+            {
+                return false;
+            }
+
+            var fields = input.Split(Separator);
+            const int fieldsNumber = 2;
+
+            var isValid = fields.Length == fieldsNumber || StringOps.IsValid(fields[0]) || StringOps.IsValid(fields[1]);
```
Matt500b: wouldn't these be `&&`s also?

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
...
+            return found;
+        }
+
+        private bool AddUser(User user)
+        {
+            if (FindUser(user))
+            {
+                throw new UserAllreadyExistsException("User already exists!");
+            }
+            StreamWriter writer;
```
Matt500b: not sure if you know about usings? but you should use one! Otherwise if something goes wrong you wont be tidying up resource.

Src/BootCamp.Chapter/UserInteraction.cs
```diff
@@ -0,0 +1,189 @@
...
+            Console.WriteLine("(1) Login");
+            Console.WriteLine("(2) Register");
+            Console.WriteLine();
+            Console.WriteLine("(0) Exit!");
+
+            try
+            {
+                ChooseMenu();
+            }
+            catch (InvalidCredentialsDbFile ex)
```
Matt500b: Optional: If you know about the catch when syntax totally be cool to use it here :)

Src/BootCamp.Chapter/UserInteraction.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            catch (UserAllreadyExistsException ex)
+            {
+                Console.WriteLine(ex.Message);
+            }
+        }
+
+        private static void ChooseMenu()
+        {
+            ConsoleKey pressedKey = Console.ReadKey(true).Key;
```
Matt500b: if you're going to read the specific key, probably should support numpad numbers also :)

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public class Credentials
+    {
+        private const string Separator = ",";
+
+        public string CredentialsFile { get; } = "credentials.db";
```
Matt500b: would anything outside of this class need to know the file location?

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
...
+            string line;
+            bool found = false;
+
+            StreamReader reader;
+            try
+            {
+                reader = new StreamReader(CredentialsFile);
+                while ((line = reader.ReadLine()) != null)
+                {
+                    var isValid = TryParse(line, out User credentials);
```
Matt500b: can inline this if you want, no need for the variable

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public Credentials(string credentialsFile)
+        {
+            CredentialsFile = credentialsFile ?? throw new ArgumentNullException(nameof(credentialsFile));
+        }
+
+        private bool FindUser(User user)
+        {
+            string line;
+            bool found = false;
```
Almantask: Keep it consistent. Either explicit all or var all except for cases where it isn't obvious. I think this should be var.

Src/BootCamp.Chapter/User.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    public readonly struct User : IEquatable<User>
```
Almantask: It's not a user, it's credentials

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public class Credentials
```
Almantask: This is not credentials. This is ``CredentialsManager`` or ``CredentialsRegistry``

Src/BootCamp.Chapter/Credentials.cs
mihailfox: This one led to refactoring the whole class.
I fixed in [2a4af80](https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/232/commits/2a4af80f11d86f38c8efd0a1debe8b0fd5ae911e)

Src/BootCamp.Chapter/StringOps.cs
mihailfox: I don;t know how could I mistake AND with OR and keep everything working.
I fixed in [2a4af80](https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/232/commits/2a4af80f11d86f38c8efd0a1debe8b0fd5ae911e)

Src/BootCamp.Chapter/Credentials.cs
mihailfox: I know about using but I chose not to use it but in the process I forgot to add the 'finally' block and close/dispose StreaWriter/Reader.
Check the latest commit [2a4af80](https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/232/commits/2a4af80f11d86f38c8efd0a1debe8b0fd5ae911e)

Src/BootCamp.Chapter/UserInteraction.cs
mihailfox: I know about catch when but first time I use it in a homework Codacy was screaming at me that it does not understand the code.

Src/BootCamp.Chapter/UserInteraction.cs
mihailfox: Last time I implemented a console menu and used both type of keys Codacy complained about Cyclomatic Complexity. I think it should not be a problem now as there are less options to choose from.

Src/BootCamp.Chapter/Credentials.cs
mihailfox: Maybe to display it while working with it? I changed it to private for now.

Src/BootCamp.Chapter/Credentials.cs
mihailfox: Indeed, I updated it.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        private const string Separator = ",";
+
+        public string CredentialsFile { get; } = "credentials.db";
+
+        public Credentials()
```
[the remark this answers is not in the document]
mihailfox: Yes. By default it will  use "credentials.db".

Src/BootCamp.Chapter/Credentials.cs
mihailfox: I was fixing/updating the code as you were reviewing my PR. I did not know that you will also review it. Thank you for the observation.

Src/BootCamp.Chapter/User.cs
mihailfox: I need to practice my English more. It was not sounding right to me but could not find better naming at that time. Cheers!

Src/BootCamp.Chapter/Credentials.cs
mihailfox: I updated my code in latest commit [2a4af80](https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/232/commits/2a4af80f11d86f38c8efd0a1debe8b0fd5ae911e)

mihailfox:
Many thanks for the reviews Kai & Matt!

mihailfox:
IMO the user should not be able to register if the user already exists.

I modified the tests in [3bbda41](https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/232/commits/3bbda41a533d4e86cab20531b6f9edb3b4143d22)

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,72 +1,55 @@
 ﻿using System;
-using System.IO;
 
 namespace BootCamp.Chapter
 {
-    public class Credentials
+    public readonly struct Credentials : IEquatable<Credentials>
```
Almantask: Love the use of interface and readonly struct. Even though it wasn't really a requirement, it certainly fits there.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,72 +1,55 @@
...
-                throw new InvalidCredentialsDbFile($"There was an error while trying to work with {CredentialsFile}", ex);
-            }
+            return HashCode.Combine(Name, Password);
+        }
 
-            writer?.Close();
-            return true;
+        public override bool Equals(object obj)
+        {
+            return obj is Credentials user && Name == user.Name && Password == user.Password;
```
Almantask: You can reuse already existing code- the other Equals overload.
```
Credentials cred = obj as Credentials;
return cred != null && cred.Equals(this)
```

Src/BootCamp.Chapter/UserInteraction.cs
```diff
@@ -21,38 +21,34 @@ public static void DisplayMainMenu()
 
             try
             {
-                ChooseMenu();
+                ChooseMenuItem();
             }
-            catch (InvalidCredentialsDbFile ex)
+            catch (Exception ex) when (ex is InvalidCredentialsDbFile ||
```
Almantask: Lovely exception handling :)

Src/BootCamp.Chapter/Credentials.cs
mihailfox: this should be:
Credentials? cred = obj as Credentials?;
return cred != null && cred.Equals(this);

mihailfox:
Thanks Kai! Btw you should sleep more :)

PR #234, opened 2020-03-22
Solved Chapter3/HW3

RoelofWobben:
I used as much as I can the ?,  and ?? but on a few places I could not. 
So if I have to use it there I need some hints how

### DIFF
--- a/Src/BootCamp.Chapter/Shop.cs
+++ b/Src/BootCamp.Chapter/Shop.cs
@@ -1,28 +1,26 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Shop
     {
-        private decimal _money;
-        public decimal GetMoney()
+        public decimal Money { get; set; }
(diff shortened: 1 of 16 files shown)


Src/BootCamp.Chapter/Shop.cs
```diff
@@ -31,6 +29,12 @@ public Item[] GetItems()
         /// </summary>
         public void Add(Item item)
         {
+            if (item == null)
```
Almantask: This can be refactored.
```
_inventory.AddItem(item ?? throw new ArgumentNullException);
```

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -40,6 +44,11 @@ public void Add(Item item)
         /// <param name="name"></param>
         public void Remove(string name)
         {
+            if (String.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException();
+            }
+            _inventory.RemoveByName(name);
```
Almantask: Can be refactored with ``??``

PR #236, opened 2020-03-27
Ch3 hw2 from Aurora

Aurora2500:


Src/BootCamp.Chapter/Application.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    class Application
```
Almantask: please declare access modifiers explicitly, as it is a convention

Src/BootCamp.Chapter/Application.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        readonly CredentialsManager manager;
```
Almantask: By convention, private fields are supposed to have a prefix with ``_``.

Src/BootCamp.Chapter/Application.cs
```diff
@@ -0,0 +1,70 @@
...
+        readonly UI ui;
+
+        public Application(string credentialsFile)
+        {
+            manager = new CredentialsManager(credentialsFile);
+            ui = new UI();
+        }
+        void Register()
+        {
+            Credentials inputCredential;
```
Almantask: ``inputCredential`` is redundant and ``credential`` is enough.

Src/BootCamp.Chapter/Application.cs
```diff
@@ -0,0 +1,70 @@
...
+            bool registerError = false;
+            do
```
Almantask: I like the use of do while, fits very well!

Src/BootCamp.Chapter/Application.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace BootCamp.Chapter
+{
+    class Application
+    {
+        readonly CredentialsManager manager;
+        readonly UI ui;
+
+        public Application(string credentialsFile)
+        {
+            manager = new CredentialsManager(credentialsFile);
```
Almantask: Please rename to CredentialsManager. Because right now- certainly you have only one component which does something that could be called managing, but that could change quite easily. Also if you come back to the code after a month or two, you will have a hard time remembering what this manager really is for.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,13 +1,22 @@
...
+
+namespace BootCamp.Chapter
 {
     // TODO: make a struct and add validation and other needed methods (if needed)
     public class Credentials
     {
         public string Username;
         public string Password;
 
+        public string ToFile { get { return $"{Username},{Password}"; } }
```
Almantask: Here you don't do anything with a file, so it looks deceiving.
It's just converting Credentials ToString.
``ToString()`` is a virtual method that you can override, so please do :). Also, don't forget that you can use expressions to simplify it
```
public string ToFile => ${Username},{Password}";
```

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,13 +1,22 @@
...
+
         public Credentials(string username, string password)
         {
+            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(password))
+            {
+                throw new ArgumentException();
```
Almantask: Please validate each separately and provide which argument was null. Use ``nameof`` for that purpose.
In general, when you throw an excpetion, you should give as much detail as possible to identify the cause.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -16,6 +25,30 @@ public Credentials(string username, string password)
...
+            string username = parts[0];
+            string password = parts[1];
+
+            credentials = new Credentials(username, password);
+            return true;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null)
```
Almantask: This check if redundant. If object is null, then obj is Credentials will also be false.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -16,6 +25,30 @@ public Credentials(string username, string password)
...
+            {
+                return false;
+            }
+            if (obj is Credentials)
```
Almantask: You could simplify this further
```
public override bool Equals(object obj)
{
    Credentials other = obj as Credentials;
    return Username == other?.Username && Password == other?.Password;
}
```

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -1,24 +1,78 @@
...
 {
     public class CredentialsManager
     {
         private readonly string _credentialsFile;
 
         public CredentialsManager(string credentialsFile)
         {
+            if (string.IsNullOrEmpty(credentialsFile))
+            {
+                throw new ArgumentException();
```
Almantask: nameof should be added to ArgumentExcpeiton

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -1,24 +1,78 @@
...
+            }
+            
+            File.AppendAllText(_credentialsFile, credentials.ToFile);
+        }
+
+        List<Credentials> GetCredentials()
+        {
+            List<Credentials> creds = new List<Credentials>();
 
+            string fileContent = File.ReadAllText(_credentialsFile);
```
Almantask: You can use ``File.ReadAllLines()`` to read file as an array of lines right away.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -4,7 +4,13 @@ class Program
     {
         static void Main(string[] args)
```
Almantask: Love the thin Main function- well done! 👍

Src/BootCamp.Chapter/UI.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class UI
+    {
+        readonly char passwordChar = '*';
```
Almantask: instead of private readonly, use a const.

Src/BootCamp.Chapter/UI.cs
```diff
@@ -0,0 +1,71 @@
...
+            Console.Write("Password: ");
+            string unparsedPassword = HiddenReadLine(passwordChar);
+            
+            var utf8 = Encoding.UTF8;
+            var bytes = utf8.GetBytes(unparsedPassword);
+            string password = string.Join("", bytes);
+            return new Credentials(username, password);
+        }   
+
+        string HiddenReadLine(char hidenChar)
```
Almantask: ``HiddenReadLine`` is a noun. ``ReadLineHidden`` is a verb. I suggest using that.

Src/BootCamp.Chapter/UI.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            string typedAsFar = "";
+            ConsoleKeyInfo key;
+
+            do
+            {
+                key = Console.ReadKey(true);
+
+                // make sure the key is in the range
+                if (((int)key.Key) >= 48 && ((int)key.Key <= 90))
```
Almantask: Please use constants to explain 48 and 90 magic values.

Src/BootCamp.Chapter/CredentialsManager.cs
Aurora2500: What would the nameof do?

PR #239, opened 2020-03-30
Max Chapter3/homework/2

HellHunterMax:
some tests fail. 
1. empty file cannot be compared 
2. passwords in the file are not encoded.

First i made the project without the tests then wanted to open a pull request and saw that my version was out of date so i pulled in the new version and kind of started over. much less code. i saved my old version in a new branch.

any comments are welcome.

EDIT: woeps i just saw that i should add back some parts because now you cant login with console. so there is no demo ill add that and push again.


HellHunterMax:
Added the demo.
no extra 1 or 2 though.

HellHunterMax:
Looked at some pointers Kai gave on Roelof and Aurora's PR and changed some code based on that.

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+{
+    public class Area51
+    {
+        private List<Credentials> loggedInAccounts = new List<Credentials>();
+        readonly CredentialsManager credentialsManager = new CredentialsManager("Login.txt");
+        const string login = "login";
+        const string create = "create";
+        const string see = "see";
+        const string quit = "quit";
+        public void Demo()
```
Almantask, line 14: Could add a blank to separate unrelated parts of code

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+            string answer = string.Empty;
+            while (answer != quit)
+            {
+                answer = GetAnswerMainMenu();
+                GoToPartOfProgram(answer);
+            }
+        }
+        private string GetAnswerMainMenu()
+        {
+
```
Almantask, line 25: Blank is not needed.

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+            string input = string.Empty;
+            bool isLoopMainMenu = true;
+
+            while (isLoopMainMenu)
+            {
+                Console.Clear();
+                Console.WriteLine("Login\r\nCreate\r\nSee\r\nQuit");
```
Almantask, line 32: prefer to use ``Environemnt.NewLine`` over \r\n because that's just for windows.

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+                Console.Clear();
+                Console.WriteLine("Create Account.");
+                Console.Write("Login Name:");
+                string username = Console.ReadLine();
+
+                Console.Write("Password:");
+                string password = Console.ReadLine();
+                try
+                {
+                    if (Credentials.TryParse(username + "," + password, out Credentials credentials)
```
Almantask, line 75: I'd split this into two lines:
```cs
var isCredentials = Credentials.TryParse(username + "," + password, out Credentials credentials);
if(isCredentials && Login(credentials)
```

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+                    Console.WriteLine("Terminating Program!");
+                    Console.ReadKey();
+                    break;
+            }
+        }
+        private void Login()
+        {
+            bool isKeepLoopingLogin = true;
+
+            while (isKeepLoopingLogin)
```
Almantask, line 64: Could split this big method into steps like:
PromptName()
PromptPassword()
...

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+
+                Console.Write("Password:");
+                string password = Console.ReadLine();
+                try
+                {
+                    if (Credentials.TryParse(username + "," + password, out Credentials credentials)
+                    && credentialsManager.Login(credentials))
+                    {
+                        Console.Clear();
+                        Console.WriteLine($"Logging {username} in now.");
```
Almantask, line 79: This could probably be a part of credentialManagers login method (only these 3 lines)

Src/BootCamp.Chapter/Area51.cs
```diff
@@ -0,0 +1,204 @@
...
+                }
+                else
+                {
+                    isInLoopForNameOrPassword = false;
+                }
+            }
+            return input;
+        }
+
+        private void See()
```
Almantask, line 175: See is a weird name, ambigious. What you do here is DisplayLoggedOnUsers

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,22 +1,66 @@
...
         public Credentials(string username, string password)
         {
+            CheckForNullAndEmpty(username, password);
+
             Username = username;
-            Password = password;
+            Password = EncodePasswordToString(password);
+        }
+
+        private static void CheckForNullAndEmpty(string username, string password)
```
Almantask, line 19: Probably an overkill to split

Almantask:
The test is failing for the wrong reasons. I have fixed it in the latest commit. I don't demand to include that with your PR. So just solve the remarks left and you're good!

Src/BootCamp.Chapter/Area51.cs
HellHunterMax: totally right. i just thought i should name it the same as in the program but this is more readable.

Src/BootCamp.Chapter/Area51.cs
HellHunterMax: is this good enough or would you go further ?

PR #240, opened 2020-04-01
Chapter3/homework/6

RoelofWobben:
solved all homework so the predicates work and it works also with events.


Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,24 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Globalization;
+
+namespace BootCamp.Chapter
 {
     public class Person
     {
-        // add missing properties
+        public string Name { get; set; }
```
Matt500b, line 8: The properties should have their correct data type and not just string because the source was a text file!

BirthDay & Gender should change types for sure.
If you really wanted to push the boundary, Country, and Email could change also. but not needed

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
+            if (!File.Exists(peopleFile))
+            {
+                throw new FileNotFoundException("File is made up"); 
+            }
+                      
+            _peopleFile = File.ReadAllText(peopleFile); 
+           
+            if (String.IsNullOrWhiteSpace(_peopleFile))
+            {
+                throw new Exception();
```
Matt500b, line 23: What is this exception for? No details just "something went wrong". Not going to be useful when debugging an stack trace. 

Tend to never want to just throw `Exception` but a specific type. `InvalidOperationException` might be better here because that suggests that something is not in the right state for this path.

Also, a message would be good.
Either

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
         /// <summary>
         /// Gets people by filter predicate.
         /// </summary>
         /// <returns></returns>
         public List<Person> Filter(Predicate<Person> predicate)
         {
             var people = new List<Person>();
-            // ToDo: implement applying filter.
+            var splittedList = _peopleFile.Split($"{Environment.NewLine}");
+            for (int i = 1; i < splittedList.Length -1 ; i++)
```
Matt500b, line 37: Comment that you've skipped the header row - if i was just reading this code i'd not understand why you did this. In fact i had to double check the data file to see if you did this because of a header or not :D

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
+       
         /// <summary>
         /// Gets people by filter predicate.
         /// </summary>
         /// <returns></returns>
         public List<Person> Filter(Predicate<Person> predicate)
         {
             var people = new List<Person>();
-            // ToDo: implement applying filter.
+            var splittedList = _peopleFile.Split($"{Environment.NewLine}");
```
Matt500b, line 36: `Environment.NewLine` is already a string so you don't need it wrap it in `$""`

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
+
+            }
+                   
             return people;
         }
+
+        public Person ConvertStringToPerson(string personString)
+        {
+            var splittedPerson = personString.Split(',');
+            var person = new Person();
```
Matt500b, line 53: It'd be tidier if you pass the split string into a person constructor and the constructor handles setting the fields.

Though i can see merit to doing it here also so might want another mentor/kai to comment on which they think is better :).

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
+            person.Name = splittedPerson[0];
+            person.SureName = splittedPerson[1];
+            person.BirthDay = splittedPerson[2];
```
Matt500b, line 56: When you've changed the types for these properties, ensure you've updated here also!

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,24 @@
...
+        public string SureName { get; set; }
+        public string Gender { get; set; }
+        public string Country { get; set; }
+        public string Email { get; set; }
+        public string StreetAdress { get; set; }
+
+        public int CalculateAge()
+        {
+            CultureInfo culture = new CultureInfo("en-US");
+            var convertedBirthday = DateTime.Parse(this.BirthDay, culture).Year;
```
Matt500b, line 19: What if someone was born 5th May 1990 and the current date is 2nd Feb 2020. What is their age?

Really they are 29 as it's not been past 5th May. but your code will produce 30 as `2020-1990 = 30`

Probably not so much of an issue for this homework - but just food for thought :)

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,24 @@
...
+            var age = DateTime.Now.Year - convertedBirthday;
```
Matt500b, line 20: can just return directly, no need to store in a variable then return. `return DateTime.Now.Year - convertedBirthday; `

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,12 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+
+            var demo = EventManager.ManageEvents();
+            demo.Run(); 
+
```
Matt500b, line 12: Lets keep code tidy :D no random blank lines!

Remove the blank lines from line 9, 12, 13, 14

Src/BootCamp.Chapter/PeoplePredicates.cs
```diff
@@ -10,20 +10,21 @@ public static class PeoplePredicates
         /// a) over 18, who do not live in UK, whose surename does not contain letter 'a'.
         /// </summary>
         /// <returns></returns>
-        public static bool IsA(Person person) => false;
+        public static bool IsA(Person person) => person.CalculateAge() > 18 &  person.Country != "UK" & !person.SureName.Contains('a')  ;
```
Matt500b, line 13: So the question is a bit difficult but.. there are no "UK"s in the data file. It's actually `United Kingdom`. So you've never actually filtered out people from the UK.

This is where a better data type would come in handy :D. 

also, if there was `Uk` this would still fail due to case sensitivity. So you would want to use `.Equals("", StringComparison.OrdinalIgnoreCase)`

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,24 @@
...
-        // add missing properties
+        public string Name { get; set; }
+        public string BirthDay { get; set; }
+        public string SureName { get; set; }
+        public string Gender { get; set; }
+        public string Country { get; set; }
+        public string Email { get; set; }
+        public string StreetAdress { get; set; }
+
+        public int CalculateAge()
```
Matt500b, line 16: You probably wouldn't want to calculate age to be something you have to do often. Just do it once and store as a property. So you would want to call CalculateAge when you create the person object. If it was done via the constructor that would be tidier.

Src/BootCamp.Chapter/PeoplePredicates.cs
Matt500b, line 13: `&&` would be better and should really always be used unless working with bits. Read up on the differences between `&` and `&&`

Src/BootCamp.Chapter/EventManager.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        private static string MenuPredicates()
+        {
+            Console.WriteLine($"Choose on the following fiters you want to use. {Environment.NewLine}");
+            Console.WriteLine("1) Persons older then 18 which do not live in the UK and which surname does not contain the character a ");
+            Console.WriteLine("2) Persons who are younger then 18 and who surname does not contain the character a");
+            Console.WriteLine("3) Persons who do no live in the Uk and where both the name and surnmae contains the character a");
+            var choice = Console.ReadLine();
+            return choice;
```
Matt500b, line 72: Again just some tidy up, can return straight away and remove some blank lines :)

Src/BootCamp.Chapter/PeoplePredicates.cs
Matt500b: I have just reviewed the tests actually and Kai used UK in the test data so this is okay. I would still use `.Equals` for safty against case sensitivity :)

Src/BootCamp.Chapter/Person.cs
RoelofWobben: Changed the types for Birthday and Gender. 
The rest I can do when I have time

Src/BootCamp.Chapter/ContactsCenter.cs
RoelofWobben: thanks for the remark
I choose Exception because of this tests "
```
 public void New_With_Empty_File_Throws()
        {
            const string path = "Input/Empty.txt";

            Action action = () => new ContactsCenter(path);

            // Leaving this up to a student to figure out.
            action.Should().Throw<Exception>();
        }
````

otherwise I would totally agree with you

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 3 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:37]: done

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 5 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:53]: oke, then I wait on Kai remarks on this.

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 4 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:36]: changed

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 6 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:56]: done

File: Src/BootCamp.Chapter/Person.cs (hunk 7 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Person.cs:19]: oke, I have then to find a better way to calcuate the age. I hope google can help me with that.

File: Src/BootCamp.Chapter/Person.cs (hunk 8 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Person.cs:20]: done

File: Src/BootCamp.Chapter/Program.cs (hunk 9 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Program.cs:12]: done

File: Src/BootCamp.Chapter/PeoplePredicates.cs (hunk 10 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/PeoplePredicates.cs:13]: changed

Comment (RoelofWobben) [reply Src/BootCamp.Chapter/PeoplePredicates.cs:13]: oke, as far as I know && is lazy so when the first is true or false  there is a chance that the rest is not handled.  With &  every case is handled. Or I misunderstood things.

File: Src/BootCamp.Chapter/EventManager.cs (hunk 12 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/EventManager.cs:72]: done

File: Src/BootCamp.Chapter/Person.cs (hunk 11 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Person.cs:16]: Schould i change this or not.  Or schould we let Kai decide as the former remark that something could be a the constructor instead of a seperate function

File: Src/BootCamp.Chapter/Person.cs (hunk 7 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Person.cs:19]: found it

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 5 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:53]: After some thought I like your idea better then what I have now. So I have decided to do all the stuff in a person constructor.

File: Src/BootCamp.Chapter/Person.cs (hunk 11 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/Person.cs:16]: changed it because I like the code then better.

File: Src/BootCamp.Chapter/ContactsCenter.cs (hunk 2 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/ContactsCenter.cs:23]: He, when I change it to a InValidOperationException and give it a message the test are still green. Can a mentor exaplian why ? Or is it that the InvalidOperationException inherits from Exception that ir works.

File: Src/BootCamp.Chapter/PeoplePredicates.cs (hunk 10 above)
Comment (RoelofWobben) [reply Src/BootCamp.Chapter/PeoplePredicates.cs:13]: After some thought you are right that && is a better choice here. If one of them fails it has no need to look at the conditions after it. the person is already not qualified.

File: Src/BootCamp.Chapter/EventManager.cs [hunk 13]
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public class EventManager
+    {
+        public static Events ManageEvents()
```
Almantask, line 9: It's a bit odd to see this. Semantically it's correct, but logically it feels wrong. You should have events attached to an object that invokes them. Non-static object.

Almantask: Like in lessons example, I attached the events not to some events class, but to the objects that actually call them- like monster who can get hit or die. 

I think for you it makes more sense to have a Demo class to have those events.

Src/BootCamp.Chapter/EventManager.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var menuChoice = MenuPredicates();
+            var contact = new ContactsCenter(@"Input/MOCK_DATA.csv");
+            var persons = new List<Person>();
+
+            var answer = new List<Person>();
+            switch (menuChoice)
+            {
+                case "1":
+                    answer = contact.Filter(PeoplePredicates.IsA);
```
Almantask, line 45: This is not an answer, answer is actually "1". This is a filter.

Src/BootCamp.Chapter/EventManager.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            Console.WriteLine("The persons who are filterd out are");
+            foreach (var person in answer)
+            {
+                Console.WriteLine($"{person.Name} {person.SureName}");
+            }
+        }
+
+        private static string MenuPredicates()
```
Almantask, line 65: DisplayPredicateChoises is a better name. Method names should be verbs, not nouns.

Src/BootCamp.Chapter/PeoplePredicates.cs
Almantask: @RoelofWobben you did misunderstand things. Both & and && will result in the same true/false. However, && stops immediately after the left side is false. Therefore, in some weird cases, you might want run both left AND right side checks (for example if a check logs something on top of just returning true/false or maybe fires an event, etc).

Src/BootCamp.Chapter/Person.cs
Almantask, line 16: Age should be a get-only property. You can ALWAYS easily get the age of a person based on birthday.

Almantask: @Matt500b age changes based on current date-time, therefore it should be recalculated every time. for example, you don't want the same age to remain after 5 years.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,27 +1,66 @@
...
+                {
+                    people.Add(person);
+                }
+
+            }
+                   
             return people;
         }
+
+        public Person ConvertStringToPerson(string personString)
```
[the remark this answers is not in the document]
RoelofWobben: so doing this in a constructor is wrong ?   I just change the code to do this in a constructor

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: oke, so the Events class needs to be renamed to Demo or the EventManager class ?

Almantask: Get rid of the events class, it just adds not needed complexity. Move the events where they logically belong, close (or inside) the object that should invoke them.

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: oke, I will renamed answer to filter. I named this answer because I thought that it holds all the persons which are filtered out by the predicate.  But also maybe I misunderstood things

Src/BootCamp.Chapter/ContactsCenter.cs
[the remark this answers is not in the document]
Almantask: Yes, it's bad to do this in the constructor because you couple the ctor to expect something that can be split always. Give more freedom of choice. However if you do want to parse, using a tryparse pattern is fine.

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: chips,, that one I did not see. Try very hard to not make that mistake again. Changed

Src/BootCamp.Chapter/Person.cs
RoelofWobben: thanks. changed it

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: oke, then I have to think well where they belong

Src/BootCamp.Chapter/ContactsCenter.cs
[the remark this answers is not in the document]
RoelofWobben: oke, then will go back to the orginal idea and try to make a Tryparse. Have to think good how it must look like. Or I have to write one for string , one for dateTime and one for the Gender as Enum. Will sleep about that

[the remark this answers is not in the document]
RoelofWobben: Do I need to use the TryParse or is this a sort of extra challenge.

I did now this : 
```
public Person ConvertStringToPerson(string personString)
        {
            var person = new Person(); 
            CultureInfo culture = new CultureInfo("en-US");
            var splittedPerson = personString.Split(',');
            person.Name = splittedPerson[0];
            person.SureName = splittedPerson[1];
            person.BirthDay = DateTime.Parse(splittedPerson[2], culture, DateTimeStyles.None);
            person.Gender = (Gender)Enum.Parse(typeof(Gender), splittedPerson[3]);
            person.Country = splittedPerson[4];
            person.Email = splittedPerson[5];
            person.StreetAdress = splittedPerson[6];
            
            return person; 
        }
```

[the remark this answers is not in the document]
RoelofWobben: done and working

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: so if you can explain what you mean. I still do not see it. As far as I see it I did everything the same as the examples you used for the lessons.

RoelofWobben: also solved on the chat

Src/BootCamp.Chapter/EventManager.cs
RoelofWobben: no, answer is a list mqaybe I can better name it filteredData or filteredPeople.
see : 
afbeelding

RoelofWobben:
if it;s well I have made all the changes requested. If I forget one , I apolize.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,66 +1,108 @@
...
         {
             var people = new List<Person>();
-            var splittedList = _peopleFile.Split($"{Environment.NewLine}");
-            for (int i = 1; i < splittedList.Length -1 ; i++)
+            var splittedList = _peopleFile.Split(Environment.NewLine);
+            // skipped the first row because its a header.
+            for (int i = 1; i < splittedList.Length - 1; i++)
             {
-                var person = ConvertStringToPerson(splittedList[i]);
+                var person = TryParse(splittedList[i]);
```
Almantask, line 38: Still this is not the tryparse pattern you did in previous lessons.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,66 +1,108 @@
...
+            var splittedPerson = personString.Split(',');
+
+            if (String.IsNullOrEmpty(splittedPerson[0]))
+            {
+                throw new ArgumentException("name cannot be empty");
+            }
+
+            if (String.IsNullOrEmpty(splittedPerson[1]))
+            {
+                throw new ArgumentException("surename cannot be empty");
```
Almantask, line 66: All of this validation should be a part of person constructor, if it really matters. I thin you are being too strict here and most of this is not really needed.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,66 +1,108 @@
...
-        public Person ConvertStringToPerson(string personString)
+        public Person TryParse(string personString)
         {
-            var splittedPerson = personString.Split(',');
+            if (String.IsNullOrEmpty(personString))
+            {
+                throw new ArgumentException("String cannot be empty");
+            }
+
             var person = new Person();
```
Almantask, line 55: Why do you create a person over there and assign properties WAY AFTER?

Almantask: keep the code that you write close to where it's being used.

Src/BootCamp.Chapter/Demo.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    public class Demo
```
Almantask, line 5: This looks fine to me, finally! :)

Src/BootCamp.Chapter/ContactsCenter.cs
RoelofWobben: hmm, then I have to look what I have to change. The earlier tryparse were simpler for example look if a string could be parsed to a int. This is parsing a string to a object.

Src/BootCamp.Chapter/ContactsCenter.cs
RoelofWobben: oke, I will delete all the valiidation.

Src/BootCamp.Chapter/ContactsCenter.cs
RoelofWobben: oke, I wil change that

Src/BootCamp.Chapter/Demo.cs
RoelofWobben: im happy to hear

Src/BootCamp.Chapter/ContactsCenter.cs
RoelofWobben: Right now I have this : 
`` 
public Person TryParse(string personString)
        {
            
            CultureInfo culture = new CultureInfo("en-US");
            var splittedPerson = personString.Split(',');

            var isValid = DateTime.TryParse(splittedPerson[2], culture, DateTimeStyles.None, out DateTime date);
            if (!isValid)
            {
                throw new ArgumentException("date is not valid");
            }

            isValid = Enum.TryParse(splittedPerson[3], true, out Gender gender);

            if (!isValid)
            {
                throw new ArgumentException("Gender can only be Male or Female");
            }

            var person = new Person();
            person.Name = splittedPerson[0];
            person.SureName = splittedPerson[1];
            person.BirthDay = date;
            person.Gender = gender;
            person.Country = splittedPerson[4];
            person.Email = splittedPerson[5];
            person.StreetAdress = splittedPerson[6];

            return person;
        }
 ```

Almantask: you should out a person like in a tryparse pattern. This should be a part of a person class.

RoelofWobben:
oke, so you mean 
```
Person,tryParse(splittedList[i] 
```

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,27 @@
...
-using System.Text;
+using System.Globalization;
+using System.IO;
 
 namespace BootCamp.Chapter
 {
     public class ContactsCenter
     {
-        private readonly List<Person> _people;
+        private readonly string _peopleFile;
```
Almantask, line 10: This is bad. Why should a contacts center keep _peopleFile if all they need is people? Before it was good.

Almantask: So please revert to what it was with just that TryParse pattern applied.

RoelofWobben: I did delete the List part from my first commit 
I have now this : 
```
 private readonly List<Person> _people;

        public ContactsCenter(string peopleFile)
        {
            if (!File.Exists(peopleFile))
            {
                throw new FileNotFoundException("File is made up");
            }

            var contents = File.ReadAllLines(peopleFile);

            if (contents.Length == 0)
            {
                throw new InvalidOperationException("File cannot be empty");
            }

            for (int i = 1; i < contents.Length; i++)
            {
                var isValid = Person.TryParse(contents[i], out Person person);

                if (isValid)
                {
                    _people.Add(person);
                }

            }
       }
```

and will tomorrow think how I can initialize the _people list.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -31,19 +42,17 @@ public ContactsCenter(string peopleFile)
         public List<Person> Filter(Predicate<Person> predicate)
```
Almantask, line 42: Can return IEnumerable<Person> instead, because it allows assigning it to more generalized forms.

PR #242, opened 2020-04-02
Max Chapter3/homework/3

HellHunterMax:
I did not do the Extra challenge in the original Homework.

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -1,35 +1,74 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
-        {
-            return new Item[0];
-        }
+        public List<Item> Items { get; set; }
 
         public Inventory()
         {
-            _items = new Item[0];
-        }
-
-        public Item[] GetItems(string name)
-        {
-            return new Item[0];
-        }
-
-        public void AddItem(Item item)
-        {
-
+            Items = new List<Item>();
         }
 
         /// <summary>
-        /// Removes item matching criteria by item.
-        /// Does nothing if no such item exists
+        /// Returns the item it found based on the name you gave, if no item is found returns null
+        /// </summary>
+        /// <param name="name"></param>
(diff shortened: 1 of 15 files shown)


Src/BootCamp.Chapter/Shop.cs
```diff
@@ -31,6 +27,13 @@ public Item[] GetItems()
         /// </summary>
         public void Add(Item item)
         {
+            Item item1 = _inventory.GetItem(item?.Name);
+            bool isSameItem = item1?.IsTheSameItem(item) ?? false;
```
HellHunterMax, line 31: I read this as if item1 ==  null then isSameItem = false; do i read this correct? and else it checks if its the same item and returns true or false from that.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,35 +1,78 @@
...
-        /// Does nothing if no such item exists
+        /// Removes ONE item matching criteria by item.
+        /// Does nothing if no such item exists.
         /// </summary>
         public void RemoveItem(Item item)
         {
-
+            for (int i = 0; i < Items.Count; i++)
+            {
+                if (Items[i].IsTheSameItem(item ?? throw new ArgumentNullException($"{nameof(item)} can't be null.")))
```
HellHunterMax, line 70: i could change this to:
if (Items[i]?.IsTheSameItem(item ?? throw new ArgumentNullException($"{nameof(item)} can't be null.")) ?? false) 
but at the moment items cannot have a null item i think.

Src/BootCamp.Chapter/Shop.cs
Almantask: It can be even more simple: ``item1?.IsTheSameItem(item) == true``. By doing == you ask to convert nullable<bool> into bool. If it's null, it will be false.

What you wrote is also valid, but a tiny bit harder to read. It reads: try to check if item1 is matching item. It it doesn't, return false, if it does- compare.

Src/BootCamp.Chapter/Inventory.cs
Almantask: I think in this case, it's better to be a bit more simple. For homework's sake, keep it, but normally you should "fail early" therefore check the item not being null at the start and especially not at the loop.

PR #245, opened 2020-04-05
Chapter1/homework/4

vladgiurca:
Homework 4 fixed all errors! ( I think i know now, how to interpret error details! ) 
Thank you very much Kaisinnel for your support!


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -17,25 +17,26 @@ public static class Checks
         public static int PromptInt(string message)
         {
             // To do: call your implementation.
```
Almantask, line 19: The comments I added can be removed.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,140 @@
...
+                    return number;
+            }            
+        }
+
+        public static float BmiCalculator(float weight, float height)
+        {
+            bool isWeightLessOrEqualToZero = weight <= 0;
+            bool isHeightLessOrEqualToZero = height <= 0;
+
+            if (isWeightLessOrEqualToZero && isHeightLessOrEqualToZero)
```
Almantask, line 114: This is very verbose and some might consider redundant. But you have your style and to me it's definitely not too bad. I like it :)

PR #246, opened 2020-04-05
Chapter1/homework/4

DavidKatona:
Added a bunch of validation functions to Lesson4.cs. Prompt functions now return the result of these validtaion functions to ensure validity of entered user input.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -4,37 +4,25 @@ using System.Text;
 
 namespace BootCamp.Chapter
 {
-    /// <summary>
-    /// Test class is used to test your implementation.
-    /// Each homework will have a set of steps that you will have to do.
-    /// You can name your functions however you want, but to validate your solution, place them here.
-    /// DO NOT CALL FUNCTIONS FROM TESTS CLASS
-    /// DO NOT IMPLEMENT FUNCTIONS IN TESTS CLASS
-    /// TESTS CLASS FUNCTIONS SHOULD ALL HAVE 1 LINE OF CODE
-    /// </summary>
     public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
             return Lesson4.PromptInt(message);
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
             return Lesson4.PromptString(message);
         }
 
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
             return Lesson4.PromptFloat(message);
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
             return Lesson4.CalculateBMI(weight, height);
         }
     }
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -17,25 +17,25 @@ namespace BootCamp.Chapter
         public static int PromptInt(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson4.PromptInt(message);
         }
 
         public static string PromptString(string message)
         {
             // To do: call your implementation. 
-            return "";
+            return Lesson4.PromptString(message);
         }
 
         public static float PromptFloat(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson4.PromptFloat(message);
         }
 
         public static float CalculateBmi(float weight, float height)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson4.CalculateBMI(weight, height);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -17,25 +17,25 @@ public static class Checks
         public static int PromptInt(string message)
         {
             // To do: call your implementation.
```
Almantask, line 19: All comments that I wrote can be removed. It was just for you, to have a hint where to write what. Not needed after the fact you wrote something.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public static float PromptFloat(string message)
+        {
+            Console.WriteLine(message);
+            return ValidateFloat(Console.ReadLine());
+        }
+
+        public static float CalculateBMI(float weight, float height)
+        {
+            if (weight <= 0 || height <= 0)
```
Almantask, line 55: To a random reader this might pose a few questions: what does it mean?
``var isInvalidBmi = weight <= 0 || height <= 0;``
this, however, eliminates all the questions.
```cs
if(isInvalidBmi)
{
  ..
}
```

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                Console.WriteLine("Failed calculating BMI. Reason:");
+                if (weight <= 0 && height <= 0)
+                {
+                    Console.WriteLine("Weight cannot be equal or less than zero, but was {0}.", weight);
+                    Console.WriteLine("Height cannot be less than zero, but was {0}.", height);
```
Almantask, line 61: Well done for spotting this silly inconsistency. It should be consistent, but then the tests fails. It's my bad that I will fix eventually. Anyways, well done.

Src/BootCamp.Chapter/Lesson4.cs
Almantask, line 55: And I think this whole block of code could go into a function of its' own:
```cs
private static bool IsValidBmi(float weight, float height)
{
  // do validation and print
  return true if valid, else invalid;
}
```

Almantask: And then:
```cs
if(!IsValidBmi(weight, height))
{
  return -1;
}
return bmiFormula...
```

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                return InvalidNumberReturnValue;
+            }
+
+            float bodyMassIndex = weight / (MathF.Pow(height, 2.0f)); //BMI equals to kg/m2 hence the number 2
+            return bodyMassIndex;
+        }
+
+        private const string InvalidStringReturnValue = "-";
```
Almantask, line 82: These 3 should be at the top of the class.

Src/BootCamp.Chapter/Checks.cs
DavidKatona: Comments have been removed as suggested.

DavidKatona: Comments have been removed as suggested.

Src/BootCamp.Chapter/Lesson4.cs
DavidKatona: Yes, you're exactly right. :) I've changed the code so now it validates the entered values (weight, height) in a separate function called IsValidBMI. It returns a bool and inside the CalculateBMI method, I only check for the returned value now and do the calculation if the validation succeeds. That should improve readability. One think I wasn't sure about is where to put the Console.WriteLine stuff but since the rest of the code had those in the validate functions, I follow the same pattern for BMI as well.

DavidKatona: New function has been added and responsiblities have been separated as described in the previous comment above.

Src/BootCamp.Chapter/Lesson4.cs
DavidKatona: I was scratching my head for at least 10-15 minutes but at least I got to know more about the tests you wrote :)

Src/BootCamp.Chapter/Lesson4.cs
DavidKatona: Constants have been moved to the top of the class as suggested.

Src/BootCamp.Chapter/Lesson4.cs
Almantask: for now it's okay, because those writeline thingies are more for logging rather than display (UI). Normally, you have a separation between the two.

Almantask: About that later in the lessons, so don't bother yourself with it yet ;)

PR #247, opened 2020-04-05
Chapter1/homework/1 and2

TheBleydel:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,32 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            //Homework 1 and 2 
+            //1
+            string name, surename;
```
Almantask, line 11: Do not declare all local variables in one place. It's a bad practice, because this makes you switching from one context to another (from assignment back to declaration). This makes it harder to remember what the types were, forces us to keep track of a bigger code context.

Declare the variables as close as where you use them.

Almantask:
Changes for variable declaration as close as possible to where you use them is needed.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -7,26 +7,31 @@ class Program
         static void Main(string[] args)
         {
             //Homework 1 and 2 
-            //1
-            string name, surename;
-            int age;
-            double weight, height, heightMeters, bmi;
-
+          
+            string name;
```
Almantask, line 11: You don't have to declare variables in separate lines. It's better for readaiblitiy if you assign and declare it at the same line, if the assignment is simple, which in this case  it is.

PR #248, opened 2020-04-05
Chapter 1/Homework #2

MikeChisowda:
Basically for this homework I created several varibles that later on will be used to another person without creating new ones. The program asks for the person's name, surname, weight, etc... All with it's proper variables (or at least I tried to do so haha) e.g. float for weight and height and did some (to be fair I don't know how they're called) "transfomation" so the answer of the ReadLine could be saved into a float variable. If there's anything wrong be sure to let me know it!! :)

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,61 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            // READ ONE PERSON'S STUFF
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,61 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            // READ ONE PERSON'S STUFF
```
Almantask, line 9: Good comments for separating blocks of code

PR #251, opened 2020-04-06
Chapter3 / Homework 7

RoelofWobben:
I hope it did it allright


Almantask:
> I hope it did it allright

Maybe. But you did not write a good PR description, it shouldn't be about your hopes, it should be about the work that you have done.

RoelofWobben:
Can also change it but do I then not also copy the homework description.

Src/BootCamp.Chapter/LinqDemo.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    internal class LinqDemo
+    {
+        internal static void Demo()
+        {
+            var list1 = new List<int> { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
+            var list2 = new List<int> { 1, 11, 12 };
+            var list3 = new List<int> { 4, 2, 5, 4 };
+
+            // Any. so Im going to check if any number in list1 is even.
```
Almantask, line 15: It would be a bit better if you had those examples each in their own methods.
They are unrelated blocks of code, which can exist by themselves, so might as well be placed in different places. Your 3 lists will be needed to be moved to private fields.

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        internal static IEnumerable<T> Demo<T>(this IEnumerable<T> list, Predicate<T> predicate)
+        {
+            list = Filter(list, predicate);
+            list = ReduceToHalfLength(list);
+
+            return list; 
+        }
+
+        private static List<T> ReduceToHalfLength<T>(IEnumerable<T> list)
```
Almantask, line 18: ``RemoveHalfElementsAtRandom`` is a more accurate name.

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var indexOfHalfList = Math.Floor((decimal) list.Count() / 2);
```
Almantask, line 20: decimal is overkill, make it float.

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
...
+            // I start the loop one after the half of the list so I can remove all the items needed 
+            for (int i = 0 ; i < indexOfHalfList; i++)
+            {
+                halfList.Add(convertedToList[i]); 
+            }
+
+            return halfList; 
+        }
+
+        private static IEnumerable<T> Filter<T>(IEnumerable<T> list, Predicate<T> predicate)
```
Almantask, line 32: This method is not needed. It's done for you in LINQ. You can simply:
``list.Where(predicate)``. So remove this method and use the suggested LINQ instead.

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public static class LinqMethodDemo
```
Almantask, line 8: This is not a demo- it's an actual extension. I would call it IEnumerableExtensions

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        internal static IEnumerable<T> Demo<T>(this IEnumerable<T> list, Predicate<T> predicate)
```
Almantask, line 10: I wouldn't call this demo, but ``SnapFingers`` or RemoveHalfOfAllElementsAtRandom()

Src/BootCamp.Chapter/LinqMethodDemo.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            list = Filter(list, predicate);
```
Almantask, line 12: Both methods are not needed and all the implementation can be moved to this methods (without private any private step functions)

Src/BootCamp.Chapter/ShuffleDemo.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    internal static class ShuffleDemo
```
Almantask, line 7: This is supposed to be named IEnumerable extensions and in fact moved to the same class as other extension.

Src/BootCamp.Chapter/ShuffleDemo.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        internal static void Demo()
```
Almantask, line 9: Remove it- not a part of this class. It's a part of other class where you demo it. You couple too much here.

Src/BootCamp.Chapter/ShuffleDemo.cs
```diff
@@ -0,0 +1,37 @@
...
+            var shuffledList = ShuffleColllectionExtension(new List<int> { 1, 2, 3, 4, 5 });
+            Console.Write("The shuffled List looks like this: ");
+            foreach (var item in shuffledList)
+            {
+                Console.Write($"{ item} ");
+            }
+            Console.WriteLine(Environment.NewLine); 
+        }
+
+        private static IEnumerable<T> ShuffleColllectionExtension<T>(this IEnumerable<T> list)
```
Almantask, line 20: Should be public and like that demonstrate that you can call this as an extension.

Src/BootCamp.Chapter/ShuffleDemo.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    internal static class ShuffleDemo
+    {
+        internal static void Demo()
+        {
+            var shuffledList = ShuffleColllectionExtension(new List<int> { 1, 2, 3, 4, 5 });
```
Almantask, line 11: This is a demo for static method, but we asked you to demo extension methods. So it's completely wrong. Also, why do you pass a new list?

Src/BootCamp.Chapter/ShuffleDemo.cs
Almantask: Demonstrate that you understand what extension methods are for. Here you demonstrate that you misunderstand the difference between a static class and extension methods.

Src/BootCamp.Chapter/ShuffleDemo.cs
Almantask: You should shuffle the list that you created. Not pass it.

Src/BootCamp.Chapter/LinqDemo.cs
RoelofWobben: sorry, this confuses me a little. First you said that they schould be in their own methods and later on you say they need to be private fields.

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: oke, changed but right now I do not delete atRandom. I delete all items after the half of the collection

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: changed

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: oke
then I have to look how to solve this error message : 
```
Argument 2: cannot convert from 'System.Predicate<T>' to 'System.Func<T, bool>'	
```

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: done

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: so I do not lt you see that the extensions are working or do I misunderstood you here complete

Src/BootCamp.Chapter/ShuffleDemo.cs
RoelofWobben: done

Src/BootCamp.Chapter/ShuffleDemo.cs
RoelofWobben: oke, so if I have put the extension in the right class , I still have to delete that one.
or do you want one demo class where I call all the three/four demo's

Src/BootCamp.Chapter/ShuffleDemo.cs
RoelofWobben: oke, I have to think what you want here exactly. Sometimes you are very cryptic and I do then not see what you mean

Src/BootCamp.Chapter/LinqDemo.cs
Almantask: Private fields fro the shared input- the 3 lists. It would be annoying to pass those 3 9 times.

Src/BootCamp.Chapter/LinqMethodDemo.cs
Almantask: It should be random as asked in the task.

Src/BootCamp.Chapter/LinqMethodDemo.cs
Almantask: Then that is your solution, the second one. Though I find it strange thst it is not a predicate.

Src/BootCamp.Chapter/LinqMethodDemo.cs
Almantask: Refer to comments below.

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: Then I overread it several times. 
I do not see anything about random here : 
``` 
 Make your own LINQ method! Create a method SnapFingers which works like LINQ methods, 
takes a predicate and removes exactly half of all the elements in the collection.
If it's not even number, it should remove 1 less.  
For example:  
if it's 2, it removes 1;  
if it's 4, it removes 2;  
but if it's 3, it removes 1
```

Src/BootCamp.Chapter/ShuffleDemo.cs
Almantask: Extension has nothing to do with demo.

You always need at least 2 classes: 
One for demo
One for implementation 

By implementing it in the same class as demo you couple demo with the thing it demos.

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: me too. but on the first one I see a error message.

Src/BootCamp.Chapter/LinqDemo.cs
RoelofWobben: hmm, can I use a private field and still have the numbers as contents?

Src/BootCamp.Chapter/ShuffleDemo.cs
RoelofWobben: oke, will do that

Src/BootCamp.Chapter/ShuffleDemo.cs
RoelofWobben: oke, will work on that too

Src/BootCamp.Chapter/LinqMethodDemo.cs
RoelofWobben: resolved

Src/BootCamp.Chapter/LinqDemo.cs
RoelofWobben: solved

Src/BootCamp.Chapter/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    public static class IEnumerableExtensions
+    {
+        public static IEnumerable<T> SnapFingers<T>(this IEnumerable<T> list, Predicate<T> predicate)
+        {
+            var list2 = list.Where(x => predicate(x)).ToList();
```
Almantask, line 11: BEAUTIFUL!

Src/BootCamp.Chapter/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            var list3 = list2.RemoveHalfElementsAtRandom();
+
+            return list3; 
+        }
+
+        public static List<T> RemoveHalfElementsAtRandom<T>(this IEnumerable<T> list)
+        {
+            var indexOfHalfList = Math.Floor((float)list.Count() / 2);
+            var convertedToList = list.ToList();
```
Almantask, line 20: You don't have to convert to list. It's not so obvious, but you can foreach every element, increment index i and break when i reaches half of total.

It's fine for now, I am sure you can do it, so no worries.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using System.Collections.Generic;
+
+internal class Lesson3
+{
+    private static List<int> list1 = new List<int> { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
+
+    internal static void Demo()
+    {
+        //Demo the Shuffle method
```
Almantask, line 11: Each of those demos, instead of a comment, can be a private method. Slightly more cleaner.

Src/BootCamp.Chapter/LinqDemo.cs
```diff
@@ -6,55 +6,75 @@ namespace BootCamp.Chapter
...
+            IntersectionDemo(list1, list2);
+            ExtractDemo(list1, list2);
+        }
 
-            // Count. So we are going to count how many even numbers there are in list1. 
-            var count = list1.Count(n => n % 2 == 0);
-            Console.WriteLine($"There are {count} even numbers in list1{Environment.NewLine}");
+        private static void ExtractDemo(List<int> list1, List<int> list2)
+        {
+            Console.Write($"The intersection of a list and another list looks like this: ");
```
Almantask, line 26: The demos are not very high effort. You copy pasted writes and forgot to even update the messages.

Almantask: In the future try to come up with more specific concepts: like people and pets or similar. Now it's just a meaningless list.

PR #253, opened 2020-04-07
Chapter1 homework 1

JayofBlu3:
Fixed problems and added second person

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -8,27 +8,39 @@ namespace BootCamp.Chapter
         {
             Console.WriteLine("Hi! I will calculate your BMI. I just need to ask a few questions. What is your full name?");
             string fullname = Console.ReadLine();
-            Console.WriteLine("How old are you?");
+
+            Console.WriteLine("How old are you?(only numbers please)");
             int age = Convert.ToInt32(Console.ReadLine());
+
             Console.WriteLine("How much do you weigh in kg?");
             double weight = Convert.ToDouble(Console.ReadLine());
+
             Console.WriteLine("What is your height in cm?");
             double heightcm = Convert.ToDouble(Console.ReadLine());
-            double heightm = heightcm / 100;
-            double heightsq = heightm * heightm;
+            double heightsq = (heightcm / 100) * (heightcm/100);
+            double bmi = weight / heightsq;
+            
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,36 @@ class Program
...
+            Console.WriteLine("How old are you?");
+            int age = Convert.ToInt32(Console.ReadLine());
+            Console.WriteLine("How much do you weigh in kg?");
+            double weight = Convert.ToDouble(Console.ReadLine());
+            Console.WriteLine("What is your height in cm?");
+            double heightcm = Convert.ToDouble(Console.ReadLine());
+            double heightm = heightcm / 100;
+            double heightsq = heightm * heightm;
+            Console.WriteLine("Great! Now I will calculate your BMI.");
+            Console.WriteLine("Your name is " + fullname + ", you are " + age + " years old, you weigh " + weight + " kgs, and you are " + heightcm + " cm tall. your BMI is " + weight / heightsq + ".");
```
Almantask, line 20: To make it more simple, you should separate calculation of bmi from displaying it (so 2 lines instead of 1)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,36 @@ class Program
...
-            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hi! I will calculate your BMI. I just need to ask a few questions. What is your full name?");
+            string fullname = Console.ReadLine();
+            Console.WriteLine("How old are you?");
+            int age = Convert.ToInt32(Console.ReadLine());
+            Console.WriteLine("How much do you weigh in kg?");
+            double weight = Convert.ToDouble(Console.ReadLine());
+            Console.WriteLine("What is your height in cm?");
+            double heightcm = Convert.ToDouble(Console.ReadLine());
+            double heightm = heightcm / 100;
```
Almantask, line 17: I like that you made the distinction in m. You could have done this in the same line (above) instead.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,36 @@ class Program
...
+            double heightsq = heightm * heightm;
```
Almantask, line 18: Unlike other variables, this seems unnecessary. Just to to put bmi into a formula here. I would recommend renaming this variable to bmi and calculating it here.

PR #257, opened 2020-04-07
Chapter1/homework/3

MikeChisowda:
Did the functions that were asked (height and weight, age, whole name,etc.) and also did a function that used these mentioned to print a message with their personal info and BMI.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,26 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            int age = Lesson3.Age(message);
+            return age;
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
+            string wholename = Lesson3.WholeName(message); 
+            return wholename;
         }
 
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            float heightAndWeight=Lesson3.HeightAndWeight(message);
+            return heightAndWeight;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            float calculateBmi = Lesson3.CalculateBMI(weight,height);
+            return calculateBmi;
         }
     }
 }
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,26 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            int age = Lesson3.Age(message);
+            int age = Lesson3.PromptInt(message);
             return age;
         }
 
         public static string PromptString(string message)
         {
-            string wholename = Lesson3.WholeName(message); 
-            return wholename;
+            string fullname = Lesson3.PromptString(message); 
+            return fullname;
         }
 
         public static float PromptFloat(string message)
         {
-            float heightAndWeight=Lesson3.HeightAndWeight(message);
+            float heightAndWeight=Lesson3.PromptFloat(message);
             return heightAndWeight;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            float calculateBmi = Lesson3.CalculateBMI(weight,height);
-            return calculateBmi;
+            float bmi = Lesson3.CalculateBMI(weight,height);
+            return bmi;
         }
     }
 }
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -10,32 +10,31 @@ namespace BootCamp.Chapter
 
         public static void Demo()
         {
-            PersonalInfo();
-            PersonalInfo();
+            PromptPersonalInfo();
+            PromptPersonalInfo();
         }
 
-        public static void PersonalInfo(){
-
-            string name = WholeName(message: "Input your name: ");
-            string surname = WholeName(message: "Input your surname: ");
-            int age = Age(message: "Input your age: ");
-            float height = HeightAndWeight(message: "Input your height in meters: ");
-            float weight = HeightAndWeight(message: "Input your weight in kg: ");
+        public static void PromptPersonalInfo()
+        {
+            string name = PromptString(message: "Input your name: ");
+            string surname = PromptString(message: "Input your surname: ");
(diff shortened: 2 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,26 @@ public static class Checks
...
-            return "";
+            string wholename = Lesson3.WholeName(message); 
+            return wholename;
         }
 
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            float heightAndWeight=Lesson3.HeightAndWeight(message);
```
Almantask, line 31: Please be consistent with your spacing :)

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,26 @@ public static class Checks
...
-            return 0;
+            int age = Lesson3.Age(message);
+            return age;
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
+            string wholename = Lesson3.WholeName(message);
```
Almantask, line 25: it's called fullname, not wholename.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,26 @@ public static class Checks
...
-            return 0;
+            float heightAndWeight=Lesson3.HeightAndWeight(message);
+            return heightAndWeight;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            float calculateBmi = Lesson3.CalculateBMI(weight,height);
```
Almantask, line 37: calculateBmi is a function name. The results (variables) should be named after nouns. 
A simple name ``bmi`` would be enough.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+    class Lesson3
+    {
+
+        public static void Demo()
+        {
+            PersonalInfo();
+            PersonalInfo();
+        }
+
+        public static void PersonalInfo(){
```
Almantask, line 17: A function should be named after some action because they do something. So a verb.
``PromptPersonalInfo`` sounds better

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+            int age = Age(message: "Input your age: ");
+            float height = HeightAndWeight(message: "Input your height in meters: ");
+            float weight = HeightAndWeight(message: "Input your weight in kg: ");
+            float bmi = CalculateBMI(weight: weight, height: height);
+
+            Console.WriteLine(name + " " + surname + " " + "is" + " " + age + " years old, his weight is " + weight + " kg and his height is " + height + " m and his/her BMI is: " + bmi);
+
+        }
+
+        public static string WholeName(string message)
```
Almantask, line 30: WholeName should be renamed to PromptString, because it doesn't matter what you get (a name, brand, nickname, etc). It's just a contactless string. It gets meaning only when is assigned to a variable. And in your case the variables are named right.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            Console.WriteLine(message);
+            string partOfName = Console.ReadLine();
+
+            return partOfName;
+        }
+
+        public static float HeightAndWeight(string message)
```
Almantask, line 38: Would it work for amount of money or speed? Yes! Don't be so specific when naming methods, if you can generalize the name, please do. This PromptsFloat. Prompt- because it asks to input something, float, because it converts the input to float and returns.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            Console.WriteLine(message);
+            float value = float.Parse(Console.ReadLine(), CultureInfo.InvariantCulture);
```
Almantask, line 41: Pretty cool to see ``CulterInfo.Invariant``. It's not needed to have it in this case, but it's not a mistake either. Could you explain why did you use it?

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            return value;
+        }
+
+        public static int Age (string message)
```
Almantask, line 46: PromptInt is a better name.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            Console.WriteLine(message);
+            int age = Convert.ToInt32(Console.ReadLine());
+
+            return age;
+        }
+
+        public static float CalculateBMI(float weight, float height)
+        {
+
```
Almantask, line 56: Blank line is not needed.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,62 @@
...
+            int age = Convert.ToInt32(Console.ReadLine());
+
+            return age;
+        }
+
+        public static float CalculateBMI(float weight, float height)
+        {
+            
+            float bmi = weight / (height * height);
+
```
Almantask, line 58: Blank line is not needed.

Src/BootCamp.Chapter/Checks.cs
MikeChisowda: Changed the value "wholename" to "fullname".

Src/BootCamp.Chapter/Checks.cs
MikeChisowda: Tried to remove inconsistent spacing.

Src/BootCamp.Chapter/Checks.cs
MikeChisowda: Changed "calculateBmi" to bmi.

Src/BootCamp.Chapter/Lesson3.cs
MikeChisowda: Super! I'll try to implement those advices.

Src/BootCamp.Chapter/Lesson3.cs
MikeChisowda: Done!

Src/BootCamp.Chapter/Lesson3.cs
MikeChisowda: Done!

PR #264, opened 2020-04-09
Chapter1/homework/5

DavidKatona:
I struggled a lot more with this one than the previous homeworks. I think I tried to "overengineer" it by adding a couple of "generalised" functions for validating null or empty arrays, and so on, which might have backfired at me as I progressed. In the end, I'm not a 100% satisfied with the outcome but I'd like to see your point on this.

There are some functions which do a "double" validation like the "RemoveFirst" or "RemoveLast" functions, as they just call the "RemoveAt" function at a given index. Since all of these functions had tests which had to pass "null" and "empty" array scenarios, I had to add these to all 3 which means, if I call "RemoveFirst" it validates the array, then goes into the "RemoveAt" function and does to validation again.

I've also left the "summaries" of the functions in because I wasn't sure whether you'd count them as regular comments like the ones we had in previous exercises or not.

(edit: I should also commit more often, at least once per implementation)

### DIFF
--- a/Src/BootCamp.Chapter1/ArrayOperations.cs
+++ b/Src/BootCamp.Chapter1/ArrayOperations.cs
@@ -72,9 +72,9 @@ namespace BootCamp.Chapter1
         /// <returns>A new array with element removed at a given index. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveAt(int[] array, int index)
         {            
-            if (!IsArrayNullOrEmpty(array) && !IsIndexOutOfRange(array, index))
+            if (!IsArrayNullOrEmpty(array) && !IsIndexOutOfRange(array, index, false))
             {
-                int bufferSize = array.Length - 1;
+                int bufferSize = array.Length - 1; //Since we remove an element we want to increase the size of the array by 1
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -19,7 +24,15 @@ public static void Sort(int[] array)
         /// <param name="array">Input array in a random order.</param>
         public static void Reverse(int[] array)
         {
-            // ToDo: implement.
+            if (!IsArrayNullOrEmpty(array))
+            {
+                for (int i = 0; i < array.Length / 2; i++)
+                {
+                    int buffer = array[i];
```
Almantask, line 31: these 3 lines are repeated in sort as well. Could you use a private function for both? :)

DavidKatona: Yes, there could have been a SwapElements function or something like that that switched 2 elements at given indices. Also, for the buffer creation, I could have made another function which would have had the option to expand or shrink the buffer (allowing reusability inside the remove and insert functions).

PR #266, opened 2020-04-09
Chapter1/homework/3

Smephey:
Submission of homework for lesson 3. Refactored lesson 1 into a class of its own, using functions rather than just a large block of code in Demo().

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,26 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation
+            return Lesson3.GetInt(message);
         }
 
         public static string PromptString(string message)
         {
             // To do: call your implementation. 
-            return "";
+            return Lesson3.GetString(message);
         }
 
         public static float PromptFloat(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson3.GetFloat(message);
         }
 
         public static float CalculateBmi(float weight, float height)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson3.CalculateBMI(height, weight);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,26 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation
```
Almantask, line 19: Comments can now be removed.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,60 @@
...
+            Console.WriteLine("Their BMI is: " + bodyMassIndex);
+        }
+
+        public static string GetString(string message)
+        {
+            Console.WriteLine(message);
+            return Console.ReadLine();
+        }
+
+        public static int GetInt(string message)
```
Almantask, line 31: PromptInt would be a better name, because you ask to get an int, you don't just get it.

Almantask: Promp also suggests that you get it from some sort of a terminal (console)

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,60 @@
...
+            string surname = GetString("What's your surname?");
+            int age = GetInt("How old are you?");
+            float weight = GetFloat("How much do you weight in kg?");
+            float height = GetFloat("How tall are you in cm?");
+            float mHeight = ConvertToMetres(height);
+            float bodyMassIndex = CalculateBMI(mHeight, weight);
+
+            Console.WriteLine(name + " " + surname + " is " + age + " years old, his weight is " + weight + "kg and his height is " + height + " cm");
+
+            Console.WriteLine("Their BMI is: " + bodyMassIndex);
```
Almantask, line 22: Great separation

Almantask: However 2nd person is missing.

PR #267, opened 2020-04-09
Chapter1/Homework/4

MikeChisowda:
I tried to do all the functions, commented what the part of the code would do and tried to be extra wary with format issues. Nonetheless, some of the tests failed and I know the last part (the function of CalculateBmi) might have a better solution but I couldn't find any other.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,88 +16,22 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            //Input
-            Console.WriteLine(message);
-            string number = Console.ReadLine();
-
-            //Check if the string is empty
-            if (string.IsNullOrEmpty(number)) return 0;
-
-            //Check if the string is a number
-            if (!int.TryParse(number, out var wholeNumber))
-            {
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,22 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Program.PromptInt(message);
```
Almantask, line 19: One of the few people who removed the comment- well done.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +9,107 @@ class Program
     {
         static void Main(string[] args)
         {
+            PromptPersonalInfo();
```
Almantask, line 12: Well done splitting the whole person thing into methods.

Almantask:
The only nitpick I have is the tests. The tests are failing. It's up to you to fix it or not, it fails not because the code is bad, but because the requirements are a bit weird. Since you did it so well, I will request a change :) You can definitely do it! Try running tests locally in your vs and refer to error messages after the tests failed. Pay the most attention to newline symbols, Write and Writeline and they are required inconsistenly.

PR #271, opened 2020-04-10
Used functions as instructed (hopefully) and did the deed.

MyAxes:
I would really appreciate it if you could give some suggestions as to how I could improve this further.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -8,23 +8,23 @@ namespace BootCamp.Chapter
     {
         public static void Demo()
         {
-             static void GatherInfo(string userNumber)
-            {
-                Console.WriteLine("\r\nHello! You are the " + userNumber + " user today.");
-                string firstName = PromptString("What is your first name?");
-                string surName = PromptString("What is your surname?");
-                int age = PromptInt("What is your age?");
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            string firstName = PromptString("What is your first name?");
+
```
Almantask, line 12: A blank line after every line is not needed. It did make sense before you didn't use functions (to isolate a function-like block), but now your functions make it readable enough.

The only bit that I would separate is reads from bmi calculation.

Also, the contents of demo could go into a function of its' own and you're missing person 2

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -8,19 +8,21 @@ class Lesson3
     {
         public static void Demo()
         {
-            string firstName = PromptString("What is your first name?");
+             static void GatherInfo(string userNumber)
```
Almantask, line 11: Don't use nested functions, move this outside of demo function, into the class scope. This makes code less maintainable because we're forced to read about all the details (function calls and implementation are in one place).

PR #272, opened 2020-04-10
Chapter1/homework/6

DavidKatona:
Added solutions for part 1 and 2 to "Solutions.txt". To me the challange seems to use a merge sort algorhythm though I could not figure it out how to debug it. :)


Almantask:
> Added solutions for part 1 and 2 to "Solutions.txt". To me the challange seems to use a merge sort algorhythm though I could not figure it out how to debug it. :)

As in so many cases, in order to use and work with stuff, we do not need to fully understand it. You manage to solve the problem and do what is asked without that knowledge and we hope that this will be a useful hint- not to overdo thinks and not to overcomplicate and try to 100% nail everything. It's a nice to have, but certainly not a must.

Keep the good job up 👍

PR #273, opened 2020-04-11
Chapter1/Homework/1And2

jeanroyce:
Homework #1: BMI Calculator

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,41 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            bool again = true;
+            do
+            {
+                // Get info for BMI calculator
+                Console.WriteLine("Body-mass index (BMI) calculator");
+                Console.WriteLine("What is your first name?");
+                string name = Console.ReadLine();
+                Console.WriteLine("What is your surname?");
+                string surName = Console.ReadLine();
+                Console.WriteLine("How old are you?");
+                int age = Int32.Parse(Console.ReadLine());
+                Console.WriteLine("What is your weight(Kg)?");
+                float weight = float.Parse(Console.ReadLine());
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,41 @@ class Program
...
+                string surName = Console.ReadLine();
+                Console.WriteLine("How old are you?");
+                int age = Int32.Parse(Console.ReadLine());
+                Console.WriteLine("What is your weight(Kg)?");
+                float weight = float.Parse(Console.ReadLine());
+                Console.WriteLine("What is your height(Cm)?");
+                float height = float.Parse(Console.ReadLine());
+
+                // BMI calculation
+                float bodyMassIndex = weight / ((height / 100) * (height / 100));
```
Almantask, line 26: I like the separation of input and calculation- well done!

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,41 @@ class Program
...
+            do
+            {
+                // Get info for BMI calculator
+                Console.WriteLine("Body-mass index (BMI) calculator");
+                Console.WriteLine("What is your first name?");
+                string name = Console.ReadLine();
+                Console.WriteLine("What is your surname?");
+                string surName = Console.ReadLine();
+                Console.WriteLine("How old are you?");
+                int age = Int32.Parse(Console.ReadLine());
```
Almantask, line 19: Int32 can be simplified to just int- it's a relict of the old times.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,41 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            bool again = true;
```
Almantask, line 9: Logical variables should be prefixed with is/should/are/was and so on.
``isAgain`` is a bit better for a name.

PR #275, opened 2020-04-11
Chapter1/homework/4

Smephey:
Submitting homework 4 solution. I'm sure the BMI validation could be done in a better way but couldn't quite find it.


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            Console.WriteLine("Their BMI is: " + friendBodyMassIndex);
+        }
+
+        public static string PromptString(string message)
+        {
+            Console.WriteLine(message);
+            string response = Console.ReadLine();
+
+            if (string.IsNullOrEmpty(response))
```
Almantask, line 41: In programming there is a principle to "Fail Early". Or "Finish early", "Exit early". This means, that every function should start with validation and return the result (or error) as soon as it can. Normally, validation is the easiest/quick way of getting the result, therefore you first validate stuff, return error/value if it was bad and only after that do the actual calculation. This makes us care about less things in a function and everything follows a similar pattern (more consistent)

PR #277, opened 2020-04-12
Chapter1/homework/4

MyAxes:
I fudged up the PR cuz i''m a dumb. Don't mind the many commits that were unnecessary.

### DIFF
--- a/Src/BootCamp.Chapter/Homework4.cs
+++ b/Src/BootCamp.Chapter/Homework4.cs
@@ -8,22 +8,23 @@ namespace BootCamp.Chapter
     {
         public static void Demo()
         {
-            static void GatherInfo(string userNumber)
-            {
-                Console.WriteLine("Hello! You are the " + userNumber + " user today.");
-                string firstName = PromptString("What is your first name?");
-                string surName = PromptString("What is your surname?");
-                int age = PromptInt("What is your age?");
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/Homework4.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Homework4
+    {
+        public static void Demo()
+        {
+            static void GatherInfo(string userNumber)
```
Almantask, line 11: This should not be nested in Demo method. Move it to Homework4 class. this makes code less readable, because you force to keep step function and implementation of a function that step is for to be in the same place. Which beats the purpose of adding a step function (other than reusability)

Src/BootCamp.Chapter/Homework4.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+            GatherInfo("first");
+            GatherInfo("second");
+        }
+        public static string PromptString(string message)
+        {
+            Console.WriteLine(message);
+            string Name = Console.ReadLine();
+
+            if (string.IsNullOrEmpty(Name))
```
Almantask, line 32: This should happen first and only then do the calculation. It's called "Fail Early" principle. makes programmer care about less things later down the line.

PR #278, opened 2020-04-12
Chapter 1/ Homework 3

ZacheMerridew:
My submission for Lesson 3 HW.


Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +9,63 @@ class Program
     {
         static void Main(string[] args)
         {
+            PersonOneBMI();
```
Almantask, line 12: Do you need two functions :) ?
Two calls- yes, but the same function can be reused!

PR #282, opened 2020-04-12
Chapter1 homework 1&2

joshuajz:
I tried to submit this earlier but then realized that you may want me to use input from the user.  Alas I think this is what you're looking for...


README.md
```diff
@@ -1,4 +1,63 @@
-# CSharp-From-Zero-To-Hero
-[![Build Status](https://dev.azure.com/almantusk/From-Zero-To-Hero/_apis/build/status/Almantask.CSharp-From-Zero-To-Hero?branchName=master)](https://dev.azure.com/almantusk/From-Zero-To-Hero/_build/latest?definitionId=5&branchName=master)
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/a358adf21c2442cd91a4827a50213ff1)](https://www.codacy.com/manual/Almantask/CSharp-From-Zero-To-Hero?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=Almantask/CSharp-From-Zero-To-Hero&amp;utm_campaign=Badge_Grade)  
-C# boot camp
+# C#: From Zero To Hero
```
Almantask, line 1: As you can see, you're trying to merge to csinn/master. However, you should be merging with csinn/Chapter1/Homework/1And2, otherwise it will be harder to tell what changes you made compared to the original homework branch.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,18 @@ class Program
...
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name1 = "Tom Jefferson";
+            int age1 = 19;
+            int weight1 = 50;
+            float height1 = 156.5f;
+            float bmi1 = weight1 / height1 / height1 * 10000;
+            Console.WriteLine($"My name is {name1} and I am {age1} years old, my weight is {weight1} kg, I'm {height1} meters tall and my BMI is {bmi1}");
+            string name2 = "Tom Holland";
```
Almantask, line 15: This code is hard to read, it looks like a blob of code. You can play a bit smarter and use blank lines separating the two people, input gathering and bmi calculation. Also comments would be nice.

Lastly, please don't hardcode the input, but rather use 
``int age1 = int.Parse(Console.ReadLine());``
for getting input (ReadLine) and converting it to the type you need (int.Parse).

PR #285, opened 2020-04-12
Chapter1/homework/4

Shareni:
Passing the tests was far more confusing than actually coding the solution, but it was quite enlightening in how to read the tests, errors, differences between Console.WriteLine and Console.Write, etc. 

I wasn't sure if it was cleaner to separate  var isNumber... from if (!isNumber) so I left it together because it seemed easier to read than separated.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Runtime.Intrinsics.X86;
 using System.Text;
 
 namespace BootCamp.Chapter
@@ -30,14 +31,16 @@ namespace BootCamp.Chapter
 
         public static float PromptFloat(string message)
         {
+            float float1 = Lesson3.GetFloat(message);
             // To do: call your implementation. 
-            return 0;
+            return float1;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
+            float bmi1 = Lesson3.Bmi(weight, height);
             // To do: call your implementation. 
-            return 0;
+            return bmi1;
         }
     }
 }
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,14 +16,16 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
+            int intTest = Lesson3.GetInt(message);
             // To do: call your implementation. 
-            return 0;
+            return intTest;
         }
 
         public static string PromptString(string message)
         {
+            string string1 = Lesson3.GetString(message);
             // To do: call your implementation. 
-            return "";
+            return string1;
         }
 
         public static float PromptFloat(string message)
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -17,30 +17,26 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            int intTest = Lesson3.GetInt(message);
-            // To do: call your implementation. 
-            return intTest;
+            int number = Lesson3.PromptInt(message);
+            return number;
         }
 
         public static string PromptString(string message)
         {
-            string string1 = Lesson3.GetString(message);
-            // To do: call your implementation. 
-            return string1;
+            string text = Lesson3.PromptString(message);
+            return text;
         }
 
         public static float PromptFloat(string message)
         {
-            float float1 = Lesson3.GetFloat(message);
-            // To do: call your implementation. 
-            return float1;
+            float number = Lesson3.PromptFloat(message);
+            return number;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            float bmi1 = Lesson3.Bmi(weight, height);
-            // To do: call your implementation. 
-            return bmi1;
+            float bmi = Lesson3.CalculateBmi(weight, height);
+            return bmi;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,30 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
+            int intTest = Lesson3.GetInt(message);
             // To do: call your implementation.
```
Almantask, line 21: Comment is no longer needed.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,30 @@ public static class Checks
...
         {
+            int intTest = Lesson3.GetInt(message);
             // To do: call your implementation. 
-            return 0;
+            return intTest;
         }
 
         public static string PromptString(string message)
         {
+            string string1 = Lesson3.GetString(message);
```
Almantask, line 27: Since there is no other string in this context, it should be just string. However, naming variables after types is no good, therefore you should make it more human-understable. You read text- it should be text.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,30 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
+            int intTest = Lesson3.GetInt(message);
```
Almantask, line 20: it's not intTest but a number. Please rename

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,30 @@ public static class Checks
...
         {
+            string string1 = Lesson3.GetString(message);
             // To do: call your implementation. 
-            return "";
+            return string1;
         }
 
         public static float PromptFloat(string message)
         {
+            float float1 = Lesson3.GetFloat(message);
```
Almantask, line 34: number instead of float1

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static float IsNumber(string input)
+        {
+            var isEmpty = String.IsNullOrEmpty(input);
```
Almantask, line 13: You can probably inline this with the line bellow- little to do still is reasonably readable.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,126 @@
...
+           
+            if (isEmpty) return 0f;
+            
+            if ((input) == "0") return 0f;
```
Almantask, line 17: This is not needed, because Parse part covers parsing "0" as well.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,126 @@
...
+            Console.WriteLine(message);
+            
+            string float1 = Console.ReadLine();
+            
+            var isFloat = IsNumber(float1);
+           
+            return isFloat;
+        }
+
+        public static float Bmi(float weight, float height)
```
Almantask, line 71: Method names are not things- they are actions. Not nouns, but verbs.
Therefore, this method is not ``Bmi``, it is ``CalculateBmi``

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            string heightError = ("Height cannot be equal or less than zero, but was " + height);
+            string weightError = ("Weight cannot be equal or less than zero, but was " + weight);
+            string failedCalc = ("Failed calculating BMI. Reason:");
+
+            bool isW0OrLess = ((weight) <= 0);
```
Almantask, line 77: Avoid naming variables with just 1 letter, unless it's a well known abbreviation. instead of 'w', please write "Weight".

Almantask: Makes code less readable otherwise and I was confused for a split second what it was for.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            float bmi = (weight / height / height);
+            
+            return bmi;
+        }
+       
+        public static void Demo()
+        {
+            static void GatherAndPrint(string userNo)
```
Almantask, line 109: This should be moved outside of function to class scope. Nested function are harder to read, because they polute the function they are at by rubbing all the implementation details we don't need to care about.

Src/BootCamp.Chapter/Checks.cs
Shareni: I named it string1 because string string = threw a compiler error. But yeah, that makes a lot more sense

Src/BootCamp.Chapter/Lesson3.cs
Shareni: aaah i get it

Src/BootCamp.Chapter/Lesson3.cs
Shareni: I'm not sure if I've done what you've meant

Almantask: It's still there. ``static void GatherAndPrint(string userNo)`` should be outside of Demo method, inside Lesson3 class directly.

Shareni: I'm confused. The demo function is just this:
public static void Demo()
        {
            Process2PersonInfo();
        }

Should the Process2PersonInfo() be called by some other function, and demo left empty? But if it's empty how is the main function going to activate the program?

PR #290, opened 2020-04-13
All Test Passed

kclaka:
All test passed

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -4,6 +4,7 @@ using System.Text;
 
 namespace BootCamp.Chapter
 {
+    
     /// <summary>
     /// Test class is used to test your implementation.
     /// Each homework will have a set of steps that you will have to do.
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +15,67 @@ namespace BootCamp.Chapter
     /// </summary>
     public static class Checks
     {
+        public static string Name { get; set; }
```
Almantask, line 18: Since we don't delve into OOP now, everything is static in your eyes, which is fine. But here is a little background: static means global- can be accessed from anywhere, is always in memory and there is ever going to be a single value.

In the example here, name, surname and other fields will be assigned for one person at a time only. When they get assigned for a second person, the first person will loose the context.

Therefore they should be kept in local function scope, rather static class scope.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +15,67 @@ namespace BootCamp.Chapter
...
+        public static float Height { get; set; }
+
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            if (!int.TryParse(Console.ReadLine(), out int age))
+            {
+                //Console.WriteLine("Age Should be a Number");
```
Almantask, line 29: This is correct, not sure why you commented this out. Because of tests I reckon- which is fine. You will need this in hw4. So all good.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +15,67 @@ namespace BootCamp.Chapter
...
+
+            return Age;
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
+            Console.WriteLine(message);
+
```
Almantask, line 52: Please refactor the function not to mess with the program state- not a global state as well (static!). This means, that a function takes input and returns output. It should rarely change state of something static. This is justified for object methods (which we will learn in chapter 2), so for now avoid changing global values in functions.

```cs
{
  // ask for input
  Console.WriteLine(message);
  // return input (+ convert it if needed)
  return Console.ReadLine();
}
```

Also, due to the scope of a function being small, adding a blank line between every line of code is not so good. It's very related.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +15,67 @@ namespace BootCamp.Chapter
...
+            else
+            {
+                Height = height;
+            }
+            return Height;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
             // To do: call your implementation.
```
Almantask, line 77: Comment no longer needed.

Src/BootCamp.Chapter/Checks.cs
Almantask: Therefore, it's safe to remove all of those fields and move each of them close to where you use them- local function scope.

Minimizing scope is one of the things that a programmer tries to seek.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+            Checks.PromptString("Enter String Data");
+            Checks.PromptInt("Enter Int Data");
+            Checks.PromptFloat("Enter Float Data");
+            Checks.CalculateBmi(Checks.Height, Checks.Weight);
+
+            Console.WriteLine($"{Checks.Name} {Checks.Surname} is {Checks.Age} old, his weight is " +
+                $"{Checks.Weight}KG and his height is {Checks.Height}cm");
+        }
+
+        //    public string Name { get; set; }
```
Almantask, line 18: Please delete all the commented out code- adds no value.

PR #291, opened 2020-04-13
Chapter1/homework/5

jeanroyce:
This one was hard to get pass in all 42 test, but I did it.

One question, why this code wont work properly?

```C#
private static void Revert(int[] array)
{
	var tempArray = array;

	for (int i = 0; i < array.Length; i++)
	{
		array[i] = tempArray[^(i + 1)];
	}
}
```
Anotação 2020-04-12 233128

### DIFF
--- a/Src/BootCamp.Chapter1/ArrayOperations.cs
+++ b/Src/BootCamp.Chapter1/ArrayOperations.cs
@@ -99,6 +99,26 @@ namespace BootCamp.Chapter1
             return RemoveElement(array, index);
         }
 
+        // Return an array without the value in a specific index.
+        private static int[] RemoveElement(int[] array, int index)
+        {
+            int[] newArray = new int[array.Length - 1];
+
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                if (i >= index)
+                {
+                    newArray[i] = array[i + 1];
+                }
+                else
+                {
+                    newArray[i] = array[i];
+                }
+            }
+
+            return newArray;
+        }
+
         /// <summary>
         /// Inserts a new array element at the start.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Almantask:
> This one was hard to get pass in all 42 test, but I did it.
> 
> One question, why this code wont work properly?
> 
> ```cs
> private static void Revert(int[] array)
> {
> 	var tempArray = array;
> 
> 	for (int i = 0; i < array.Length; i++)
> 	{
> 		array[i] = tempArray[^(i + 1)];
> 	}
> }
> ```
> 
> Anotação 2020-04-12 233128

It's not working because arrays **are passed by reference**. tempArray and array are still pointing to the same reference (the same array). In order to create a new reference, you need to clone the array or create a new array and copy all the values.

What actually happened is:
iteration 1:
``{2 0 1}`` - places 2 at the start of array because it's the last element.
iteration 2:
``{ 2 0 1}`` - places 0 at the 2nd position of array (so nothing happens) because it's second last element
iteration 3:
``{2 0 2}`` - places 2 at the end of array because it's the first element (based on the already modified array)

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -9,7 +9,25 @@ public static class ArrayOperations
         /// <param name="array">Input array in a random order.</param>
         public static void Sort(int[] array)
         {
-            // ToDo: implement.
+            if (IsNullOrEmpty(array)) return;
+
+            bool isSorted;
+            do
+            {
+                isSorted = true;
```
Almantask, line 17: It's the opposite I think. Is it sorted at the start? No.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -9,7 +9,25 @@ public static class ArrayOperations
         /// <param name="array">Input array in a random order.</param>
         public static void Sort(int[] array)
         {
-            // ToDo: implement.
+            if (IsNullOrEmpty(array)) return;
+
+            bool isSorted;
+            do
```
Almantask, line 15: Use 2 for loops instead.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -9,7 +9,25 @@ public static class ArrayOperations
...
+            {
+                isSorted = true;
+                for (int i = 0; i < array.Length - 1; i++)
+                {
+                    if (array[i] > array[i + 1])
+                    {
+                        int oldIndexValue = array[i + 1];
```
Almantask, line 22: How many cups do you need to move wine from cup 1 to cup 2?

jeanroyce: 3 is enough

Src/BootCamp.Chapter1/ArrayOperations.cs
jeanroyce: I should have done the opposite.
I was about to lost my mind in this homework and I didn't pay attention, but now I refactored this part of code.

Src/BootCamp.Chapter1/ArrayOperations.cs
jeanroyce: Done now with 2 for loops.

jeanroyce:
> It's not working because arrays **are passed by reference**. tempArray and array are still pointing to the same reference (the same array). In order to create a new reference, you need to clone the array or create a new array and copy all the values.

I get it, thanks!

Almantask:
Looks good

PR #292, opened 2020-04-13
Chapter1/homework/1 and2

Trollermaner:
Finished the first homework

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,10 +6,56 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
+            //Person1
+            Console.WriteLine("Person #1:");
+            
             //Name
             Console.Write("Name: ");
             string name1 = Console.ReadLine();
 
+            //Age
+            Console.Write("Age: ");
+            int age1 =Convert.ToInt32( Console.ReadLine());
+
+            //Weight
+            Console.Write("Weight (kg): ");
+            float weight1 =Convert.ToInt32( Console.ReadLine());
+
+            //Height
+            Console.Write("Height (cm): ");
+            float height1 = Convert.ToInt32(Console.ReadLine());
+
+            //Bmi
+            float bmi1 = weight1 / (height1 * height1) * 10000;
+
+            Console.WriteLine(name1 +" is " + age1+" years old, his weight is " + weight1 + " kg, his height is "+height1+ " and his bmi is " +bmi1+".");
+
+            //------------------------------------------------------------------------------------------------------------------------------------------------------
+            Console.WriteLine(" ");
+
+            //Person2
+            Console.WriteLine("Person #2:");
+
+            //Name
+            Console.Write("Name: ");
+            string name2 = Console.ReadLine();
+
+            //Age
+            Console.Write("Age: ");
+            int age2 = Convert.ToInt32(Console.ReadLine());
+
+            //Weight
+            Console.Write("Weight (kg): ");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,29 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            //Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm. 
+            string Name1 = "Tom Jefferson";
+            int Age1 = 19;
```
Almantask, line 11: local variables should be named camelCase, not PascalCase.

Almantask, line 11: You should read input from the console. Use ``Console.ReadLine()`` and ``int/float.Parse`` to convert console input to values that you can work with.

Trollermaner: I'm having a problem where visual studio says : The application is in break mode
I've tried to fix it, but nothing worked

Almantask: Are you writing code while debugging? Maybe that's why?

Trollermaner: I don't know how but I fixed it somehow

Almantask: Well restart the VS or a PC next time if you have issues like that. Happens.

Trollermaner: I redid the code for everything

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,28 +6,29 @@ class Program
...
         {
-            //Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm. 
-            string Name1 = "Tom Jefferson";
-            int Age1 = 19;
-            int Weight1 = 50;
-            float Height1 = 156.5f;
-            float Height1v2 = Height1 / 100;
-            float Height1v3 = Height1v2 * Height1v2;
-            float Bmi1 = Weight1 / Height1v3;
+            //Bmi Calculator
```
Almantask, line 9: Comment isn't needed- code speaks for itself.

Almantask: All comments in this case can be removed as they don't give any extra info.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,28 +6,29 @@ class Program
     {
         static void Main(string[] args)
         {
```
Almantask, line 8: Second person is missing

Src/BootCamp.Chapter/Program.cs
Almantask, line 9: Main should call high level functions. Calcluation of Bmi (full contents of main) should be moved into a function of its' own as well.

Trollermaner: fixed?

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,30 +6,33 @@ class Program
...
-            //Bmi Calculator
-            Console.WriteLine("BMI Calculator:");
-            
-            //Full Name
+            Bmi1(number:1);
```
Almantask, line 9: Bmi1 is a noun. Also 1 next to method doesn't mean anything. Lastly, ``number`` next to :1 is not needed, because it is called a positional stuff and is used when you specify argument out of order, especially when it's optional.

Should be: ``CalculateBmi(1);``

PR #295, opened 2020-04-15
Homework 5! This took way longer than I had hoped it would.

MyAxes:


Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,171 @@
...
-        /// <returns>A new array with element inserted at a given index. If an array is empty or null, returns new array with number in it.</returns>
         public static int[] InsertAt(int[] array, int number, int index)
         {
-            // ToDo: implement.
-            return array;
+            if (CheckForInsert(array, index, false))
+            {
+                return new int[] { number };
+            }
+            else if (!CheckForInsert(array, index, true))
```
Almantask, line 115: else is not needed, because the first return will terminate the function.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,171 @@
...
+            {
+                return array;
+            }
+            else
```
Almantask, line 119: else is not needed, because the first return will terminate the function.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,171 @@
...
+            {
+                int[] addElement = new int[array.Length + 1];
```
Almantask, line 121: ``elementsWithInsertion`` or similar should be the name. ``AddElement`` is a verb and a variable name should be a noun.

PR #296, opened 2020-04-15
Chapter1/homework/3

Flebberflep:
Code refactored, tests passed.

I'm sure there must be a better way to get user input than having one method for each variable type, but would that be overcomplicating things at this stage?

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            //Get input from the user.
+            string fName = GetStringInput("Enter your first name: ");
+            string lName = GetStringInput("Enter your last name: ");
+            int age = GetIntInput("Enter your age: ");
+            float height = GetFloatInput("Enter your height (in M): ");
+            float weight = GetFloatInput("Enter your weight (in kg): ");
+
+            //Print back information and follow up with BMI calculation.
+            Console.WriteLine("{0} {1} is {2} years old, weighs {3} kg and is {4} cm tall.", fName, lName, age, weight, height);
+            Console.WriteLine("{0}'s BMI is {1}.", fName, CalcBMI(height, weight));
+
+            //Halt program so output can be read.
+            Console.ReadKey();
+        }
+
+        public static float CalcBMI(float weight, float height)
+        {
+            var bmi = weight / (height * height);
+            return bmi;
+        }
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,50 @@
...
+            int age = GetIntInput("Enter your age: ");
+            float height = GetFloatInput("Enter your height (in M): ");
+            float weight = GetFloatInput("Enter your weight (in kg): ");
+
+            //Print back information and follow up with BMI calculation.
+            Console.WriteLine("{0} {1} is {2} years old, weighs {3} kg and is {4} cm tall.", fName, lName, age, weight, height);
+            Console.WriteLine("{0}'s BMI is {1}.", fName, CalcBMI(height, weight));
+
+            //Halt program so output can be read.
+            Console.ReadKey();
```
Almantask, line 23: Second person is missing.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            //Get input from the user.
```
Almantask, line 11: Contents of demo should call another function- ProcessPersonBmi, PromptBmi or similar.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,50 @@
...
+            string fName = GetStringInput("Enter your first name: ");
```
Almantask, line 12: No need to use abriaviation in naming variables, unless they are common knowledege

Flebberflep:
Made requested changed to the code, have also added comments to clarify some things.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -7,22 +7,27 @@ namespace BootCamp.Chapter
...
-
-            //Halt program so output can be read.
-            Console.ReadKey();
+            Console.WriteLine($"{firstName} {lastName} is {age} years old, weighs {weight} kg and is {height} M tall.");
+            Console.WriteLine($"{firstName}'s BMI is " + CalcBMI(weight, height));
+            //Tried breaking these print lines into their own method block, however it would need 5 input parameters and would need to be
+            //called in this block to keep the variables in scope, so seemd to be a bit pointless. I imagine there's a better way to do it.
         }
-
         public static float CalcBMI(float weight, float height)
```
Almantask, line 31: Don't try to make it shortened, use full words unless it's a common knowledge. Also use PascalCase.
``CalculateBmi()`` is a better name.

PR #300, opened 2020-04-16
Chapter1/homework/7

jeanroyce:
Pass in all 32 tests.

I think my code is a real mess, but I think I will wait for feedback.

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -7,21 +7,15 @@ namespace BootCamp.Chapter
 {
     public static class BalanceStats
     {
-
-
         /// <summary>
         /// Return name and balance(current) of person who had the biggest historic balance.
         /// </summary>       
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
             if (IsNullOrEmpty(peopleAndBalances)) return "N/A.";
-
             var (peopleNames, money) = ProcessPeopleAndAbalances(peopleAndBalances, "max");
 
-            var listToBuid = AnalisePeopleAndMoney(peopleNames, money, money.Max());
-            var listOfNames = BuildListOfNames(listToBuid.ToString());
-
-            return $"{listOfNames} had the most money ever. {FixMoneyFormat(money.Max())}.";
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,199 @@
...
     public static class BalanceStats
     {
         /// <summary>
         /// Return name and balance(current) of person who had the biggest historic balance.
-        /// </summary>
+        /// </summary>       
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            if (IsNullOrEmpty(peopleAndBalances)) return "N/A.";
```
Almantask, line 15: "N/A" should be moved to a constant, because now it needs to change what is returned for invalid value- we will have to change multiple places.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,199 @@
...
+
+            var isMaxValue = (option == "max" || option == "rich");
+            float moneyToAnalyse = (isMaxValue) ? money.Max() : money.Min();
+            var listToBuid = AnalisePeopleAndMoney(peopleNames, money, moneyToAnalyse);
+            var listOfNames = BuildListOfNames(listToBuid.ToString());
+
+            return (listOfNames, moneyToAnalyse);
+        }
+
+        // Convert an array of string to array of floats
```
Almantask, line 114: Comment is not needed

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +37,73 @@ public static class TextTable
...
+
+            return text;
+        }
+
+        private static StringBuilder BuildMessage(string[] array, string emptySpace)
+        {
+            var temp = new StringBuilder();
+            foreach (var words in array)
+            {
+                temp.Append($"|{emptySpace}{words}{emptySpace}|{Environment.NewLine}");
```
Almantask, line 96: instead of providing Environment.NewLine, you can use AppendLine

jeanroyce:
Thanks for the feedback.

PR #305, opened 2020-04-17
Completed homework for Chapter 1 Lesson 1

Cl0utEastwood:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,41 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string Name1 = "todd";
```
Almantask, line 9: Variables should be camelCased named (first letter non capital followed by capitalized letter of every next word).

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,41 @@ class Program
     {
         static void Main(string[] args)
```
Almantask, line 7: you need to read everything from the console

PR #307, opened 2020-04-17
My submission for Chapter 1 homework 1and2

Bosco-2019:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,28 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            //Can this method work for both Tom Jefferson and the second person?
```
Almantask, line 9: Looks good, but second person is missing.

Bosco-2019: Have made those changes now, let me know if this approach is ok. Thanks

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,38 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name1 = ("Tom Jefferson");
+            int age1 = 19;
```
Almantask, line 10: One more change is needed- pleas read from console. Refer to ``Console.ReadLine`` ``inter.Parse`` double.Parse``

Src/BootCamp.Chapter/Program.cs
Almantask: Almost, but still need to read input from console.

Src/BootCamp.Chapter/Program.cs
Bosco-2019: not entirely sure what that means?

Almantask: instead of 
```
            double weight1 = 50;
            double height1 = 156.5;
```
You need to input those numbers using console. because user won't be changing the code.
So it should be:
```
            double weight1 = double.Parse(Console.ReadLine());
            double height1 = double.Parse(Console.ReadLine());
```

PR #309, opened 2020-04-18
Chapter1/homework/8

jeanroyce:
Throw everywhere lol.
I virtually passed in all tests, only need to change one line (commented) to throw errors to pass in almost all. I build my code to ignore all non common letters(and some symbols) in name, and .(dot) is not in whitelist. 2 balances are not in final uncorrupted balances.txt, but I log to txt as exception.
Also I refactored TextTable and BalanceStats with a different logic as an exercise.


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
         public static string FindRichestPerson(string[] peopleAndBalances)
         {
-            return "";
+            if (IsNullOrEmpty(peopleAndBalances)) return Constants.Nothing;
+            Thread.CurrentThread.CurrentCulture = Config.Culture();
+
+            var (names, money, isPlural) = ReturnNamesMoneyAndIsPlural(peopleAndBalances, "rich");
+            var dictionary = new string[]
+            {
+                (isPlural) ? "are" : "is",          // [0]
```
Almantask, line 51: Weird comments, can be removed.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
         }
 
         /// <summary>
         /// Return name and current money of the richest person.
         /// </summary>
         public static string FindRichestPerson(string[] peopleAndBalances)
         {
-            return "";
+            if (IsNullOrEmpty(peopleAndBalances)) return Constants.Nothing;
+            Thread.CurrentThread.CurrentCulture = Config.Culture();
```
Almantask, line 46: I was expecting just Console.OutputEncoding to be changed. I think changing more stuff might have unexpected consequences (for example this now changes DateTime formatting as well). You just want to change text currency formatting and nothing else.

A+ for effort and originality though!

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
         /// <summary>
         /// Return name and current money of the most poor person.
         /// </summary>
         public static string FindMostPoorPerson(string[] peopleAndBalances)
         {
-            return "";
+            if (IsNullOrEmpty(peopleAndBalances)) return Constants.Nothing;
+            Thread.CurrentThread.CurrentCulture = Config.Culture();
+
+            var (names, money, isPlural) = ReturnNamesMoneyAndIsPlural(peopleAndBalances, "poor");
```
Almantask, line 66: It's interesting to see object deconstruction. I don't think it's that great, to me this looks like flexing 🍡

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+            var dictionary = new string[]
+            {
+                (isPlural) ? "have" : "has",        // [0]
+            };
+
+            return $"{names} {dictionary[0]} the least money. {FixMoneyFormat(money)}.";
+        }
+
+        private static (string names, decimal money, bool isPlural) ReturnNamesMoneyAndIsPlural(string[] peopleAndBalances, string option)
```
Almantask, line 75: The function name is suboptimal. Every function does as return, so that part is redundant.
A better name would be ``FindPeopleBalancesByOption``

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+            {
+                if (money[i] == moneyToCompare) people.Add(name[i]);
+            }
+
+            var (processedNames, isPlural) = ProcessName(people);
+
+            return (processedNames, moneyToCompare, isPlural);
+        }
+
+        private static (List<string> name, List<decimal>) ProcessPeopleBalance(string[] peopleAndBalances, string option)
```
Almantask, line 93: Avoid using list, avoid using class for the first chapter. Pretend you don't know about the list yet. Also, don't return tuples or deconstruction. Return a single value and call different functions multiple times. For chapter 2 you will learn objects, then you can prefer that. This is not very readable.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+        {
+            var nameList = new List<string>();
+            var moneyProcessed = new List<decimal>();
+            for (var i = 0; i < peopleAndBalances.Length; i++)
+            {
+                try
+                {
+                    var moneyList = CheckData(peopleAndBalances[i], out var name);
```
Almantask, line 101: What does check data mean? For something so vague, please add a comment.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+                    Console.WriteLine($"The following data from index: {i} is not valid, skipping.");
+                    Console.WriteLine($"\"{peopleAndBalances[i]}\"");
+                    Console.ResetColor();
+                    continue;
+                }
+            }
+
+            if (nameList.Count != moneyProcessed.Count || nameList == null || moneyProcessed == null || nameList.Count == 0 || moneyProcessed.Count == 0)
+            {
+                throw new Exception($"Something went very wrong in function ProcessPeopleBalance().");
```
Almantask, line 140: If you do throw an exception, don't throw default exception. Make your own, custom exception. Also, something went wrong is not very helpful, in fact it's worse that not handling the exception (because you don't leave any traces for the error that happened. Normally the error should be logged and rethrown or extra details provided, but never less details provided. This might have been an imitation for giving a user friendly error, but you still throw an exception, so that is not the case.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+            }
+            return (nameList, moneyProcessed);
+        }
+
+        private static decimal CheckBalance(List<decimal> moneyList, string option)
```
Almantask, line 145: Ambigious name, have no idea what it does.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +15,214 @@ public static class BalanceStats
...
+
+            return option switch
+            {
+                "max" => balance.Max(),
+                "min" => balance.Min(),
+                _ => 0
+            };
+        }
+
+        private static List<decimal> CheckData(string line, out string name)
```
Almantask, line 162: ambigious name. Better name would be ValidateBalancesFile.

Src/BootCamp.Chapter/FileCleaner.cs
```diff
@@ -12,9 +14,136 @@ public static class FileCleaner
...
+        {
+            const string emptyLine = "Line is null or empty.";
+            const string emptyName = "Name of account not found.";
+            const string nonLatin = "[Name] has non-latin character valid for name.";
+            const string moneyError = "Error in money registry.";
+
+            var data = line.Split(",");
+            if (data.Length == 0) throw new InvalidBalancesException(emptyLine);
+            if (data[0] == "" || string.IsNullOrWhiteSpace(data[0])) throw new InvalidBalancesException(emptyName);
+            if (!Regex.IsMatch(data[0], Constants.AllCommonLatinLetters)) throw new InvalidBalancesException(nonLatin);
```
Almantask, line 116: Overkill for this homework 🗡️

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +38,67 @@ public static class TextTable
...
+            BuildMainLines(padding, main, maxLength);
+            var paddingLines = BuildPaddingLines(padding, maxLength);
+            var border = $"+{GenSpaces((padding * 2 + maxLength), '-')}+";
+
+            return $"{BuildFinalMessage(border, paddingLines, main)}{Environment.NewLine}";
+        }
+
+        private static List<string> InitializeMain(string message, out int maxLength)
+        {
+            var main = new List<string>();
```
Almantask, line 54: Don't use arrays or lists, etc for string building operations. Use ``StringBuilder``, due to optimization reasons and utility method.

jeanroyce:
> Try to keep it simple. I think you overdid it with multiple returns, regex matching.
> Also, be mindful how you name things. Most of the function names are ambiguous and up to interpretation for what they do.

You are right, my code wasn't good, even I couldn't understand this code today, KISS code 0.
I remade most of parts. I dont think is good now, but I think is more simple now. Removed all trace of List and Tuple.

One more thing, I think there is something wrong with Test in Visual studio, `BalancesNotAName2y.Invalid` is suppose to fail because of the name `To.m`, but `Sydney Godehard.sf` and `Barbie Pankhurst.` are in expected files, so I changed those files myself.

Src/BootCamp.Chapter/TextTable.cs
jeanroyce: Ok

Src/BootCamp.Chapter/BalanceStats.cs
jeanroyce: Im not a creative person, variables and functions names is something painful for me kkkk.

Src/BootCamp.Chapter/BalanceStats.cs
jeanroyce: Changed to CalculateNetGain.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -61,153 +51,108 @@ public static string FindRichestPerson(string[] peopleAndBalances)
...
-            var (names, money, isPlural) = ReturnNamesMoneyAndIsPlural(peopleAndBalances, "poor");
+            var balances = FindPeopleBalancesByOption(peopleAndBalances, "poor");
             var dictionary = new string[]
             {
-                (isPlural) ? "have" : "has",        // [0]
+                (balances[1] == "1") ? "have" : "has",
             };
 
-            return $"{names} {dictionary[0]} the least money. {FixMoneyFormat(money)}.";
+            return $"{balances[0]} {dictionary[0]} the least money. {balances[2]}.";
```
Almantask, line 60: Index of an array is a magic number. Use constants to name the index based on representing column.

Almantask:
This test is failing 🧚  Please fix.
``FileCleaner_Clean_Given_Invalid_Parameters_Throws_ArgumentException``

Src/BootCamp.Chapter/BalanceStats.cs
jeanroyce: You say something like this?
```c#
var dictionary = new string[2];
dictionary[0] = (balances[1] == "1") ? "are" : "is";
dictionary[1] = (balances[1] == "1") ? "people" : "person";
```

PR #310, opened 2020-04-18
Chapter1/homework/7

DavidKatona:
I've spent way much more time on this than I originally wanted but here it goes. I've changed my workflow a bit based on the experience I've got from previous HWs and from the live coding interview we had.

I've decided to go simple first and ignore the fact that I have a lot of repetition in the code. Then I gradually noted down the similarities and differences between functions and moved all that was similar to utility functions. These are found under the static class ArrayProcessor and ArrayOperations.

In the end, I think I still have some repetition in the ArrayProcessor itself, but I had a hard time figuring it out how to change that and I'd like to continue with the lessons.

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -16,6 +16,7 @@ namespace BootCamp.Chapter
 
             var splitArray = peopleAndBalances[0].Split(',');
             var decimalArray = ConvertStringArrayToDecimalArray(splitArray[1..]);
+            var highestValueInArray = FindHighestDecimalInArray(decimalArray);
 
             return "";
         }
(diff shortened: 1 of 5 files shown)


Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -1,40 +1,87 @@
...
+            AppendHorizontalBorder(sbTable, messageWidth);
+            AppendEmptyLines(sbTable, maxLength, padding);
+            AppendLinesWithContent(words, sbTable, maxLength, padding);
+            AppendEmptyLines(sbTable, maxLength, padding);
+            AppendHorizontalBorder(sbTable, messageWidth);
+
+            return sbTable.ToString();
+        }
+
+        private static void AppendLinesWithContent(string[] inputArray, StringBuilder sbTable, int maxLength, int padding)
```
Almantask, line 40: Can be just ``AppendLine``

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,88 @@ public static class BalanceStats
...
+            if (ArrayOperations.IsArrayNullOrEmpty(peopleAndBalances))
+                return "N/A.";
+
+            var lowestBalanceEver = AccountProcessor.ProcessFirstAccount(peopleAndBalances, OperationType.FindSmallestValue, out string[] name);
+            var individualsWithLowestBalance = name;
+
+            lowestBalanceEver = AccountProcessor.ProcessRemainingAccounts(peopleAndBalances, lowestBalanceEver, individualsWithLowestBalance, OperationType.FindSmallestValue, out string[] names);
+            individualsWithLowestBalance = names;
+
+            var formattedLowestBalance = $"¤{lowestBalanceEver}";
```
Almantask, line 82: Ideally, try using formatting with currency. ``lowestBalance.ToString(new Culturue.Invariant())`` or similar.

PR #312, opened 2020-04-19
Chapter1/Homework/9

jeanroyce:


Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -1,19 +1,42 @@
...
     public static class BinaryConverter
     {
         public static long ToInteger(string binary)
         {
-            return 0;
+            if (string.IsNullOrEmpty(binary)) return 0;
+            if (!Regex.IsMatch(binary, @"^[01]+$")) throw new InvalidBinaryNumberException(binary);
+
+            var num = 0;
+            foreach (var bin in binary)
```
Almantask, line 15: should be ``foreach(digit in binary)``

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -1,19 +1,42 @@
...
+            {
+                if (!int.TryParse(bin.ToString(), out var value)) throw new ArgumentOutOfRangeException("An unexpected error occurred");
```
Almantask, line 17: If exception message is ever going to be just that, please don't give any message. It's noise. And besides, in this case don't use int.TryParse, just use int.Parse and it will fail giving more info than you give right now.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,120 @@
...
 {
     class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            Console.OutputEncoding = System.Text.Encoding.Unicode;
+
+            MainMenu();
+
```
Almantask, line 14: Random blank line, can be removed.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,120 @@
...
+        }
+
+        private static void MainMenu()
+        {
+            Console.Clear();
+            Console.WriteLine("Welcome! Choose one option");
+
+            var options = new List<string>
```
Almantask, line 22: Forget about using list in this chapter! 😄 For this specific case, array, in fact is better, because you have a predefined list of commands that you know. Since size won't change and all the commands are known- array is better.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,120 @@
...
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            Console.OutputEncoding = System.Text.Encoding.Unicode;
+
+            MainMenu();
+
+        }
+
+        private static void MainMenu()
```
Almantask, line 17: It shouldn't be a noun, it should be a verb. DisplayMenu is a better name.

PR #313, opened 2020-04-19
Max Chapter3/homework/6

HellHunterMax:
pretty happy about this. i think it could have some improvements i missed on.
and i'm not quite sure i did the events part as it should be.
happy it works though.

### DIFF
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -23,7 +23,6 @@ namespace BootCamp.Chapter
 
         private void LoadPeople(string peopleFile)
         {
-            //TODO LoadPoeple
             string[] peoples = File.ReadAllText(peopleFile).Split(Environment.NewLine);
 
             if (peoples.Length == 1)
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -23,7 +23,6 @@ namespace BootCamp.Chapter
             // Assing what it must do first, for example with a lambda.
             //predicate = p => p.Age > 18;
             var people = new List<Person>();
-            // ToDo: implement applying filter.
             foreach (Person person in _people)
             {
                 if (predicate(person))
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -19,8 +19,18 @@ namespace BootCamp.Chapter
         /// <returns></returns>
         public List<Person> Filter(Predicate<Person> predicate)
         {
+            // MY COMMENT  predicate doesnt do anything untill called, just like delegate.
+            // Assing what it must do first, for example with a lambda.
+            //predicate = p => p.Age > 18;
             var people = new List<Person>();
             // ToDo: implement applying filter.
+            foreach (Person person in _people)
+            {
+                if (predicate(person))
+                {
+                    people.Add(person);
+                }
+            }
             return people;
         }
     }
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
+using System.IO;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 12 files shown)


Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -10,7 +10,34 @@ public class ContactsCenter
...
+            }
+        }
+
+        private void LoadPeople(string peopleFile)
+        {
+            string[] peoples = File.ReadAllText(peopleFile).Split(Environment.NewLine);
+
+            if (peoples.Length == 1)
+            {
+                throw new Exception($"{peopleFile} was Empty");
```
Almantask, line 30: Don't throw default exceptions. Either throw specialized exception or make one yourself. In this case- 
```InvalidPeoplesFileException(file)```

File: Src/BootCamp.Chapter/ContactsCenter.cs [hunk 2]
```diff
@@ -10,7 +10,34 @@ public class ContactsCenter
...
+            }
+
+            foreach (string personLine in peoples)
```
Almantask, line 33: LIne is redundant, can simply be person in people

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -19,8 +46,17 @@ public ContactsCenter(string peopleFile)
         /// <returns></returns>
         public List<Person> Filter(Predicate<Person> predicate)
         {
+            // MY COMMENT  predicate doesnt do anything untill called, just like delegate.
```
Almantask, line 49: "MY COMMENT" can be removed.

Src/BootCamp.Chapter/Demo.cs
```diff
@@ -0,0 +1,133 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace BootCamp.Chapter
+{
+
+    public class Demo
+    {
+        Dictionary<string, ConsoleKey> consoleOptions = new Dictionary<string, ConsoleKey>();
+        public event LogEventHandler logEventHandler;
```
Almantask, line 10: Should be renamed to ``OnMenuKeyPressed``

Almantask: And in general should belong to the menu class. inside demo, you can demo how you handle that even. But it's good enough for this homework.

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,105 @@
...
+            FirstName = name;
+            SurName = surName;
+            Birthday = birthday;
+            Gender = gender;
+            Country = country;
+            Email = email;
+            StreetAddress = streetAddress;
+        }
+
+        public int GetAge()
```
Almantask, line 31: Should be a property.

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,105 @@
...
+        {
+            string now = DateTime.Now.ToString().Split(' ')[0];
+            int day = int.Parse(now.Split('/')[0]);
```
Almantask, line 34: All this is redundant. 
DateTime has properties for day, hour, month, etc...

Src/BootCamp.Chapter/Tests.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    public static class Tests
```
Almantask, line 5: Rename from ``Tests`` to ``Validator``.

Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,105 @@
...
+        {
+            foreach (string line in split)
+            {
+                if(String.IsNullOrWhiteSpace(line))
+                {
+                    return false;
+                }
+            }
+
+            if (!Tests.IsValidBirthday(split[2]))
```
Almantask, line 84: Magic numbers for array indices. Should use constants or enum and index like that. It's hard to keep track otherwise.

HellHunterMax:
Thank you for the review! All valid and understandable remarks, I'll work on them tonight.

Src/BootCamp.Chapter/Demo.cs
HellHunterMax: Do I understand correctly that I should make a new class for the menu and put the login in there? Sounds better. I'll see if I can get that fixed.

Almantask: Yes

Src/BootCamp.Chapter/ContactsCenter.cs
HellHunterMax: But its not a person yet and I have an out called person, that is an actual person?

Src/BootCamp.Chapter/Demo.cs
HellHunterMax: much cleaner. thank you.

HellHunterMax:
should the LogEventHandler class be in its own class or above the Menu class ?

Src/BootCamp.Chapter/Logger.cs
```diff
@@ -6,12 +6,12 @@ namespace BootCamp.Chapter
 {
     public class Logger
```
Almantask, line 7: I guess that could work, but then this would be a specialized logger, like ``MainMenuLogger``.

PR #318, opened 2020-04-19
Homework 1 & 2 submission

Abubakar113:
This is my first attempt at writing C# code. Thank you C# Inn for providing this type of  environment

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -5,8 +5,41 @@ namespace BootCamp.Chapter
     class Program
     {
         static void Main(string[] args)
-        {
-            Console.WriteLine("Hello World!");
+        {   /*
+                This is the code for Homework 1 and 2 
+                This section I have written the variables for Tom Jefferson
+
+            */
+            Console.WriteLine("This is for Homework 1");
+            int tomAge = 19;
+            float tomWeight = 50.0f;
+            string tomName = "Tom Jefferson";
+            float tomHeight = 156.5f;
+            const int cmToM = 10000;
+            float tomBMI = tomWeight / ((tomHeight * tomHeight) / cmToM);
+
+            // This is where Tom's Info is displayed
+            Console.Write(tomName + " is " + tomAge + " years old, ");
+            Console.Write("his weight is " + (int)tomWeight);
+            Console.WriteLine(" kg and his height is " + tomHeight + " cm.");
+            Console.WriteLine(tomName + "'s BMI is: " + tomBMI);
+
+            //This is where a user's info is being entered
+            Console.Write("Please enter the name of the user: ");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -5,8 +5,41 @@ namespace BootCamp.Chapter
...
+                This is the code for Homework 1 and 2 
+                This section I have written the variables for Tom Jefferson
+
+            */
+            Console.WriteLine("This is for Homework 1");
+            int tomAge = 19;
+            float tomWeight = 50.0f;
+            string tomName = "Tom Jefferson";
+            float tomHeight = 156.5f;
+            const int cmToM = 10000;
```
Almantask, line 18: I think the purpose of this code section was to practice variables, const, etc. Not a bad thing, but expectation was to read all from console. You did that in second section, so all good :)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -5,8 +5,41 @@ namespace BootCamp.Chapter
...
+
+            //This is where a user's info is being entered
+            Console.Write("Please enter the name of the user: ");
+            string userName = Console.ReadLine();
+            Console.Write("Please enter the age of " + userName + ": ");
+            int userAge = Int32.Parse(Console.ReadLine());
+            Console.Write("Please enter the weight of " + userName + ": ");
+            float userWeight = Single.Parse(Console.ReadLine());
+            Console.Write("Please enter the height of " + userName + ": ");
+            float userHeight = Single.Parse(Console.ReadLine());
```
Almantask, line 35: Instead of single, you can use float.Parse. Slightly simplified.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -5,8 +5,41 @@ namespace BootCamp.Chapter
...
+            Console.Write(tomName + " is " + tomAge + " years old, ");
+            Console.Write("his weight is " + (int)tomWeight);
+            Console.WriteLine(" kg and his height is " + tomHeight + " cm.");
+            Console.WriteLine(tomName + "'s BMI is: " + tomBMI);
+
+            //This is where a user's info is being entered
+            Console.Write("Please enter the name of the user: ");
+            string userName = Console.ReadLine();
+            Console.Write("Please enter the age of " + userName + ": ");
+            int userAge = Int32.Parse(Console.ReadLine());
```
Almantask, line 31: instead of Int32, you can use int. Slightly simplified.

Src/BootCamp.Chapter/Program.cs
Abubakar113: Ohhh I see. I thought the the way how the homework was written was we need to enter the first one by ourselves and the second one through the console. I am sorry English isn't my first language.

PR #321, opened 2020-04-21
Chapter1/homework/5

JereTR:
I spent more time researching and learning than testing to see what works & doesn't work on this one.  I wanted to know how something worked, then looked into why code worked; whether online or through discord.  

I dunno if researching is better than testing & failing, but after bonking my head on the last homework for too long, I wanted to understand the subject.

I received plenty of help in discord, and appreciated all assistance.

### DIFF
--- a/Src/BootCamp.Chapter1/ArrayOperations.cs
+++ b/Src/BootCamp.Chapter1/ArrayOperations.cs
@@ -106,9 +106,14 @@
         /// <returns>A new array with element added at a given index. If an array is empty or null, returns new array with number in it.</returns>
         public static int[] InsertFirst(int[] array, int number)
         {
+            if (array == null)
+            {
+                int[] First = new[] { number };
+                return First;
+            }
 
-            // ToDo: implement.
-            return array;
+            int[] First2 = InsertAt(array, number, 0);
+            return First2;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Almantask:
> I spent more time researching and learning than testing to see what works & doesn't work on this one. I wanted to know how something worked, then looked into why code worked; whether online or through discord.
> 
> I dunno if researching is better than testing & failing, but after bonking my head on the last homework for too long, I wanted to understand the subject.
> 
> I received plenty of help in discord, and appreciated all assistance.

It really depends on your goals. For general completion of task, it is not needed to achieve expertise or have a good understanding on subject X. In programming sense, all you need is a good understanding of input and the expected output. The details start to matter when you are creating, inventing, upgrading or cleaning up things. But practising like this is definitely good because you will remember. It's also good because you pushed yourself for going deep into a given subject.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -2,94 +2,137 @@
...
         {
-            // ToDo: implement.
+            if (array == null || array.Length == 0)
+            {
+                return;
+            }
+            int revArray = array.Length;
+            for (int i = 0; i < revArray / 2; i++)
+            {
+                int temp = array[i];
```
Almantask, line 34: Reverse algorithm and BubbleSort actually share one similar part- element swap :)
You can create Swap method and reuse it in both functions.

PR #322, opened 2020-04-21
Chapter1/homework/6

Smephey:
Part 2 took me forever to realise that the password was actually being spit into the console 😂 I was digging through the debugger console trying to find the secret password. But I think I got it all now. 
Part 1 -> "Debugger!", "51", "79", "77"
Part 2 -> "A minore ad maius a solis ortu usque ad occasum ab uno disce omnes"
-> "smaller, sunrise, sunset"


Almantask:
Well done! :)

PR #324, opened 2020-04-21
Chapter1/homework/1 and2

Dreniak:
Did the deed.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,33 +6,25 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            string name;
-            string age;
-            string weight;
-            float Weight;
-            string height;
-            float Height;
-            float bmi;
-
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,53 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
```
Almantask, line 9: Don't declare all variables in the same place. Declare them right next to the place they are used to.
Age- next to getting and parsing age;
Weight- next to getting and parsing weight;
...

Dreniak:
I declared the variables in the same place so I could have the exact same code for any number of people, just repeating that section. But I guess this practice would hinder people from understanding my code later on.

Almantask:
> I declared the variables in the same place so I could have the exact same code for any number of people, just repeating that section. But I guess this practice could hinder people from understanding my code later on.

Looks good now :)

PR #325, opened 2020-04-21
Homework 3

joshuajz:
I've passed all of the required tests but I'm unsure as to what the implementation for the Demo() function is supposed to be (ie. what is it supposed to do?)

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -17,25 +17,25 @@ namespace BootCamp.Chapter
         public static int PromptInt(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson3.IntInput(message);
         }
 
         public static string PromptString(string message)
         {
             // To do: call your implementation. 
-            return "";
+            return Lesson3.StringInput(message);
         }
 
         public static float PromptFloat(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson3.FloatInput(message);
         }
 
         public static float CalculateBmi(float weight, float height)
         {
             // To do: call your implementation. 
-            return 0;
+            return Lesson3.BMICalculation(weight, height);
         }
     }
 }
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,22 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -17,25 +17,25 @@ public static class Checks
         public static int PromptInt(string message)
         {
             // To do: call your implementation.
```
Almantask, line 19: Comments can be removed as they are no longer needed.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            Console.WriteLine("Temp, not sure what to do here yet");
```
Almantask, line 11: Demo should call the functions that you made- specifically test the two people bmi calculation that you did in homework 1. and 2

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+        
+        // Takes the weight and height and returns BMI
+        public static float BMICalculation(float weight, float height)
```
Almantask, line 15: Method names should be nouns, not verbs. CalculateBmi sounds better.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            return weight / (height * height);
+        }
+
+        // Prompts the user for an input, converts it to an integer and returns it
+        public static int IntInput(string message)
```
Almantask, line 21: PromptInt is the name I would expect. Why? Because you ask for inputting an integer and such name tell that. IntInput is a bit technical, not obvious, up for interpretation and thus is not as good.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -8,33 +8,51 @@ class Lesson3
     {
         public static void Demo()
         {
-            Console.WriteLine("Temp, not sure what to do here yet");
+            // Person 1 Input (name, age, weight, height)
```
Almantask, line 11: Use function for bmi processing as well. Now you have 2 duplicate blocks of code which can be perfectly extracted to a function.

joshuajz: Wow that's very smart! I didn't even think of that!

PR #326, opened 2020-04-21
Chapter2/homework/1

jeanroyce:
I removed `Barbie Pankhurst.` from all test files.


jeanroyce:
> Instead of working with strings (the way you did in chapter1), try working with objects. Convert the parsed lines into people and then process people and not strings.

Done!

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -0,0 +1,155 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public static class BalanceStats
+    {
+        public static string FindHighestBalanceEver(string[] peopleAndBalances)
```
Almantask, line 10: You're still working with string[]. That needs to be changed to Person[].

jeanroyce:
BalanceStats now only process people (object), I think 😁

Src/BootCamp.Chapter/Person.cs
```diff
@@ -0,0 +1,23 @@
+namespace BootCamp.Chapter
+{
+    public class Person
+    {
+        private readonly string _name;
+        public string GetName()
```
Almantask, line 6: Use properties and not getter methods!

Almantask:
Looks good now. But if you can put logic into models in isolation, you should.

PR #329, opened 2020-04-22
Chapter1/homework/3

Trollermaner:
Completed homework 3 of chapter 1

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,27 +16,30 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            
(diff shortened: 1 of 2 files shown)


Almantask:
Bmi calculation is faulty. Try to run the tests and see for yourself. It is 1000x higher Bmi than expected.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -39,7 +39,7 @@ public static float PromptFloat(string message)
 
         public static float CalculateBmi(float weight, float height)
         {
-            return weight / height / height * 10000f;
+            return weight / height / height * 1f;
```
Almantask, line 42: No need to multiple from anything 😄

PR #330, opened 2020-04-23
Chapter1/homework/1 and2

lukas-ostrovskis:
Homework 1&2: Calculation of BMI

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,7 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,7 +7,6 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,38 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
```
Almantask, line 9: Don't declare all variables in one place- it disrupts readability, because you will have to scroll up and down all the time.

What you should be doing instead is declaring variables right next to where you use them.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,38 @@ class Program
...
+            string surname;
+            int age;
+            double weight;
+            double heightCM, heightM;
+
+            double BMI;
+
+            for(int i = 0; i < 2; i++)
```
Almantask, line 17: Well done for using a loop :)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,38 @@ class Program
...
+            {
+                Console.Write("Name: ");
+                name = Console.ReadLine();
+
+                Console.Write("Surname: ");
+                surname = Console.ReadLine();
+
+                Console.Write("Age: ");
```
Almantask, line 25: Well done for separating different parts with a blank line.

PR #331, opened 2020-04-23
Chapter2/Homework/3

jeanroyce:


Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -1,15 +1,71 @@
...
 
 namespace BootCamp.Chapter
 {
-    public class MsFactory
+    public class MsFactory : Factory
     {
         public DesktopComputer Assemble()
         {
-            return new DesktopComputer();
+            var pc = new DesktopComputer();
```
Almantask, line 12: If you compare computer assembling in Ms and Mac, what's the difference? The assemble is exact same, except that individual steps are different.

Almantask: This flow (assembling) can be implemented in abstract class.

Src/BootCamp.Chapter/Parts.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace BootCamp.Chapter
+{
+    public class Parts
```
Almantask, line 3: Please remove redundant class.

jeanroyce:
Don't know if is correct, but I think abstract class make more sense in my mind now! Thanks :D

jeanroyce:
Maybe this time is correct, maybe 😅

PR #333, opened 2020-04-24
Chapter3/homework/9

mihailfox:
changed comment just to mess with a certain person :)


Src/BootCamp.Chapter/TransactionModels/Transaction.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace BootCamp.Chapter
+{
+    public class Transaction
+    {
+        public Shop Shop { get; set; }
+        public Item Item { get; set; }
+        public DateTime DateTime { get; set; }
+
+        public Transaction()
```
Almantask, line 16: No need to have an empty ctor.

Src/BootCamp.Chapter/Csv/CsvBase.cs
```diff
@@ -0,0 +1,117 @@
...
+        public string Footer { get; protected set; } = string.Empty;
+        public CsvDelimiter Delimiter { get; protected set; } = CsvDelimiter.Comma;
+        protected string FileName { get; set; }
+        protected bool HasHeader { get; set; }
+        protected bool HasFooter { get; set; }
+        protected IList<CsvRow> Rows { get; set; } = new List<CsvRow>();
+
+        public CsvBase(string fileName)
+        {
+            if (!fileName.IsValid())
```
Almantask: Overkill to have an extension method for validating a string like this. It's not much different from string.IsNullOrEmpty, however what is a valid fileName I would need to dig. In fact, I would need to go as far as extensions.

Extension methods are used for giving extra functionality to existing classes. They are not supposed to contain business logic, just that data structure-specific logic. IsValid is logic that belong to CsvBase and not string.

Src/BootCamp.Chapter/Csv/CsvBase.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                throw new ArgumentException("fileName cannot be null or empty");
+            }
+
+            FileName = fileName;
+            Delimiter = delimiter;
+            HasHeader = hasHeader;
+        }
+
+        public CsvBase(string fileName, CsvDelimiter delimiter, bool hasHeader, bool hasFooter)
```
Almantask: Remove the first ctor and instead have optional 4th parameter with a default value...
```cs
public CsvBase(string fileName, CsvDelimiter delimiter, bool hasHeader, bool hasFooter = default)
{
  ...
}
```

Src/BootCamp.Chapter/Csv/CsvBase.cs
```diff
@@ -0,0 +1,117 @@
...
+                throw new ArgumentException("fileName cannot be null or empty");
+            }
+
+            FileName = fileName;
+            Delimiter = delimiter;
+            HasHeader = hasHeader;
+            HasFooter = hasFooter;
+        }
+
+        protected bool TryParseRow(string input, out CsvRow csvRow)
```
Almantask: I wonder why doesn't this belong to CsvRow if it parses it?

Src/BootCamp.Chapter/Csv/CsvBase.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (!input.IsValid())
+            {
+                return false;
+            }
+
+            var builder = new StringBuilder();
+            var isQuote = false;
+
+            foreach (var letter in input)
```
Almantask: Split into a function of its' own. ``CreateRowString`` or similar.

Src/BootCamp.Chapter/Csv/CsvBase.cs
```diff
@@ -0,0 +1,117 @@
...
+            var firstColumn = true;
+            var builder = new StringBuilder();
+
+            foreach (var field in csvRow)
+            {
+                if (field != null)
+                {
+                    if (!firstColumn)
+                    {
+                        builder.Append((char)Delimiter).Append(spaceChar);
```
Almantask: Choo choo! It's a train! 
Try to split those longer chained statements keeping only 1 chain per line.
```
builder.Append((char)Delimiter)
        .Append(spaceChar);
```

Src/BootCamp.Chapter/Csv/CsvDelimiter.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace BootCamp.Chapter.Csv
+{
+    public enum CsvDelimiter
```
Almantask: Sometimes it's better to just use a constant instead of enum if we want to get a value. It's not a tag, it's rather a named constant.

Src/BootCamp.Chapter/Csv/CsvRow.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter.Csv
+{
+    public class CsvRow : List<string>
```
Almantask: This is a classical example of bad inheritance. Are you extending List<string> functionality or are you just using it? Is ``CsvRow`` a ``List<string>``? The answer is no. CsvRow **has** List<string>. So the relation should be composition and not inheritance. Prefer composition over inheritance.

Almantask (changes requested):

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: IsValid just checks the string if it's null, white space or empty

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: That's how I did it the first time then I read on codacy test (I have my own account) we should not use optional parameters in constructors

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: Yes, that's where it should be, I forgot to  move it after I implemented CsvRow class

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: I thinks this is even better, what do you think?
```cs
builder
       .Append((char)Delimiter)
       .Append(spaceChar);
```

Src/BootCamp.Chapter/TransactionModels/Transaction.cs
mihailfox: I call it it in the TryParse method to init the new object

Src/BootCamp.Chapter/Csv/CsvRow.cs
mihailfox: when I initially wrote this an `CsvRow` was just a `List<string>` and a collection of List<string>'s would have been `List<List<string>>` . To me this looked ugly/unreadable and the solution was the `CsvRow` class inheriting `List<string>`. CsvRow is a list of strings AND Microsoft recommends to do it this way... I will add the link if I can find it again. There is a stackoverflow discussion about it that's how I got to microsoft page

Src/BootCamp.Chapter/TransactionModels/Transaction.cs
Almantask: Make get-only properties and use a ctor- cleaner that way. If the object should not change- make it that way.

Src/BootCamp.Chapter/Csv/CsvBase.cs
Almantask: yup, and that is alraedy done for you- string.IsNullOrWhitespace

Src/BootCamp.Chapter/Csv/CsvBase.cs
Almantask: I find that hard to justify. Many things have a right spot. If you can have a ctor with default parameter, what wrong with that?

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: I thought about how to refactor it since I first wrote it couple of days ago but I don't see a proper way to do it. Can you shed some light on how to do it?

Almantask:
```cs
protected bool TryParseRow(string input, out CsvRow csvRow)
{
	        if (string.IsNullOrWhitespace(input))
            {
                return false;
            }
			
			// have string builder inside Csvrow()
            row = new CsvRow();

            foreach (var letter in input)
            {
				
				// move this logic to row itself.
				// like this: row.TakeLetter(letter);
				var isQuote = letter == '\"';
                if(!isQuote)
				{
					// internally will use SB
					row.AppendLetter(letter);
				}
                else if (letter == (char)Delimiter)
                {
					// internally will use SB
                    row.NextWord();
                }
            }
			
			csvRow.Add(builder.ToString().Trim());
            return true;
        }
}
```

Src/BootCamp.Chapter/Csv/CsvBase.cs
mihailfox: I think I've been using IsValid as a stanalone method since Chapter-1 and I would have swore that IsNullOrWhitespace does not check for empty string. I allways used both IsNullOrWhitespace || IsNullOrEmpty

Oh well it's better late than never 😄

Almantask:
This should have been approved a long time ago, regardless of final result, a lot of good effort was put and that is enough to practice the topic 😄

PR #334, opened 2020-04-24
Chapter1/homework/5

JohnGalt03:
Honestly, I'm still alot confused because I based most of my answers off the internet, but I'll try studying it to learn more


Almantask:
Note that you're trying to merge into master. You should merge to the same branch that you have, except in Csinn

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <param name="array">Input array in a random order.</param>
+        public static void Sort(int[] array)
+        {
+            if (array == null || array.Length == 0) return;
+            for (int i = 0; i < array.Length; i++)
+            {
+                for (int j = i + 1; j < array.Length; j++)
+                {
+                    if (array[j] < array[i])
+                    {
```
Almantask, line 20: This is a swap. You can extract this part to a function, as it is repeated also in reverse array.

PR #335, opened 2020-04-25
Chapter2/Homework/4

jeanroyce:


Src/BootCamp.Chapter/Logger/LogConsole.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Threading;
+
+namespace BootCamp.Chapter.Logger
+{
+    public class LogConsole : ILogger
```
Almantask: ConsoleLoggerr is a better name.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,78 @@
...
+                var name = Console.ReadLine();
+                
+                Print.Log.Log("What is your surname?");
+                var surName = Console.ReadLine();
+                
+                age: Print.Log.Log("How old are you?");
+                if (!int.TryParse(Console.ReadLine(), out var age))
+                {
+                    Print.Log.Error(Constants.Invalid);
+                    goto age;
```
Almantask: Yikes! Avoid goto statements!

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,78 @@
...
     class Program
     {
         static void Main(string[] args)
         {
+            Print.Log.Warn("Program started.");
+            CalculateBmi();
+            Print.Log.Warn("Program finished.");
+        }
+
+        private static void CalculateBmi()
```
Almantask: instead of having raw code all in one place, split unrelated parts into functions, like you did in homework 3 of chapter 1.

Almantask (changes requested):
Everything looks fine except a few things:
1) DON'T use goto
2) Keep on naming variables with conventional names: methods- verbs; variables/classes- nouns.
3) Split big functions into steps so that big function and each step does only one thing and has no side effects.

Src/BootCamp.Chapter/BmiCalculator.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using BootCamp.Chapter.Logger;
+
+namespace BootCamp.Chapter
+{
+    public static class BmiCalculator
+    {
+        private static readonly ILogger ConsoleLogger = new ConsoleLogger();
```
Almantask: With OOP, forget static. Static = single instance. You might want multiple instances of Bmi calculator.

Src/BootCamp.Chapter/BmiCalculator.cs
```diff
@@ -0,0 +1,74 @@
...
+        private static readonly ILogger FileLogger = new FileLogger(@"Log.txt");
+        public static readonly ILogger Logger = (Constants.DebugLevel == "Console") ? ConsoleLogger : FileLogger;
+        
+        public static void CalculateBmi()
```
Almantask: not static

Almantask (changes requested):
Please remove all the static.

jeanroyce:
I made some changes to code, maybe a little overkill to this homework. Now Bmi calculator is a object, and a new function get all data for calculation.

Src/BootCamp.Chapter/BmiCalculator.cs
jeanroyce: You say, call the function as an object?
Something like this?
```c#
var bmi = new BmiCalculator();
bmi.Calculate();
```

Src/BootCamp.Chapter/CalculateBmi.cs
```diff
@@ -0,0 +1,76 @@
...
+                               $"is {bmi.Person.Age} year old, his weight is {bmi.Weight} " +
+                               $"kg and his height is {bmi.Height} cm.";
+            log.Logger(personalInfo, Log.Level.Log);
+            var bmiValue = $"IBM: {bmi.Bmi}";
+            log.Logger(bmiValue, Log.Level.Log);
+        
+            log.Logger("Program finished.", Log.Level.Warn);
+        }
+        
+        private static ILogger SetDebugMode()
```
Almantask: I don't see the point of this. It should come from outside. Public or injection (via ctor)

Src/BootCamp.Chapter/CalculateBmi.cs
```diff
@@ -0,0 +1,76 @@
...
+            do
+            {
+                if (int.TryParse(Console.ReadLine(), out var intValue)) return intValue;
+                log.Logger(Constants.Invalid, Log.Level.Error);
+            } while (true);
+        }
+        
+        private static float ParseToFloat()
+        {
+            var log = SetDebugMode();
```
Almantask: This function doies too many things (like other functions which sets debug mode. It might be to show an example, but this is a big code smell and is not good.

Almantask: Apply the mode consistently, across the class, instead of individually, per method.

Src/BootCamp.Chapter/CalculateBmi.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using BootCamp.Chapter.Calculator;
+using BootCamp.Chapter.Logger;
+
+namespace BootCamp.Chapter
+{
+    public class CalculateBmi
```
Almantask: I think this is a demo class. So you should name it like so.
Also, use nouns for naming classes.

Src/BootCamp.Chapter/Calculator/Bmi.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace BootCamp.Chapter.Calculator
+{
+    public class BmiCalculator
```
Almantask: This is not a calculator, this is a calculation to. In any case, it's wrong.
It's bad, because for bmi, you do not need a whole person. Weight and height is all you need.
Bmi itself should be something that is returned and not stored in an object.

Almantask: For a class that has no state, that just takes input and returns output, it's perfectly fine to have it as static.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,15 +1,11 @@
...
+﻿namespace BootCamp.Chapter
 {
     class Program
     {
         static void Main(string[] args)
         {
-            BmiCalculator.Logger.Warn("Program started.");
-            BmiCalculator.CalculateBmi();
-            BmiCalculator.Logger.Warn("Program finished.");
+            var bmiInstance = new CalculateBmi();
```
Almantask: Instance is redundant. Use real life words for naming stuff.
It's a calculator. Since the scope is just 2 lines, that is a good enough name to clarify the intention.

Almantask (changes requested):

jeanroyce:
First of all thanks!! I appreciate your feedback in all my homework.

Sorry for my mess code :( I made some change, tried to not use knowledge from chapter 3.

Src/BootCamp.Chapter/Calculator/Bmi.cs
```diff
@@ -1,26 +1,12 @@
 ﻿namespace BootCamp.Chapter.Calculator
 {
-    public class BmiCalculator
+    public static class Bmi
```
Almantask, line 3: Don't lie with code :) This is not bmi, this is calcular for it.

Src/BootCamp.Chapter/Logger/FileLogger.cs
```diff
@@ -7,9 +7,9 @@ public class FileLogger : ILogger
     {
         private readonly string _file;
 
-        public void Logger(string message, Log.Level logLevel)
+        public void Logger(string message, LogLevel logLevel)
```
Almantask, line 10: Nesting classes like that is smart 👍

Almantask (changes requested):

PR #336, opened 2020-04-25
Chapter1/homework/8

DavidKatona:
This one was a real challange. It gave me a lot of headaches figuring out how to pass all tests. I also had to do some cleanup on the code I've pasted over from my previous homework so the tests did not fail. Overall, this was a fun, although sometimes frustrating experience but I think it turned out decently. Definitely not the best but I tried to keep code smell to the minimal and refactored all I could to make it more readable. :)


Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -1,40 +1,87 @@
...
         /// Padding means how many spaces will a message be wrapped with.
-        /// Table itself is made of: "+-" symbols. 
+        /// Table itself is made of: "+-" symbols.
         /// </summary>
         public static string Build(string message, int padding)
         {
-            return "";
+            if (string.IsNullOrWhiteSpace(message))
+            {
+                return "";
```
Almantask, line 21: Prefer string.Empty over "".

DavidKatona: Thanks you. I'll remember that! :)

PR #337, opened 2020-04-25
Homework 3: Functions

lukas-ostrovskis:
Functions for calculating BMI in the Lesson3 class.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Globalization;
+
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
+            {
+                string name = PromptString("Name: ");
```
Almantask, line 14: The body of for loop could be a method as well- ProcessPersonBmi or something like that.

Almantask (changes requested):

PR #338, opened 2020-04-25
Chapter1/Homework/1and2

hmyers94:
My solution for homework 1 and 2.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,33 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
+            int age;
+            double weight;
+            double height;
+
+            for (int i = 0; i < 2; i++)
+            {
+
+                Console.WriteLine("Name: ");
+                name = Console.ReadLine();
+
+                Console.WriteLine("Age: ");
+                age = Int32.Parse(Console.ReadLine());
+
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,33 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
```
Almantask, line 9: Don't declare all variables at the same place. YOu should declare them right next to where you plan to use them (next to parse. It's not good, because you will have to scroll from declaration back to consumptions and there is simply no point to put the responsibility of remembered about them across a whole function scope.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,33 @@ class Program
...
+            int age;
+            double weight;
+            double height;
+
+            for (int i = 0; i < 2; i++)
```
Almantask, line 14: Well done for using a loop.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,33 @@ class Program
...
+            {
+
+                Console.WriteLine("Name: ");
+                name = Console.ReadLine();
+
+                Console.WriteLine("Age: ");
+                age = Int32.Parse(Console.ReadLine());
+
```
Almantask, line 22: Neat separation of unrelated code blocks with a blank line.

PR #341, opened 2020-04-26
Max Chapter3/homework/7 and8

HellHunterMax:
LINQ Query Syntax is still hard to do for me but i do see the use of it.


Src/BootCamp.Chapter/LINQExtension.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public static class LINQExtension
+    {
+        public static IEnumerable<T> SnapFingers<T>(this IEnumerable<T> source)
+        {
+            List<T> list = new List<T>();
+
+            int i = 0;
+
+            foreach (var element in source)
```
Almantask, line 13: This is ``source.Count()``

Almantask: You should have used LINQ. Also there is no random factor what gets selected- always the first half.

Almantask: Can you implement a select using LINQ :) ?

Almantask (changes requested):
Double check SnapFingers function

Src/BootCamp.Chapter/LINQExtension.cs
HellHunterMax: woeps forgot about this.  missed the using LINQ thought it was weird i could not use Count().
can you please re review it for me thanks kai!

PR #343, opened 2020-04-26
Chapter1/Homework/3

Dreniak:


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -17,25 +17,25 @@ public static class Checks
         public static int PromptInt(string message)
         {
             // To do: call your implementation.
```
Almantask, line 19: Comments can now be removed.

PR #347, opened 2020-04-28
Chapter 1/Homework 3 

Poordevv:
I was a little confused at first but eventually, I figured I had to change my previous homework 1 & 2 source code to match up with Checks.cs. Anyway, if your reading this, have a great day!


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,27 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation.
```
Almantask, line 19: Comments can be removed after implementation.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
+            {
+                var name = promptStringValue("Name: ");
```
Almantask, line 13: Loop body can be replaced with a function to ProcessPersonBmi or similar

Almantask:
> I was a little confused at first but eventually, I figured I had to change my previous homework 1 & 2 source code to match up with Checks.cs. Anyway, if your reading this, have a great day!

Cheers! 😄

Src/BootCamp.Chapter/Checks.cs
Poordevv: Removed.

Almantask:
Looks good 😄

PR #348, opened 2020-04-28
Chapter1/homework/9

DavidKatona:
This was a fun exercise. I enjoyed the googling I had to do to figure out conversion between decimal and binary and vice versa. I tried using only an image or something that told me how to do both but did not show any code or implementation. :)

I also learned pattern matching which I've found really useful!

### DIFF
--- a/Src/BootCamp.Chapter/ArrowMovement.cs
+++ b/Src/BootCamp.Chapter/ArrowMovement.cs
@@ -18,7 +18,14 @@ namespace BootCamp.Chapter
         /// <returns>One of the arrow characters. '↥' by default.</returns>
         public static char GetIndicator(char symbol)
         {
-            return '-';
+            return char.ToLower(symbol) switch
+            {
+                'w' => '↥',
+                'a' => '↤',
+                's' => '↧',
+                'd' => '↦',
+                _ => '↥'
+            };
         }
     }
 }
--- a/Src/BootCamp.Chapter/BinaryConverter.cs
+++ b/Src/BootCamp.Chapter/BinaryConverter.cs
@@ -11,25 +11,29 @@ namespace BootCamp.Chapter
 
         public static long ToInteger(string binary)
         {
-            var integer = 0;
+            var result = 0; // We assume that the resulting number will be 0
 
-            if (!string.IsNullOrEmpty(binary))
+            if (string.IsNullOrEmpty(binary)) return result;
+            
+            var indexofLastBit = binary.Length - 1;
+            var exponent = 0; // The numerical place of a bit going from right to the left which is used in calculating the result
+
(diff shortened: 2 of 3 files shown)


Src/BootCamp.Chapter/ArrowMovement.cs
```diff
@@ -18,7 +18,14 @@ public static class ArrowMovement
         /// <returns>One of the arrow characters. '↥' by default.</returns>
         public static char GetIndicator(char symbol)
         {
-            return '-';
+            return char.ToLower(symbol) switch
```
DavidKatona, line 21: I wasn't familiar with pattern matching at all and I was reluctant to use it. Eventually, I did some reading and holy, this is really useful!

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -6,14 +6,45 @@ namespace BootCamp.Chapter
 {
     public static class BinaryConverter
     {
+        private const int OnBit = 1;
+        private const int OffBit = 0;
+
         public static long ToInteger(string binary)
         {
-            return 0;
```
DavidKatona, line 11: Originall I only had:
`        public static long ToInteger(string binary)
        {
            var integer = 0;

            if (!string.IsNullOrEmpty(binary))
            {
                try
                {
                    integer = Convert.ToInt32(binary, 2);
                }
                catch (Exception)
                {
                    throw new InvalidBinaryNumberException(binary);
                }
            }
                
            return integer;
        }`

but I felt like I've cheated on the homework so I decided to implement my own logic.

Src/BootCamp.Chapter/ArrowMovement.cs
Almantask: damn right 👍

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -6,14 +6,45 @@ namespace BootCamp.Chapter
...
+            var result = 0; // We assume that the resulting number will be 0
```
Almantask, line 14: Why would we make such an assumption? I think a better comment would be...
```
Default number is 0.
```
But I would remove the comment altogether, because it's a pretty standard thing to do it like this.

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -6,14 +6,45 @@ namespace BootCamp.Chapter
...
+
+            if (string.IsNullOrEmpty(binary)) return result;
+            
+            var indexofLastBit = binary.Length - 1;
+            var exponent = 0; // The numerical place of a bit going from right to the left which is used in calculating the result
```
Almantask, line 19: Don't inline comments. Write them above this line of code.

PR #349, opened 2020-04-29
Chapter1/homework/6

JohnGalt03:
Spammed step into like a madman
image


Almantask:
I hope you enjoyed it. Remember, there is nothing that a good bit of googling and debugging cannot do. 😄

PR #350, opened 2020-04-29
Chapter3/homework/3

jeanroyce:
Not much to say, done with small changes :)

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -7,85 +7,117 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
-        private Weapon _weapon;
-        public void SetWeapon(Weapon weapon)
-        {
+        public Weapon Weapon { get; set; }
+        public Headpiece Head { get; set; }
+        public Chestpiece Chest { get; set; }
+        public Shoulderpiece LeftShoulder { get; set; }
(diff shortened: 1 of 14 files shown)


jeanroyce:
Just in case,  I made Null or Empty validations in public methods and strings in ctor.

Almantask:
> Just in case, I made Null or Empty validations in public methods and strings in ctor.

Which is a very good thing 😃

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -2,90 +2,111 @@
...
-        }
-
-        private Shoulderpiece _leftShoulder;
-        public void SetLeftShoulder(Shoulderpiece should)
-        {
-
+            get
+            {
+                UpdateDefense();
+                return _weight;
```
Almantask, line 35: Weight is not defense.

Src/BootCamp.Chapter/Item.cs
```diff
@@ -1,26 +1,19 @@
...
+        public float Weight { get; }
+        
         public Item(string name, decimal price, float weight)
         {
-            _name = name;
-            _price = price;
-            _weight = weight;
+            if (string.IsNullOrEmpty(name)) throw new ArgumentNullException(nameof(name));
+            Name = name;
+            Price = price;
```
Almantask, line 15: Could have done the same for price and weight (cannot be negative or 0)

PR #351, opened 2020-04-29
Chapter1/homework/4

Dreniak:
I took a different approach on dealing with invalid user input. Rather than returning an "invalid" number or string if the user made a mistake, having to basically go through all the aplication to find out what was wrong, and getting the error outputted either way in the end, I chose to declare the error and ask the question again until a valid answer is submitted.

The main advantages of this approach are that if the user mistype the answer, it would be able to immediately fix it, also we can have a simpler function to calculate the BMI, as there's no in-app way to provide an invalid parameter.

It didn't went outside the scope of the class, but if it needs to be as specified in the assignment I'd have no problem changing it.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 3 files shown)


Almantask:
> I took a different approach on dealing with invalid user input. Rather than returning an "invalid" number or string if the user made a mistake, having to basically go through all the aplication to find out what was wrong, and getting the error outputted either way in the end, I chose to declare the error and ask the question again until a valid answer is submitted.
> 
> The main advantages of this approach are that if the user mistype the answer, it would be able to immediately fix it, also we can have a simpler function to calculate the BMI, as there's no in-app way to provide an invalid parameter.
> 
> It didn't went outside the scope of the class, but if it needs to be as specified in the assignment I'd have no problem changing it.

It's definitely the right way to do it. However, you could have done both, by putting this logic into another function like ``TryPromptInt`` which calls the function that is required to be made the way it should be. If you look at the bottom of the PR, you will see "tests failing". That's exactly why.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,84 @@
...
+        public static void Demo()
+        {
+            Console.WriteLine("Body-Mass Index (BMI) Calculator");
+            Console.WriteLine("");
+
+            ProcessPersonBMI();
+            ProcessPersonBMI();
+        }
+
+        private static void ProcessPersonBMI()
```
Almantask, line 20: Try to keep it conventional and don't have more than 1 capitalized letter in a row. Use PascalCase.
``ProcessPersonBMI`` should be ``ProcessPersonBmi``

Almantask:
This is good enough for now, carry on 😄

PR #354, opened 2020-04-30
Chapter3/homework/4

jeanroyce:
This last lesson was hard, I don't know if I fully understood it. I think I need to get used to understand more about covariant and contravariant.


Almantask:
> This last lesson was hard, I don't know if I fully understood it. I think I need to get used to understand more about covariant and contravariant.

This topic is often misunderstood (or not even heard of) by even senior developers. So a beginner being exposed to it is quite something. Don't worry too much about it, come back to it later at later times and you will get it 👍

Src/BootCamp.Chapter/Students/Student.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using BootCamp.Chapter.Subjects;
+using BootCamp.Chapter.Teachers;
+
+namespace BootCamp.Chapter.Students
+{
+    public abstract class Student : IStudent
```
Almantask, line 7: Not abstract, because a student can exist by itself. It doesn't have anything that is not unimplemented.

Almantask (approved):
Based on what you did it wasn't that hard. Well done 😸

PR #355, opened 2020-04-30
Well, i just modified it to be more efficient so i used the same vari…

ozlarus-creator:
…ables for the two persons and made a loop on it. It takes the first data input and prints it, then it clears the values and starts to the second person. After the second person the looper value will be 3 so the program ends.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,34 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+
+            string name = "Tom Jefferson";
+            int age;
+            int looper = 1;
+            float weight, mheight;
+
+            do
+            {
+                Console.WriteLine("Enter Name");
+                name = Console.ReadLine();
+
+                Console.WriteLine("Enter Age");
+                age = Convert.ToInt32(Console.ReadLine());
+
+                Console.WriteLine("Enter Weight");
+                weight = float.Parse(Console.ReadLine());
+
+                Console.WriteLine("Enter Height");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,34 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+
+            string name = "Tom Jefferson";
```
Almantask, line 10: Avoid delcaring variables at the top. Declare them as close as possible to where they are used.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,34 @@ class Program
...
+            int age;
+            int looper = 1;
+            float weight, mheight;
+
+            do
+            {
+                Console.WriteLine("Enter Name");
+                name = Console.ReadLine();
```
Almantask, line 18: string name = ...

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,34 @@ class Program
...
+
+                Console.WriteLine("Enter Age");
+                age = Convert.ToInt32(Console.ReadLine());
```
Almantask, line 21: int age = ..

Almantask (approved):
It's good, just keep 2 things in mind:
1) don't declare all variables in one place- declare them as close as possible to where they are used. This reduces the amount of time we need to scroll up and down to refer back to the initial declaration of the variable. It also fits well in our vision scope.
2) Please try to pick a proper name for PR, like Homework 1. Use comments section to add remarks.

Regardless of that, well done! Welcome to the boot camp and carry on.

PR #359, opened 2020-05-01
Chapter1/homework/10

DavidKatona:
This was an interesting homework. I had a good time scratching my head why "x % y" did not return expected numbers when "y" was negative.

Turns out, % in C# stands for remainder and not for modulo which behaves differently when a divisior is negative. I still not fully understand what happens behind the scenes so I'll have some good additional headscratches this afternoon.

### DIFF
--- a/Src/BootCamp.Chapter/CaesarCipher.cs
+++ b/Src/BootCamp.Chapter/CaesarCipher.cs
@@ -11,36 +11,40 @@ namespace BootCamp.Chapter
     /// </summary>
     public static class CaesarCipher
     {
+        private const char UpperA = 'A';
+        private const char LowerA = 'a';
+        private const int numberofKeys = 26;
+
         public static string Encrypt(string message, byte shift)
         {
-            return ApplyCipher(message, shift);
+            return CiperMessage(message, shift);
         }
 
         public static string Decrypt(string message, byte shift)
         {
-            return ApplyCipher(message, -shift);
+            return CiperMessage(message, 26 - shift);
         }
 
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -4,19 +4,47 @@
...
 namespace BootCamp.Chapter
 {
+    // ToDo: Change code so it only allows letters from an alphabet.
+
     /// <summary>
     /// Caesar cipher which supports all ASCII character (256 in total)
     /// </summary>
     public static class CaesarCipher
     {
+        private const char UpperA = 'A';
```
Almantask, line 14: This variable name is absolutely redundant. It still doesn't explain what the a's are used for.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -4,19 +4,47 @@
...
         public static string Encrypt(string message, byte shift)
         {
-            return message;
+            return CiperMessage(message, shift);
         }
 
         public static string Decrypt(string message, byte shift)
         {
-            return message;
+            return CiperMessage(message, numberofKeys - shift);
```
Almantask, line 25: Spot on 👍

Almantask (approved):
Well done! Just try to name variables based on what they are used for and not what they are.

Src/BootCamp.Chapter/CaesarCipher.cs
DavidKatona: Yes, that makes sense! I'll keep this in mind! :)

PR #363, opened 2020-05-05
Chapter2/homework/2

DavidKatona:
I got real excited with this one as I'm working on a small game in Unity and the inventory system was a long pending component of that. This really helped me with figuring out how it should be built up.

It's also pretty amazing how straightforward it can become to work with complex data when you work with classes of your own rather than simple array of strings.

Also, here are the tests from the extra parts of this homework.
image

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -3,24 +3,51 @@
     public class Inventory
     {
         private Item[] _items;
+
         public Item[] GetItems()
         {
             return _items;
         }
 
-        public Inventory()
+        private int _size;
+
+        public int GetSize()
         {
-            _items = new Item[0];
+            return _size;
         }
 
-        public Item[] GetItems(string name)
+        private int _remainingSpace;
+
+        public int GetRemainingSpace()
+        {
+            return _remainingSpace;
+        }
+
+        public Inventory(int size)
+        {
+            _size = size;
+            _items = new Item[size];
+            _remainingSpace = size;
+        }
+
+        public Item[] GetItem(string name)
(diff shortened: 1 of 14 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,10 +1,58 @@
...
     {
         static void Main(string[] args)
         {
+            Player playerOne = new Player("Arthas Menethil", 50, 10, new Inventory(), new Equipment());
+            EquipItems_And_DisplayStatistics(playerOne);
+        }
+
+        // This code is ugly but it properly displays the stats of the player after items are equipped.
+        private static void EquipItems_And_DisplayStatistics(Player player)
+        {
```
DavidKatona, line 16: This is real ugly code there, but I've wanted a quick way of testing whether equipping items change stats.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -2,25 +2,64 @@
 {
     public class Inventory
     {
+        private const int defaultInventorySize = 10;
```
DavidKatona, line 5: I know it wasn't part of the requirements that an inventory should have a size but I've wanted to include that extra bit in the code so I can get a better understanding of how it should be done in a real game with certain amount of slots in an inventory.

It also makes adding/removing items a bit easier as I don't have to cache a copy of the inventory and then recreate a new inventory with a size that accounts for that new/deleted item. Expanding the inventory with upgrades, however, will definitely need this, though it wasn't part of this hw.

Almantask (approved):
Looks good. Just consider splitting big functions into smaller parts.

PR #364, opened 2020-05-05
Chapter2/homework/3

DavidKatona:
Not sure if this was the intended aim of this exercise but I tried implementing polymorphism and thus using abstract classes and methods as much as I could.

Now, in a real world application, the factory classes should have a different implementation of the assemble method. In a real world scenario, I guess you'd have a collection of components that are given to the factories and the factory processes these objects and adds them to a DesktopComputer.

For simplicity's sake, I've just statically coded these components onto the abstract methods of the ms and mac factories. Btw, how would you go on about implementing the previous method I've mentioned?

Also, here's a screenshot of what each step in the assembly outputs to the console depending on which factory handled the assembly.
image

### DIFF
--- a/Src/BootCamp.Chapter/Computer/Gpu.cs
+++ b/Src/BootCamp.Chapter/Computer/Gpu.cs
@@ -47,4 +47,11 @@
             _gpuClock = gpuClock;
         }
     }
+
+    public enum GPUMemoryType
+    {
+        GDDR5,
+        GDDR5X,
+        GDDR6
+    }
 }
--- a/Src/BootCamp.Chapter/Computer/Gpu.cs
+++ b/Src/BootCamp.Chapter/Computer/Gpu.cs
@@ -9,9 +9,9 @@
             return _chipsetManufacturer;
         }
 
-        private string _memoryType;
+        private GPUMemoryType _memoryType;
 
-        public string GetMemoryType()
+        public GPUMemoryType GetMemoryType()
         {
             return _memoryType;
         }
@@ -38,7 +38,7 @@
         }
 
 
-        public Gpu(string id, string manufacturer, string modelName, string chipsetManufacturer, string memoryType, int memorySize, int memoryClock, int gpuClock) : base (id, manufacturer, modelName)
+        public Gpu(string id, string manufacturer, string modelName, string chipsetManufacturer, GPUMemoryType memoryType, int memorySize, int memoryClock, int gpuClock) : base (id, manufacturer, modelName)
         {
             _chipsetManufacturer = chipsetManufacturer;
             _memoryType = memoryType;
--- a/Src/BootCamp.Chapter/Computer/Gpu.cs
+++ b/Src/BootCamp.Chapter/Computer/Gpu.cs
@@ -1,6 +1,50 @@
 ﻿namespace BootCamp.Chapter.Computer
 {
-    public class Gpu
+    public class Gpu : Component
     {
+        private string _chipsetManufacturer;
+
+        public string GetChipsetManufacturer()
+        {
+            return _chipsetManufacturer;
+        }
+
+        private string _memoryType;
+
+        public string GetMemoryType()
+        {
+            return _memoryType;
+        }
+
+        private int _memorySize;
+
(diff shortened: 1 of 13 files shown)


Src/BootCamp.Chapter/Computer/Gpu.cs
```diff
@@ -1,6 +1,52 @@
...
+
+        private int _gpuClock;
+        public int GetGpuClock()
+        {
+            return _gpuClock;
+        }
+
+
+        public Gpu(string id, string manufacturer, string modelName, string chipsetManufacturer, GPUMemoryType memoryType, int memorySize, int memoryClock, int gpuClock) : base (id, manufacturer, modelName)
+        {
```
DavidKatona, line 37: I got a little bit carried away with this list of specifications :D

Src/BootCamp.Chapter/Computer/Gpu.cs
```diff
@@ -1,6 +1,52 @@
...
+            _chipsetManufacturer = chipsetManufacturer;
+            _memoryType = memoryType;
+            _memorySize = memorySize;
+            _memoryClock = memoryClock;
+            _gpuClock = gpuClock;
+        }
+    }
+
+    public enum GPUMemoryType
+    {
```
DavidKatona, line 47: Also, are enums fine in this case or would you use something else here? My main goal was to avoid having to guess for memory types. Similar stuff to this is implemented at other components as well.

I guess in a real app, you'd have a database or something that would store these right?

Src/BootCamp.Chapter/Factory.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace BootCamp.Chapter
+{
+    public abstract class Factory
+    {
+        protected Factory() { }
+
+        public DesktopComputer Assemble()
+        {
+            DesktopComputer computer = new DesktopComputer();
```
DavidKatona, line 14: At first, I wanted to make the Assemble method abstract thus both ms and mac factories would have had different implementations.

But seeing that I had to go through the same steps when the assembly happens, I've made the steps abstract methods instead.

Again, this could probably be done in a much more elegant way and components should be handled to the Assemble method so it's not statically coded inside the abstract methods.

Src/BootCamp.Chapter/MacFactory.cs
```diff
@@ -5,11 +5,58 @@
...
+        {
+            Motherboard motherboard = new Motherboard("ARS-Z390-E-G", "Asus", "ASUS ROG STRIX Z390-E GAMING", "Z390-E", "Intel", SocketType.LGA1151_300);
+            Console.WriteLine($"Installing motherboard {motherboard.GetModelName()} onto motherboard.");
+            computer.SetMotherboard(motherboard);
+        }
+
+        protected override void InstallCpu(DesktopComputer computer)
+        {
+            Cpu cpu = new Cpu("INTEL-I5-9600K", "Intel", "INTEL Core i5-9600K", SocketType.LGA1151_300, 6, 6, 3700);
+            Console.WriteLine($"Applying thermal paste to cpu socket... and installing {cpu.GetModelName()} onto motherboard.");
```
DavidKatona, line 31: Yes, I know you apply it to the CPU not the socket, but hey, it's Apple right? 😄

Src/BootCamp.Chapter/Factory.cs
```diff
@@ -0,0 +1,34 @@
+﻿using BootCamp.Chapter.Computer;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public abstract class Factory
+    {
+        protected Factory() { }
```
Almantask, line 10: Not needed, can be removed.

Src/BootCamp.Chapter/Factory.cs
Almantask: That was the point of the exercise. The exercise explicitly states "Both factories have the same flow of assembling parts, just individual steps are different".
So yes, this is the correct approach.

Almantask (approved):
Looks good

PR #367, opened 2020-05-06
Chapter3 / HW 10

ghost:


Src/BootCamp.Chapter/Car.cs
```diff
@@ -1,14 +1,13 @@
...
-
-namespace BootCamp.Chapter
+﻿namespace BootCamp.Chapter
 {
     // This should have the following attribute
-    // [Textable]
+    [Textable]
     public class Car
     {
+        [Textable]
```
Almantask, line 7: You don't format individual properties in their own boxes, just the whole car.

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,30 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace BootCamp.Chapter
+﻿namespace BootCamp.Chapter
 {
     public static class TextBoxPrinter
```
Almantask, line 3: I actually like the way you split it a lot. Having a break definitely helped you a lot 😄

Almantask (approved):
Looks great!

Src/BootCamp.Chapter/Car.cs
ghost: hmm, I did but then the attribute did not work at all

Src/BootCamp.Chapter/TextBoxPrinter.cs
ghost: Thanks

Src/BootCamp.Chapter/Car.cs
ghost: then this line gives null :  `requiredAttributeTextable = AttributesGetter.GetPropertyAttribute<Textable>(prop); `

PR #370, opened 2020-05-08
Chapter2/homework/4

DavidKatona:
Here's my interpretation of the homework. Hope this one's fine. :)

I have to say, I'll have to practice interfaces a lot more. I'll probably end up watching your lesson a couple of additional times :)


Src/BootCamp.Chapter/Logging/Logger.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace BootCamp.Chapter.Logging
+{
+    public class Logger : ILogger
```
Almantask: This class is not needed. You have different types of logger, container logger, in this case, serves no purpose. It could work, but no reason to use it like that.

Src/BootCamp.Chapter/HealthCheckSimulator.cs
```diff
@@ -0,0 +1,67 @@
...
+            else
+            {
+                Console.WriteLine("Invalid value, defaulting to 'Console'.");
+                return new Logger(LogTarget.Console);
+            }
+
+            switch (logDestination)
+            {
+                case 0:
+                    return new Logger(LogTarget.Console);
```
Almantask: you should just create a specific logger, instead of setting logging target.

Src/BootCamp.Chapter/HealthCheckSimulator.cs
```diff
@@ -0,0 +1,67 @@
...
+            Logger sessionLogger = SetupLogger();
+
+            Person user = new Person(); //We create our test subject
+            ConsolePrompter.WelcomeUser();
+            SetPersonalDetails(user, sessionLogger);
+
+            Console.WriteLine($"BMI equals: {BMICalculator.CalculateBMI(user, sessionLogger)}");
+        }
+
+        private static Logger SetupLogger()
```
Almantask: Extract this method to ``LoggerFactory``

Almantask (changes requested):

Src/BootCamp.Chapter/Logging/Logger.cs
DavidKatona: I'm actually really glad you threw this back at me 😄 It made me realize that instead of using the Logger container and the target types, I could have just used ILogger in all my functions instead of passing in a specific logger type.

It was so obvious but I didn't see that. Thanks sensei!

Src/BootCamp.Chapter/HealthCheckSimulator.cs
DavidKatona: I've got rid of LogTarget completely and instead used ILogger in most of my functions. This way it looks so much nicer!

Src/BootCamp.Chapter/HealthCheckSimulator.cs
DavidKatona: Extracted!

Src/BootCamp.Chapter/ConsolePrompter.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+
+namespace BootCamp.Chapter
+{
+    public static class ConsolePrompter
+    {
+        public static ILogger SetupSession()
+        {
+            Console.Write("Please select where would you like to log for this session. 0 = Console, 1 = Logfile: ");
+
```
DavidKatona, line 11: Not sure if I did right by calling the LoggerFactories inside of ConsolePrompter (I feel like this doesn't belong here), but I did not want to place this huge block of code inside of HealthCheckSimulator.

Also, the comments might be unnecessary but I wanted to avoid magic numbers (and "1" smelled like it).

Src/BootCamp.Chapter/Logging/LoggerFactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter.Logging
+{
+    public static class LoggerFactory
```
Almantask, line 7: A factory usually works that given a key, it creates the object you want (IObject). This is not a typical factory pattern. but for now it's fine.

Almantask (approved):
Looks good now

PR #371, opened 2020-05-08
rewrote a lot of files to xml and json format as the exercise CH3/HW 11 wanted

ghost:
Rewrote a a lot of inut and ouput files to xml and json format as the challenge wanted.

Also some file could not be rewritten because they have no header or the files where at the end a rush hour must be placed. 
So these files are not rewrittten by me because I did not know how it must be done 
All files are in the output folder of the test project

-  CityItemsMax.cvs 
_  CityItemsMin.cvs 
- CityMoney Max.cvs 
- CityMoneyMin.cvs 
- FullDay.cvs 
- Night.cvs


Src/BootCamp.Chapter/ConsoleArgumentsParser.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public static class ConsoleArgumentsParser
+    {
+        public static void ArgumentsParser(string[] args)
```
Almantask, line 8: Raw code. This function is completely unorganized and should be split into private step functions. I don't want to read 60 lines of code to understand what it does.

Almantask, line 8: Function name should be a **verb** (not ok) and class name should be a **noun** (ok)

Src/BootCamp.Chapter/OutputTime.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace BootCamp.Chapter
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+
+    namespace BootCamp.Chapter
+    {
+        public class OutputTime
```
Almantask, line 13: Output is a horrible name. Please rename it so that it represent business logic.

Almantask (changes requested):

ghost:
Chips, my normal code and the new json and xml classes for the tests are mixed up. 

Is there a way the last 2 commits are deleted so this one is all about the writing of new json and xml classes and not about my code that tries to solve this challenge.

ghost:
Should I close this PR and make a new PR with only the changes  I made to the test so that they have the input and output as described in the challenge. 

Then if the tests are all allright,  I will work on  solving it.

PR #372, opened 2020-05-09
Chapter1/Homework1and2.

darwinjks:
I'm trying my best to understand github, it took me hours to pull the request. If there's a better way to make some space in the console without using writeline i would like to know. thanks


Almantask (changes requested):
Input should not be hardcoded. Input should be read from console as text and converted to int and float (or double) accordingly. Use: ``Console.ReadLine``; ``Convert.ToInt32`` or ``int.TryParse(input, out var number)``

PR #377, opened 2020-05-13
Max Chapter3/homework/9

HellHunterMax:
I did not do the extra's.
I see the added benefit of LINQ. Though it takes me so long to create them for me its not worth it yet.
My mind just wont wrap around them. I'll practice them more because i think i have to learn them.
i might do the other challenges after this has been approved.
If you have any learning tips behind practicing them would be appreciated.


Almantask:
> I did not do the extra's.
> I see the added benefit of LINQ. Though it takes me so long to create them for me its not worth it yet.
> My mind just wont wrap around them. I'll practice them more because i think i have to learn them.
> i might do the other challenges after this has been approved.
> If you have any learning tips behind practicing them would be appreciated.

Yup, it's something you have to practice. And yes, it's something that comes with time, naturally. I hope that in the Bootcamp project you will be exposed to a lot of it and, through sheer reading, it will come to you.

Almantask:
This homework is extremely hard for someone who is just starting out with LINQ. So it's amazing that you did you best and was not intimidated!

Src/BootCamp.Chapter/ArgsReader.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
+{
+    public static class ArgsReader
```
Almantask, line 5: I like the class- it's a good direction 😄

Src/BootCamp.Chapter/Command.cs
```diff
@@ -0,0 +1,185 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    public static class Command
```
Almantask: instead of having one static command class split each methods into a command of its' own. THen have a command factory.
Lastly, this class should simply be:
```
interface ICommand
{
  void Execute();
}
```
meaning which you just pass all the input through ctor and all commands can be executed in the same way.

Almantask (changes requested):
Other than massive command class, it look alright ;)

Src/BootCamp.Chapter/Command.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            else
+            {
+                throw new InvalidCommandException($"{command[1]} is not a valid parameter.");
+            }
+        }
+
+        private static IEnumerable<String> GroupedByTime(List<Transaction> transactions, DateTime[] times)
+        {
+            const string topRow = "Hour, Count, Earned";
```
Almantask: headers is a better name.

Almantask (changes requested):

Almantask:
I think LINQ itself looks just fine 😄

Src/BootCamp.Chapter/Command.cs
HellHunterMax: i think this is what you meant ?

Src/BootCamp.Chapter/ArgsReader.cs
```diff
@@ -6,32 +6,34 @@ public static class ArgsReader
...
-        private const int fileToWrite = 2;
+        private const int fileToReadInt = 0;
+        private const int commandInt = 1;
+        private const int fileToWriteInt = 2;
 
         public static void Read(string[] args)
         {
-            ArgsChecksLength(args);
-            string[] commandArr = ReadCommand(args[command]);
+            CheckArgsLength(args);
```
Almantask, line 15: It's an improvement, but better to just call it ValidateArgs(args);

Src/BootCamp.Chapter/ArgsReader.cs
```diff
@@ -6,32 +6,34 @@ public static class ArgsReader
...
+            string[] commandArr = ReadAndCheckCommand(args[commandInt]);
```
Almantask, line 16: No need the and. First validate, then read.

Src/BootCamp.Chapter/CityCommand.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    internal class CityCommand : ICommand
```
Almantask, line 7: What about it? What does it do with the City? Be more descriptive.

Almantask (approved):

Almantask:
Looks okay.

PR #379, opened 2020-05-13
Marlon M - Chapter 1: Homework 1 and 2

marlon-malheiros:
Add my tasks

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,30 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            for(int i = 0; i < 2; i++)
+            {
+                string name;
+                string surename;
+                int age;
+                double weight;
+                double height;
+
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,30 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            for(int i = 0; i < 2; i++)
+            {
+                string name;
```
Almantask, line 11: Declare variable right next to where you use them.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,30 @@ class Program
...
+                string surename;
+                int age;
+                double weight;
+                double height;
+
+                Console.Write("Enter the person's first name: ");
```
Almantask, line 17: Code right now is a blob. Try to put a blank line between unrelated blocks of code (reading, writing and different fields)

Almantask (changes requested):

PR #380, opened 2020-05-14
Chapter2/homework/1

matthew-hole:
Apologies for the absolute mess this pull request/merge is, as you can see I am still learning.

As you will see, the tests have not been correctly done and the Balances.In and Balances.Corrupt files not used. This is because I worked external to the repo to begin with before realising I had to work inside it.

Please could you review the current solution and let me know of any changes I could make and the best way of incorporating the current code to utilise the tests.


Src/BootCamp.Chapter/AccountDetails.cs
```diff
@@ -0,0 +1,67 @@
...
+        public string GetName()
+        {
+            return _accountName;
+        }
+
+        public void AddBalance(decimal balance)
+        {
+            _accountBalanceHistory.Add(balance);
+            highestBalance = Math.Max(balance, highestBalance);
+            biggestLoss = Math.Min(balance, biggestLoss);
```
Almantask, line 31: Biggest loss is calculated differently: 
You take a difference between ith and i +1 th balances and keep the result, then find the biggest such difference.

Src/BootCamp.Chapter/AccountDetails.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public string GetName()
+        {
+            return _accountName;
+        }
+
+        public void AddBalance(decimal balance)
+        {
+            _accountBalanceHistory.Add(balance);
+            highestBalance = Math.Max(balance, highestBalance);
```
Almantask, line 30: This is smart, well done

Src/BootCamp.Chapter/FileCleaner.cs
```diff
@@ -0,0 +1,48 @@
...
+                        var cleanedLine = RemovingInvalidChar(line);
+                        writer.WriteLine(cleanedLine);
+                    }
+                    
+                    line = reader.ReadLine();
+                }
+            }
+        }
+
+        private string RemovingInvalidChar(string line)
```
Almantask, line 31: *ing* is not needed.

Src/BootCamp.Chapter/FileCleaner.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class FileCleaner
+    {
+        private readonly string corruptedPath = @"..\..\..\Input\Balances.corrupted";
```
Almantask, line 10: use const instead of readonly

Src/BootCamp.Chapter/Statistics.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Xml;
+
+namespace BootCamp.Chapter
+{
+    class Statistics
```
Almantask, line 7: Should be a static class

Src/BootCamp.Chapter/Statistics.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        private decimal highestBalanceEver;
```
Almantask, line 9: Should be a part of BalanceStatistic object

Src/BootCamp.Chapter/Statistics.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public Statistics()
+        {
+            highestBalanceEver = 0;
+            personWithBiggestLoss = 0;
+            richestPerson = 0;
+            mostPoorPerson = 0;
+        }
+
+        public void CalculateStats()
```
Almantask, line 26: Should return BalanceStatistics

Src/BootCamp.Chapter/Statistics.cs
```diff
@@ -0,0 +1,103 @@
...
+            for(int i = 0; i < accountNames.Count; i++)
+            {
+                HighestBalanceEver(accountNames, i);
+                PersonWithBiggestLoss(accountNames, i);
+                RichestPerson(accountNames, i);
+                MostPoorPerson(accountNames, i);
+            }
+        }
+
+        public void DisplayStats()
```
Almantask, line 39: Does not belong to this class

Src/BootCamp.Chapter/Statistics.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            Console.WriteLine($"{highestBalanceEverName}: {ToCurrency(highestBalanceEver)}");
```
Almantask, line 41: Move the printing logic (more like string forming logic) to BalanceStatistics object

Almantask (changes requested):
Statistics class needs fixing.

PR #381, opened 2020-05-14
Chapter1/Homework/1And2

OrangeCluster:
Finished the first homework and hopefully I'm doing this correctly. Hopefully also not way too late to start these home work assignments.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -5,25 +5,28 @@ namespace BootCamp.Chapter
     class Program
     {
         static void Main(string[] args)
-        {
-            string name;
-            int age;
-            double weight;
-            double height;
-            double BMI;
+        {    
+            // BMI Calculator. Collects info from user first.
             Console.WriteLine("BMI Calculator");
             Console.WriteLine("Please enter your name: ");
-            name = Console.ReadLine();
+            string name = Console.ReadLine();
+
             Console.WriteLine("Please enter your age: ");
-            age = int.Parse(Console.ReadLine());
+            int age = int.Parse(Console.ReadLine());
+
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,24 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
```
Almantask: Declare variables right next to where you use them.

Almantask: This means not declaring all the variables in one place.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,24 @@ class Program
...
+            int age;
+            double weight;
+            double height;
+            double BMI;
+            Console.WriteLine("BMI Calculator");
+            Console.WriteLine("Please enter your name: ");
```
Almantask, line 11: right now, code looks like a blob. Try to be more subtle and improve readability by separating unrelated code parts with a blank line.

For example, unrelated parts are:
Read age
Read name
Read weight
Calculate bmi

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
OrangeCluster: Thank you for looking this over. Bad habits left over from coding 15 years ago... I've adjusted it following your recommendations.

Src/BootCamp.Chapter/Program.cs
OrangeCluster: Done. Thank you for the suggestions.

Almantask (approved):
Looks great now

PR #382, opened 2020-05-15
Chapter4/homework/1

ghost:
Maked all the tests if possible. If not possible then I added a comment why I think the code could not be tested. 

Found also a problem in the code. When you have a empty Card list andyou try to take the card from the top it fails with a IndexOutOfRange exception. As far as I can see there is no code to deal with that case


Tests/BootCamp.Chapter.Tests/CardTests.cs
```diff
@@ -0,0 +1,42 @@
+﻿using BootCamp.Chapter.Gambling;
+using Xunit;
+using static BootCamp.Chapter.Gambling.Card;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class CardTests
+    {
+        private readonly Card card;
```
Almantask: Having expectations in test suite class is a bad idea in most cases. Different tests have different expectations (per test case and not per test suite)
Should be moved to each individual tests.

Tests/BootCamp.Chapter.Tests/CardTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly Suites expectedSuite = Suites.Diamonds;
+        private readonly Ranks expectedRank = Ranks.Eight;
+
+        public CardTests()
+        {
+            card = new Card(expectedSuite, expectedRank);
+        }
+
+        [Fact]
+        public void Card_When_Given_Suite_Sets_Suite() => Assert.Equal(expectedSuite, card.Suite);
```
Almantask: These two first cases are probably not needed, because they are covered by other tests.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,50 @@
+﻿using BootCamp.Chapter.Gambling;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class DeckTest
+    {
+        // I cannot test the happy path on the constructor because the variable that  holds the contents is private
```
Almantask, line 11: It's fine, it happens indirectly through other tests.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        [Fact]
+        public void New_Given_Null_List_Throws_NullArgumentException()
+        {
+            Assert.Throws<ArgumentNullException>(() => new Deck(null));
```
Almantask: Split and and assert steps (2 lines)

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Theory]
+        [MemberData(nameof(deckExpectations))]
+        public void DrawFromTop_Given_List_Return_Last_Item(Deck deck, Card expected)
+        {
+            var actual = deck.DrawFromTop();
+
+            Assert.Equal(expected, actual);
+        }
+
+        // DrawRandom cannot be tested because it needed a mock and we did not learn that on lesson1.
```
Almantask: It can be tested. There are a few cases:
1) DrawRandom, when there is one card- draws that card
2) DrawRandom, when there are no cards- exception

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        // Shuffle cannot be tested because there is not way I can make a test that always returns the same.
```
Almantask: Test, if after shuffle, it includes all the initial cards (or the count at least).

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        // ValidateDeckSize cannot be tested because the method is private.
```
Almantask: Which is fine, because you already tested it by testing draw methods.

Almantask: Not every function has to be tested directly.

Tests/BootCamp.Chapter.Tests/HandTest.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class HandTest
+    {
+        // The constructor cannot be tested because cards is protected. 
+
+        // AddCards cannot be tested because the return type is void and Cards cannot be read outside the class.
```
Almantask: Yes, but there is a `Reveal` method, through which you can get all the cards. Not always the result of calling a function is a direct output. Sometimes you need to check the effects of the system under test through additional functions, like in this case- Reveal.

Almantask (changes requested):
Missing a few test cases and some other tests are not needed because they are covered already by other

Tests/BootCamp.Chapter.Tests/CardTests.cs
ghost: oke, deleted the code and place the code in the tests individual.

Tests/BootCamp.Chapter.Tests/CardTests.cs
ghost: deleted

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: oke, learned another thing

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: written

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: How do you mean that ?

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: oke, then I have to find out how to count the cards of a deck.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: oke

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: Overall you shoud actually have 3 steps to make the intention of documenting your test more clear:
either:
```
Deck deck = null;

Action action = () => new Deck(null);

Assert.Throws<ArgumentNullException>(action);
```
or
```
Deck deck = null;

void Create() => new Deck(null);

Assert.Throws<ArgumentNullException>(action);
```

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: True that. Just check if it doesn't throw or leave it untested.

ghost: then I leave it untested

Tests/BootCamp.Chapter.Tests/DeckTest.cs
ghost: thanks for the explanation

Tests/BootCamp.Chapter.Tests/HandTest.cs
ghost: Thanks, also changed

Tests/BootCamp.Chapter.Tests/CardTests.cs
```diff
@@ -1,32 +1,22 @@
...
-        }
-
-        [Fact]
-        public void Card_When_Given_Suite_Sets_Suite() => Assert.Equal(expectedSuite, card.Suite);
-
-        [Fact]
-        public void Card_When_Given_Suite_Sets_Ranks() => Assert.Equal(expectedRank, card.Rank);
-
         [Fact]
         public void Card_When_Equivalent_Cards_Returns_True()
```
Almantask, line 11: Card will never be equivalent to cards. Please more careful with the naming. should be Card_When_Equivalent_**Card**_Returns_True

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -12,15 +13,21 @@ public class DeckTest
         [Fact]
         public void New_Given_Null_List_Throws_NullArgumentException()
```
Almantask, line 14: string, int, list- we don't really get a contextual meaning from the type. Please give them logical meaning. In this case it's just wrong on many levels.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -12,15 +13,21 @@ public class DeckTest
...
         {
-            Assert.Throws<ArgumentNullException>(() => new Deck(null));
+            Deck deck = null;
```
Almantask, line 16: wtf. This is a distraction, it shifts my attention to a deck. Do we create deck by passing a deck?

Almantask: Instead of deck, it should have been cards.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: In this case it's supposed to be cards and not a deck.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -32,11 +39,29 @@ public void DrawFromTop_Given_List_Return_Last_Item(Deck deck, Card expected)
             Assert.Equal(expected, actual);
         }
 
-        // DrawRandom cannot be tested because it needed a mock and we did not learn that on lesson1.
+        [Fact]
+        public void DrawRandom_Given_List_One_Card_Returns_That_Card()
```
Almantask, line 43: Again you bring the list back. I don't know what is the list you a referring to, talk to me via test as if you are talking to a customer, person, wife, daughter. 

If you tell them- "there is a list in the deck"- they will look at you strange. There are cards in the deck!

In this case *_Given_One_Card*

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -32,11 +39,29 @@ public void DrawFromTop_Given_List_Return_Last_Item(Deck deck, Card expected)
...
+
+            var actual = deck.DrawRandom();
+
+            Assert.Equal(expected, actual); 
 
-        // Shuffle cannot be tested because there is not way I can make a test that always returns the same.
+        }
+
+        [Fact]
+        public void DrawRandom_Empty_List_Throws_OutOfCardExpection()
```
Almantask, line 55: Should be: *.._Given_No_Cards..*

Tests/BootCamp.Chapter.Tests/HandTest.cs
```diff
@@ -10,7 +10,18 @@ public class HandTest
     {
         // The constructor cannot be tested because cards is protected. 
 
-        // AddCards cannot be tested because the return type is void and Cards cannot be read outside the class. 
+        [Fact]
+        public void AddCards_Given_Empty_List_And_A_List_With_Elements_Returns_A_List_With_Elements()
```
Almantask, line 14: *Given* is for preconditions
*When* is for args

In your case:
`AddCards_Given_NoCards_When_SingleCard_Returns_Deck_With_SingleCard`

Almantask: Read the test to your wife. If she cannot understand it- the test is not being named well.

Almantask (approved):
It's tested, but your naming could be improved. 

Note: I am very passionate for testing and I am giving top standards for people doing this homework. 
It's not hard to name well, but it needs to be a habit. Even seniors tend to name tests worse than you name them now!

PR #383, opened 2020-05-15
Chapter3/homework/1

DavidKatona:
I've refactored most of the code so it uses properties and lists instead of standard get and set methods & arrays.

I've also made a slight change to your ShopTests class and how it checks items. It didn't make sense to have a List<Item> exposed as a property in the Shop class when I had composed it of an Inventory object. It's much more logical to reuse the inventory class for players, shops, and all entities that might use a similar inventory system and will lead to less code in the long run.

Originally, the tests required me to have a list of items exposed in the Shop and they tested whether items are added/removed correctly from that collection. So I've refactored the ShopTests class so it checks the Inventory component inside the Shop class to do these validations.

Hope that's fine! :)

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -1,44 +1,29 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Collections.Generic;
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
         private const int defaultInventorySize = 10;
 
-        private Item[] _items;
-
-        public Item[] GetItems()
-        {
-            return _items;
-        }
-
-        private int _size;
-
-        public int GetSize()
-        {
-            return _size;
-        }
-
-        private int _remainingSpace;
-
-        public int GetRemainingSpace()
-        {
-            return _remainingSpace;
-        }
+        public List<Item> Items { get; private set; }
+        public int Size { get; private set; }
+        public int RemainingSpace { get; set; }
 
         public Inventory()
         {
-            _size = defaultInventorySize;
-            _remainingSpace = defaultInventorySize;
-            _items = new Item[defaultInventorySize];
+            Size = defaultInventorySize;
+            RemainingSpace = defaultInventorySize;
+            Items = new List<Item>(defaultInventorySize);
         }
 
         public Item[] GetItem(string name)
         {
-            if (GetRemainingSpace() == GetSize()) return new Item[0];
+            if (RemainingSpace == Size) return new Item[0];
 
-            foreach (var item in _items)
+            foreach (var item in Items)
             {
-                if (item.GetName() == name)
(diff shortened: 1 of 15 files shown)


Tests/BootCamp.Chapter.Tests/ShopTests.cs
```diff
@@ -24,7 +24,7 @@ public void Add_New_Item_Adds_Item_To_Stock()
             var sword = new Item(ItemName, ItemPrice, 2);
             _shop.Add(sword);
 
-            _shop.Items.Should().Contain(sword);
+            _shop.Inventory.Items.Should().Contain(sword);
```
DavidKatona, line 27: Some slight changes to your ShopTests class so it verifies whether the inventory component contains the select items.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,28 +1,21 @@
...
     public class Shop
     {
-        private decimal _money;
-        public decimal GetMoney()
-        {
-            return _money;
-        }
-
-        private Inventory _inventory;
+        public Inventory Inventory { get; private set; }
```
DavidKatona, line 7: I had to expose the inventory component through a public getter so it can be used inside ShopTests.cs.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,43 @@
...
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
-        {
-            return new Item[0];
-        }
+        private const int defaultInventorySize = 10;
+
+        public List<Item> Items { get; private set; }
```
Almantask, line 9: If you set something in ctor, it doesn't need a private setter. Can be just a readonly (`get;`) property.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,10 +1,59 @@
...
     {
         static void Main(string[] args)
         {
+            Player playerOne = new Player("Arthas Menethil", 50, 10, new Inventory(), new Equipment());
+            
+            EquipItems_And_DisplayStatistics(playerOne);
+        }
+
+        // This code is ugly but it properly displays the stats of the player after items are equipped.
+        private static void EquipItems_And_DisplayStatistics(Player player)
```
Almantask, line 16: No underscores in names

Almantask (approved):
Looks good! 👍

Src/BootCamp.Chapter/Shop.cs
Almantask: That's alright. It's a problem of a test, but it's okay.

Src/BootCamp.Chapter/Inventory.cs
DavidKatona: Thank you! I'll keep this in mind!

PR #384, opened 2020-05-17
Max Chapter3/homework/10

HellHunterMax:
i took the logic for creating the table from chapter2/Homework1.
At first i had a hard time with it. But the code you gave from examples lesson helped me.

### DIFF
--- a/Src/BootCamp.Chapter/TextBoxPrinter.cs
+++ b/Src/BootCamp.Chapter/TextBoxPrinter.cs
@@ -1,26 +1,10 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Text;
 
 namespace BootCamp.Chapter
 {
-    /*
-     * public class ImportantItem
-    {
-        public ImportantItem()
-        {
-            var print = AttributesGetter.GetClassAttribute<PrintAttribute>(GetType());
-            if (print != null)
-            {
-                Console.WriteLine(print.Message);
-            }
-        }
-    }
-    */
-
     public static class TextBoxPrinter
     {
-        // TODO: Implement.
         public static string Print(object obj)
         {
             var atri = AttributesGetter.GetClassAttribute<TextableAttribute>(obj.GetType());
@@ -33,10 +17,9 @@ namespace BootCamp.Chapter
                 return obj.ToString();
             }
         }
+
         public static string MakePadding(string message, int padding = 0, char horizontal = '-', char vertical = '|', char corner = '+')
         {
-            //[TextTable(padding, sideTop, sideLeft, corner]
-
             string[] splitmessage = message.Split(Environment.NewLine);
             int height = splitmessage.Length + (padding * 2) + 1;
             int width = 0;
@@ -49,6 +32,7 @@ namespace BootCamp.Chapter
                 }
             }
             width += (padding * 2) + 1;
+
             return MakeTabel();
 
             string MakeTabel()
@@ -58,7 +42,6 @@ namespace BootCamp.Chapter
                 {
                     for (int kWidth = 0; kWidth <= width; kWidth++)
                     {
-
                         if (IsThisCorner(iHeight, kWidth))
                         {
                             final.Append(corner);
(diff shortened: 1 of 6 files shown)


Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,137 @@
...
+            {
+                return MakePadding(obj.ToString(), atri.Padding, atri.Horizontal, atri.Vertical, atri.Corner);
+            }
+            else
+            {
+                return obj.ToString();
+            }
+        }
+
+        public static string MakePadding(string message, int padding = 0, char horizontal = '-', char vertical = '|', char corner = '+')
```
Almantask, line 21: Make is a bit of an unconventional name. BuildPadding is a bit better.
I think an even better name would be to just ``Print``. The context (``TextBoxPrinter``) implies on the padding.

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,137 @@
...
+        {
+            string[] splitmessage = message.Split(Environment.NewLine);
+            int height = splitmessage.Length + (padding * 2) + 1;
+            int width = 0;
+
+            foreach (string line in splitmessage) // checking which word is the longest to know how wide the padding should be.
```
Almantask, line 27: Do not inline the comments, add them in a separate line.

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,137 @@
...
+                if (line.Length > width)
+                {
+                    width = line.Length;
+                }
+            }
+            width += (padding * 2) + 1;
+
+            return MakeTabel();
+
+            string MakeTabel()
```
Almantask, line 38: Have a nested method like this beats the purpose: then purpose of nested methods is either reusability or clarification:
1) You do not reuse anything
2) the method is still inside the function, so you don't really hide a step

What you should be doing instead, is- use a private method in a class scope and move it below all the public methods, so that the method becomes an optional detail that you could refer to when reading the method, but are not enforced to.

Src/BootCamp.Chapter/TextBoxPrinter.cs
Almantask, line 27: Instead of writing a comment, just split this to a private function ``FindLongestLineLength``

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,137 @@
...
+                StringBuilder final = new StringBuilder();
+                for (int iHeight = 0; iHeight <= height; iHeight++)
+                {
+                    for (int kWidth = 0; kWidth <= width; kWidth++)
+                    {
+                        if (IsThisCorner(iHeight, kWidth))
+                        {
+                            final.Append(corner);
+                        }
+                        else if (IsThisHorizontal(iHeight, kWidth))
```
Almantask, line 49: ``This`` is redundant

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,137 @@
...
+                int textRow = iHeight - heightOfFirstTextLine;
+
+                if ((iHeight >= heightOfFirstTextLine && iHeight < heightOfFirstTextLine + numberOfTextLines) &&
+                    (kWidth >= heightOfFirstTextLine && kWidth < heightOfFirstTextLine + splitmessage[textRow].Length))
+                {
+                    return true;
+                }
+                return false;
+            }
+            string WhatLetterToBePlaced(int iHeight, int kWidth)
```
Almantask, line 127: instead of **WhatLetterToBePlaced** write **FindLetterToBePlaced**

Almantask (changes requested):

Almantask (approved):

Almantask:
It's okayish, you overdid it with nested classes and in some big function splitting a block of code with a comment into a function is a better thing to do.

Good enough to pass 😄

PR #386, opened 2020-05-17
Chapter 1 : Homework 1And2

Sqelson:
I believe I have completed the criteria, sentence is output to console, as is BMI and it should repeat another 2 times due to implementation of a FOR loop. (comments included too)


Almantask (approved):
Looks like a great first try. Everything is spot on!

PR #388, opened 2020-05-18
Chapter3/homework/7And8

jeanroyce:
TBH my demo are very simple (and ugly) but I think the real job about LINQ is in HW9.


Almantask (approved):
All good

PR #389, opened 2020-05-18
Chapter3/homework/9

jeanroyce:
This HW was really challenger for me. Tasks and extra done.
As I said in previous PR, read `.csv` is painful, too much rules 😅, I would prefer to use a library for this job. I still have bad creativity and bad grammar for name stuffs, Ill work to improve this problem.


Src/BootCamp.Chapter/TransactionProcessors/FullProcessor.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        
+        private static void GetShopSummary(List<Transaction> shop)
+        {
+            var storeName = shop.FirstOrDefault()?.Shop;
+            var summary = new StringBuilder();
+            summary.AppendLine("City, Street, Item, DateTime, Price");
+            
+            foreach (var transaction in shop)
```
Almantask, line 29: If logic can be in lower levels, it should. What is the lowest level in this case? It's the transaction. Let transaction print its' own info.

Then higher-level code will have less burden of low-level concerns.

Almantask (approved):
It's good. I read this a few times but cannot find what to pick on.

Almantask (approved):
It's good. I read this a few times but cannot find what to pick on.

PR #391, opened 2020-05-19
Homework 1and2

hditano:
Added a few inputs
Added a BMI Calculation

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,7 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,12 +6,23 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
-            Console.WriteLine("Hello World!");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,24 @@ class Program
...
+            Console.WriteLine("Type your age: ");
+            int age = Convert.ToInt32(Console.ReadLine());
+            Console.WriteLine("Type your Weight");
+            float weight = float.Parse(Console.ReadLine()); 
+            Console.WriteLine("Type your Height");
+            double height = float.Parse(Console.ReadLine());
+
+            Console.WriteLine(firstName +" " + lastName +" is " + age +" years old, his weight is " + weight + " and his height is " + height);
+
+            double bmi = Convert.ToDouble(weight / (height * height)) * 1000;
```
Almantask, line 22: You don't need to convert this to double, float is okay. If you wanted double, you could have converted the variables above to double instaed.

Almantask (approved):
Overall, it's a good first try :)

In the future, try to be more conscious on the requirements and don't do unnecessary things (like double conversion).
Also, try putting a blank line between unrelated blocks of code, to indicate a separation.

Src/BootCamp.Chapter/Program.cs
Almantask: > Overall, it's a good first try :)
> 
> In the future, try to be more conscious on the requirements and don't do unnecessary things (like double conversion).
> Also, try putting a blank line between unrelated blocks of code, to indicate a separation.

A blank line helps, because code becomes more readable, less of a blob.

PR #392, opened 2020-05-19
Chapter3/Homework/10

jeanroyce:
Reflection is something strange but homework done, example from lesson helped a lot.

### DIFF
--- a/Src/BootCamp.Chapter/TextBoxPrinter.cs
+++ b/Src/BootCamp.Chapter/TextBoxPrinter.cs
@@ -1,12 +1,45 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿using System.Text;
 
 namespace BootCamp.Chapter
 {
     public static class TextBoxPrinter
     {
-        // TODO: Implement.
-        public static string Print(object obj) => "";
+        public static string Print(object obj)
+        {
+            var textTableAttribute = AttributesGetter.GetClassAttribute<TextTableAttribute>(obj.GetType());
+            if (textTableAttribute == null) return obj.ToString();
+            
+            return BuildTable(textTableAttribute, obj.ToString());
+        }
+
+        private static string BuildTable(TextTableAttribute textTable, string message)
+        {
+            var table = new StringBuilder();
+            var paddingSpaceLength = textTable.Padding * 2;
+            var maxTableLines = 3 + paddingSpaceLength;
+            
+            var topAndBottom = textTable.Corner +
+                               new string(textTable.SideTop, message.Length + paddingSpaceLength) + 
+                               textTable.Corner;
+            var mainLine = textTable.SideLeft + new string(' ', textTable.Padding) +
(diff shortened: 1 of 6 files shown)


Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,45 @@
...
-        public static string Print(object obj) => "";
+        public static string Print(object obj)
+        {
+            var textTableAttribute = AttributesGetter.GetClassAttribute<TextTableAttribute>(obj.GetType());
+            if (textTableAttribute == null) return obj.ToString();
+            
+            return BuildTable(textTableAttribute, obj.ToString());
+        }
+
+        private static string BuildTable(TextTableAttribute textTable, string message)
```
jeanroyce, line 15: This function could be done better, but I think Ill only over complicate, I make it simple

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,45 @@
...
+            
+            var topAndBottom = textTable.Corner +
+                               new string(textTable.SideTop, message.Length + paddingSpaceLength) + 
+                               textTable.Corner;
+            var mainLine = textTable.SideLeft + new string(' ', textTable.Padding) +
+                           message.PadLeft(textTable.Padding).PadRight(textTable.Padding) +
+                           new string(' ', textTable.Padding) + textTable.SideLeft;
+            var paddingLine = textTable.SideLeft +
+                               new string(' ', message.Length + paddingSpaceLength) + 
+                               textTable.SideLeft;
```
jeanroyce, line 29: Like this part

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,45 @@
...
+                if (i == 0 || i == maxTableLines - 1)
+                {
+                    table.AppendLine(topAndBottom);
+                    continue;
+                }
+
+                table.AppendLine(i == maxTableLines / 2 ? mainLine : paddingLine);
+            }
+            
+            return table.ToString().TrimEnd();
```
jeanroyce, line 42: I think tests should have Environment.NewLine in end of expected text because of Stringbuilder.ToString behavior, but TrimEnd() solved my problem

Src/BootCamp.Chapter/TextBoxPrinter.cs
Almantask: It really is fine.

Src/BootCamp.Chapter/TextBoxPrinter.cs
Almantask: Maybe. It's not a must for tests to check that I understand a case where tests might not want it. So it's all good and nice that you solved it.

Almantask (approved):
Great

PR #396, opened 2020-05-19
Chapter2/Homework/2

matthew-hole:
First commit. 

PR contains the solution to meet the listed requirements (Unit Tests), but does not include the 'Extra Challenge: Equipment' functionality which will be implemented at a later date.

Only the ChestPiece class was refactored into an 'Item' child class to demonstrate the functionality. A test was performed in the 'Program.cs' file where the functionality is tested.

In the next commit, the Equipment Functionality will be implemented alongside a 'Player/Shop Simulator' which will test all core functionality with edge cases and robustness testing.

### DIFF
--- a/Src/BootCamp.Chapter/Inventory.cs
+++ b/Src/BootCamp.Chapter/Inventory.cs
@@ -1,26 +1,62 @@
-﻿namespace BootCamp.Chapter
+﻿using System.Runtime.ExceptionServices;
+using System;
+using System.Reflection.Metadata.Ecma335;
+
+namespace BootCamp.Chapter
 {
     public class Inventory
     {
-        private Item[] _items;
+        protected Item[] _items;
+
+        // displays all items in inventory
+        // renamed from 'GetItems' to show the clear different from the 'GetItems' method below
+        public void ShowAllItems()
+        {
+            if (!Array.TrueForAll(_items, IsNull => IsNull == null))
+            {
+                Console.WriteLine("Items in inventory:\n");
+                foreach (var item in _items)
+                {
+                    if (item != null)
+                    {
+                        Console.WriteLine(item.GetName());
+                    }
+                    else
+                    {
+                        continue;
+                    }
+                }
+            }
+            else
+            {
+                Console.WriteLine("The inventory is currently empty.");
+            }
+        }
+
+        public Item GetItem(string name)
+        {
+            return _items[Array.FindIndex(_items, element => element.GetName().Equals(name))];         
+        }
+
         public Item[] GetItems()
         {
-            return new Item[0];
+            return _items;
         }
 
-        public Inventory()
-        {
-            _items = new Item[0];
-        }
 
-        public Item[] GetItems(string name)
-        {
-            return new Item[0];
-        }
-
+        // adding an item to the array at the next available slot
         public void AddItem(Item item)
         {
-
+            for(int i = 0; i < _items.Length; i++)
+            {
+                if(_items[i] == null)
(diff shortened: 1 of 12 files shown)


Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,62 @@
...
 {
     public class Inventory
     {
-        private Item[] _items;
-        public Item[] GetItems()
+        protected Item[] _items;
+
+        // displays all items in inventory
+        // renamed from 'GetItems' to show the clear different from the 'GetItems' method below
+        public void ShowAllItems()
```
Almantask, line 13: It doesn't show all items, it displays them; try to be more conventional.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,62 @@
...
         {
-            return new Item[0];
+            if (!Array.TrueForAll(_items, IsNull => IsNull == null))
```
Almantask, line 15: What if array is empty?

Almantask: Try to fail early and avoid negations.
If there are no elements- end the function.
if(Array.TrueForAll(_items, IsNull => IsNull == null))
{
  Console.WriteLine("The inventory is currently empty.");
  return;
}
// no need else. Reduced nesting.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,62 @@
...
+            {
+                Console.WriteLine("Items in inventory:\n");
+                foreach (var item in _items)
+                {
+                    if (item != null)
```
Almantask, line 20: Invert the if.
```
if(item == null) continue;
Console.WriteLine(item.GetName())
```

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,62 @@
...
+            {
+                Console.WriteLine("The inventory is currently empty.");
+            }
         }
 
-        public Inventory()
+        public Item GetItem(string name)
         {
-            _items = new Item[0];
+            return _items[Array.FindIndex(_items, element => element.GetName().Equals(name))];
```
Almantask, line 38: It does work, but in general you should treat C# as raw, without all the neat language features.

Src/BootCamp.Chapter/Player.cs
```diff
@@ -28,34 +34,55 @@ public class Player
         /// <summary>
         /// Player items. There can be multiple of items with same name.
         /// </summary>
-        private Inventory _inventory;
+        private PlayerInventory _inventory;
```
Almantask, line 37: Why have a distinction between just inventory? Why is player inventory is any special? Keep it simple, don't be redundant.

Src/BootCamp.Chapter/Player.cs
```diff
@@ -28,34 +34,55 @@ public class Player
...
+          
         }
 
         /// <summary>
         /// Adds item to player's inventory
         /// </summary>
+        // The inventory method 'AddItem' will throw an expection if no room is available in the inventory for an another item
         public void AddItem(Item item)
         {
+            try
```
Almantask, line 73: It's not that excpetional that inventory is full. Instead, check the inventory size before adding more items to it. throwing an exception is costly and it should also be reserved for unexpected situations.

Almantask (approved):
Overall, looks good.

PR #401, opened 2020-05-21
Chapter 1 / homework 1 & 2

tuta-bits:
Made new changes as requested

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -9,34 +9,41 @@ namespace BootCamp.Chapter
             Console.WriteLine("Hello World!");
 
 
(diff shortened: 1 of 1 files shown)


Almantask (changes requested):
Please redo the homework so that input is actually read from the console. Thanks and welcome!

tuta-bits (commented):
Second Attempt on Chapter 1/Homework 1&2

Almantask (approved):
Looks good. You overdid this a bit, be careful not to do too many things in the next homework because tests will be failing.
Also mind how you format if statements. A readable way is:
```
if(condition)
{
  //
}
else
{
}
```
We don't put brackets at the same line in C#.

PR #407, opened 2020-05-21
Max Chapter4/homework/1

HellHunterMax:
Should i just have tested DrawAt(Index) because everything calls that so if that works the rest is kind of sure to work?
i did not test DrawRandom() because i  don't think its worth the time to test the method. to complex to test randomness ? and it will return a card because it uses DrawAt().

### DIFF
--- a/Src/BootCamp.Chapter/Gambling/GameComponents/Deck.cs
+++ b/Src/BootCamp.Chapter/Gambling/GameComponents/Deck.cs
@@ -39,9 +39,14 @@ namespace BootCamp.Chapter.Gambling
 
         private void ValidateDeckSize(int index)
         {
-            if (index >= _cards.Count)
+            
+            if (_cards.Count == 0)
             {
-                throw new OutOfCardsException(index, _cards.Count);
+                throw new OutOfCardsException();
+            }
+            if (index < 0 || index >= _cards.Count)
+            {
+                throw new IndexOutOfRangeException($"Cant pick a card at index {index} becouse deck size is {_cards.Count} and cant be less than 0");
             }
         }
     }
--- a/Tests/BootCamp.Chapter.Tests/DeckTest.cs
+++ b/Tests/BootCamp.Chapter.Tests/DeckTest.cs
@@ -36,8 +36,7 @@ namespace BootCamp.Chapter.Tests
         public void DrawAt_OutOfIndex_Throws_IndexOutOfRangeException(int index)
         {
             //Arrange
-            List<Card> cards = BuildFullListOfCards();
-            IDeck deck = new Gambling.Deck(cards);
+            IDeck deck = BuildFullDeckOfCards();
 
             //Act
             Action action = () => deck.DrawAt(index);
@@ -71,9 +70,8 @@ namespace BootCamp.Chapter.Tests
             //Act
             Card pickedCard = deck.DrawAt(1);
 
-            //Assert wich is better and why?
+            //Assert
             Assert.Equal(cards[1], pickedCard);
-            //pickedCard.Should().Be(cards[index]);
         }
 
         [Fact]
@@ -140,5 +138,11 @@ namespace BootCamp.Chapter.Tests
             return cardsDeck;
         }
 
+        private static IDeck BuildFullDeckOfCards()
+        {
+            List<Card> cards = BuildFullListOfCards();
+            IDeck deck = new Gambling.Deck(cards);
+            return deck;
+        }
     }
 }
--- a/Tests/BootCamp.Chapter.Tests/DeckTest.cs
+++ b/Tests/BootCamp.Chapter.Tests/DeckTest.cs
@@ -21,7 +21,7 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void Draw_When_DeckHasNoCards_Throws_OutOfCardsException()
+        public void Draw_Given_DeckDeckWithNoCards_Throws_OutOfCardsException()
         {
             IDeck deck = new Gambling.Deck(new List<Card>());
 
--- a/Tests/BootCamp.Chapter.Tests/DeckTest.cs
+++ b/Tests/BootCamp.Chapter.Tests/DeckTest.cs
@@ -59,7 +59,7 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void DrawRandom_DeckWith1Card_returns_ExcpectedCard()
+        public void DrawRandom_Given_DeckWith1Card_returns_ExcpectedCard()
         {
 
             //Arrange
@@ -73,7 +73,7 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void DrawRandom_DeckWithNoCards_Throws_OutOfCardsException()
+        public void DrawRandom_Given_DeckWithNoCards_Throws_OutOfCardsException()
         {
             IDeck deck = new Gambling.Deck(new List<Card>());
 
@@ -106,7 +106,7 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void DrawFromTop_Given_DeckWith1ExtraCard_Returns_ExpectedCard()
+        public void DrawFromTop_Given_DeckWith1Card_Returns_ExpectedCard()
         {
             //Arrange
             BuildDeckWith1Card(out Card lastCard, out IDeck deck);
@@ -119,7 +119,7 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void Shuffle_Should_Return_Different_Order()
+        public void Shuffle_Returns_DifferentCardOrder()
         {
             //Arrange
             BuildListOfCardsGiveItToDeckAndShuffle(out List<Card> cards, out IDeck deck, out List<Card> shuffledDeck);
--- a/Tests/BootCamp.Chapter.Tests/DeckTest.cs
+++ b/Tests/BootCamp.Chapter.Tests/DeckTest.cs
@@ -110,10 +110,7 @@ namespace BootCamp.Chapter.Tests
         public void Shuffle_Should_Return_Different_Order()
         {
             //Arrange
-            List<Card> cards = BuildFullListOfCards();
-            IDeck deck = new Gambling.Deck(BuildFullListOfCards());
-            List<Card> shuffledDeck = new List<Card>();
-            cards.Reverse();
+            BuildDeckAndListOfCards(out List<Card> cards, out IDeck deck, out List<Card> shuffledDeck);
 
             //Act
             deck.Shuffle();
@@ -125,6 +122,15 @@ namespace BootCamp.Chapter.Tests
             //Assert
             shuffledDeck.Should().Contain(cards).And.NotEqual(cards);
         }
+
+        private static void BuildDeckAndListOfCards(out List<Card> cards, out IDeck deck, out List<Card> shuffledDeck)
+        {
+            cards = BuildFullListOfCards();
+            deck = new Gambling.Deck(BuildFullListOfCards());
+            shuffledDeck = new List<Card>();
+            cards.Reverse();
+        }
+
         private static List<Card> BuildFullListOfCards()
         {
             List<Card> cardsDeck = new List<Card>();
--- a/Tests/BootCamp.Chapter.Tests/DeckTest.cs
+++ b/Tests/BootCamp.Chapter.Tests/DeckTest.cs
@@ -58,19 +58,35 @@ namespace BootCamp.Chapter.Tests
             newCards.Should().Contain(cards).And.NotEqual(cards);
         }
 
-        private static void BuildListOfCardsGiveItToDeckAndCreateEmptyListOfCards(out List<Card> cards, out IDeck deck, out List<Card> newCards)
+        [Fact]
+        public void DrawRandom_DeckWith1Card_returns_ExcpectedCard()
         {
-            cards = BuildFullListOfCards();
-            deck = new Gambling.Deck(cards);
-            newCards = new List<Card>();
+
+            //Arrange
+            BuildDeckWith1Card(out Card lastCard, out IDeck deck);
+
+            //Act
+            Card cardDrawn = deck.DrawRandom();
+
+            //Assert
+            cardDrawn.Should().Be(lastCard);
+        }
+
+        [Fact]
+        public void DrawRandom_DeckWithNoCards_Throws_OutOfCardsException()
+        {
+            IDeck deck = new Gambling.Deck(new List<Card>());
+
+            Action action = () => deck.DrawRandom();
+
+            action.Should().Throw<OutOfCardsException>();
         }
 
         [Fact]
         public void DrawAt_When_IndexWithinRange_Returns_ExpectedCard()
         {
             //Arrange
-            List<Card> cards = BuildFullListOfCards();
-            IDeck deck = new Gambling.Deck(cards);
+            BuildDeckAndGiveItFullListOfCards(out List<Card> cards, out IDeck deck);
 
             //Act
             Card pickedCard = deck.DrawAt(1);
@@ -80,16 +96,10 @@ namespace BootCamp.Chapter.Tests
         }
 
         [Fact]
-        public void DrawFromTop_RemovesACard_And_DrawAt0_Throws_OutOfCardsException()
+        public void DrawAt0_Given_EmptyDeck_Throws_OutOfCardsException()
         {
-            //Test if DrawFromTop actually removes a card from deck.
+            IDeck deck = new Deck(new List<Card>());
 
-            Card lastCard = new Card(Card.Suites.Hearts, Card.Ranks.Ace);
-            List<Card> cards = new List<Card>();
-            cards.Add(lastCard);
-            IDeck deck = new Gambling.Deck(cards);
-
-            deck.DrawFromTop();
             Action action = () => deck.DrawAt(0);
 
             action.Should().Throw<OutOfCardsException>();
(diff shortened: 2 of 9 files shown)


HellHunterMax:
oops forgot to test HAND class.

HellHunterMax:
still created a DrawRandom() test. in theory it could fail because it pulled them random in the same order as they were put in, but the chance of that happening is 0,0000000000001 ? 😛

Src/BootCamp.Chapter/Gambling/GameComponents/Deck.cs
```diff
@@ -39,9 +39,14 @@ public Card DrawAt(int index)
 
         private void ValidateDeckSize(int index)
         {
-            if (index >= _cards.Count)
+            
+            if (_cards.Count == 0)
```
Almantask, line 43: This one is not needed, only the condition below. If you have 0 cards, then index will be equal, which will throw.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class TestsDeck
+    {
+        private static List<Card> FullDeckOfCards()
```
Almantask: name it as a verb.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+using FluentAssertions;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class TestsDeck
```
Almantask: should be ``DeckTest``

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+                foreach (Card.Ranks rank in (Card.Ranks[])Enum.GetValues(typeof(Card.Ranks)))
+                {
+                    cardsDeck.Add(new Card(suite, rank));
+                }
+            }
+            return cardsDeck;
+        }
+
+        [Fact]
+        public void When_No_Cards_Given_To_Deck_Should_Throw_ArgumentNullException()
```
Almantask: All tests should start with the action name, followed by preconditions and parameters, followed by expectation.
``Act_Given_Precondition_When_Argument_Expectation``
In your case, it should be: 
`New_When_CardsNull_Throws_ArgumentException`

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            List<Card> cards = null;
+
+            Action action = () => new Gambling.Deck(cards);
+
+            action.Should().Throw<ArgumentNullException>();
+        }
+
+        [Fact]
+        public void Draw_When_Deck_Has_No_Cards_Should_Throw_OutOfCardsException()
```
Almantask: When -> Given

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            Action action = () => deck.DrawFromTop();
+
+            action.Should().Throw<OutOfCardsException>();
+        }
+
+        [Theory]
+        [InlineData(-1)]
+        [InlineData(100)]
+        public void DrawAt_Out_Of_Index_Should_Throw_Out_Of_Range(int index)
```
Almantask: DrawAt_When_IndexOutOfBounds_Throws_IndexOutOfRangeException()

Almantask: Also no need to put a `_` after every word, just after every part of the 3A

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            //Act
+            Action action = () => deck.DrawAt(index);
+
+            //Assert
+            action.Should().Throw<IndexOutOfRangeException>();
+        }
+
+        [Fact]
+        public void DrawRandom_52_Times_Test()
```
Almantask: eww 😄! DrawRandom_52Times_DoesNotThrow()

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+            for (int i = 0; i < 52; i++)
+            {
+                newCards.Add(deck.DrawRandom());
+            }
+
+            newCards.Should().Contain(cards).And.NotEqual(cards);
+        }
+
+        [Fact]
+        public void DrawAt_Index_Within_Range_should_Give_Correct_Card()
```
Almantask: DrawAt_When_IndexWithinRange_Returns_ExpectedCard

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+            //Assert
+            cardDrawn.Should().Be(lastCard);
+        }
+
+        [Fact]
+        public void Shuffle_Should_Return_Different_Order()
+        {
+            //Arrange
+            List<Card> cards = FullDeckOfCards();
+            IDeck deck = new Gambling.Deck(FullDeckOfCards());
```
Almantask: Instead of just  creating a list of card, create a deck os that the arrange step will be minimalistic and we won't have to look at al the setup that we should not care about.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            //Act
+            deck.Shuffle();
```
Almantask: too much is going on. Act should ever be a single line of code.

Almantask: AddCards, reverse, new deck, like wtf 🗡️

Tests/BootCamp.Chapter.Tests/TestsHand.cs
```diff
@@ -0,0 +1,34 @@
+﻿using BootCamp.Chapter.Gambling;
+using System.Collections.Generic;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class TestsHand : Gambling.Hand
+    {
+        //PS cant remove cards from hand? weird....
```
Almantask: The intention was to have a Poker hand. Poker hand does not need removing cards. You just add cards one by one, at the end of the game reveal them, and then start over with new cards.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+            IDeck deck = new Gambling.Deck(cards);
+
+            deck.DrawFromTop();
+            Action action = () => deck.DrawAt(0);
+
+            action.Should().Throw<OutOfCardsException>();
+        }
+
+        [Fact]
+        public void DrawFromTop_Should_Give_Last_Card_In_List()
```
Almantask: I think should is not needed, because you always will have to add it, making it extremely redundant. There is always something that should happen.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            //Arrange
+            Card lastCard = new Card(Card.Suites.Hearts, Card.Ranks.Ace);
```
Almantask: Try to create a bunch of creation methods, so that you can pass whatever test needs (like rank and suite of a card for adding the last card). Like this, you would have a single line of code for setting up your test case, which makes it so much more self documenting.

Almantask (changes requested):
Try to fix the test names based on the remarks.

Src/BootCamp.Chapter/Gambling/GameComponents/Deck.cs
HellHunterMax: i wanted a different Error message. one is for OutOfCardsException(); when there are no cards in the deck. the other IndexOutOfRangeException().

HellHunterMax:
Assert.Equal(cards[1], pickedCard);
pickedCard.Should().Be(cards[index]);

Which one is better and why ?

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
HellHunterMax: much more readable.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
HellHunterMax: in most cases both the deck and Cards list are used so there it would not look cleaner and just be more work ? In one case I did change it and it did look cleaner.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
HellHunterMax: changed it a bit I did want to make this test to see if the shuffle works.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
HellHunterMax: Testing if it returns the last card and not any other card is not a valid test? it could return the first card then it would fail.

Almantask:
> Should i just have tested DrawAt(Index) because everything calls that so if that works the rest is kind of sure to work?
> i did not test DrawRandom() because i don't think its worth the time to test the method. to complex to test randomness ? and it will return a card because it uses DrawAt().

It's worth testing, but under condition that you control. What can you control? Size of a deck.
Draw random - no cards- exception.
Draw random- 1 card- draws that card. That should be your test cases.
You can still make bugs on how you generate the random number (the ranges), the range depends on size of a deck therefore it should be tested 😄

Almantask:
> Assert.Equal(cards[1], pickedCard);
> pickedCard.Should().Be(cards[index]);
> 
> Which one is better and why ?

It's a matter of style. However, I would prefer the second one, because if it fails, it gives more detail where it failed and it is more readable.

Almantask:
> still created a DrawRandom() test. in theory it could fail because it pulled them random in the same order as they were put in, but the chance of that happening is 0,0000000000001 ? 😛

I might sound crazy, but it's too risky 🔪 
Imagine this was for testing rocket lunch... Is 0,0000000000001 of failure good enough? Especially when it's known.

Tests/BootCamp.Chapter.Tests/TestsDeck.cs
Almantask: I completely failed to express myself here.
The test is needed. I meant the name.
All tests will always end with should, if you add it. It's redundant.
Word `Should` can be removed.

Tests/BootCamp.Chapter.Tests/HandTest.cs
```diff
@@ -0,0 +1,34 @@
...
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class HandTest : Gambling.Hand
+    {
+        //PS cant remove cards from hand? weird....
+
+        [Fact]
+        public void Reveal_Returns_Hand_Given()
```
Almantask: your assert talks about cards and not hand. Reveal_Returns_CardsInHand()

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            //Assert
+            cardDrawn.Should().Be(lastCard);
+        }
+
+        [Fact]
+        public void Shuffle_Should_Return_Different_Order()
+        {
+            //Arrange
+            List<Card> cards = BuildFullListOfCards();
```
Almantask: Still too much noise. The whole setup is irrelevant so hide it under a single method.

Almantask (changes requested):
Try to hide test details under private methods in those cases, where those details are not mentioned in a test name.

HellHunterMax:
> > still created a DrawRandom() test. in theory it could fail because it pulled them random in the same order as they were put in, but the chance of that happening is 0,0000000000001 ? 😛
> 
> I might sound crazy, but it's too risky 🔪
> Imagine this was for testing rocket lunch... Is 0,0000000000001 of failure good enough? Especially when it's known.

i think you sound crazy :P there is way more chance of a rocket blowing up i think. (Falcon 9's 95% success rate) the chance of cards coming out the same are :  1 in 10 to the power 68 (or 1 followed by 68 zeros). 0,000000000000000000000000000000000000000000000000000000000000000000001
i think i would take that chance, if its shuffled correctly.
but this is just a test to see if the method is working as intended.

HellHunterMax:
I don't think this is clean?
and there are too many single fact tests?

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            //Act
+            Card pickedCard = deck.DrawAt(1);
+
+            //Assert
+            Assert.Equal(cards[1], pickedCard);
+        }
+
+        [Fact]
+        public void DrawAt0_Given_EmptyDeck_Throws_OutOfCardsException()
```
jdphenix, line 99: `DrawAt_When_Empty_Deck_Given_0_Throws_OutOfCardsException`

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            //Act
+            Card cardDrawn = deck.DrawRandom();
+
+            //Assert
+            cardDrawn.Should().Be(lastCard);
+        }
+
+        [Fact]
+        public void DrawRandom_Given_DeckWithNoCards_Throws_OutOfCardsException()
```
jdphenix, line 76: `When` not `Given`

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,187 @@
...
+            for (int i = 0; i < 52; i++)
+            {
+                newCards.Add(deck.DrawRandom());
+            }
+
+            newCards.Should().Contain(cards).And.NotEqual(cards);
+        }
+
+        [Fact]
+        public void DrawRandom_Given_DeckWith1Card_returns_ExcpectedCard()
```
jdphenix, line 62: `When_Has_One_Card`

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class DeckTest
+    {
+        [Fact]
+        public void Deck_Created_WithoutCards_Throws_ArgumentNullException()
```
jdphenix, line 14: Maybe `Deck_Created_With_Null_Cards...` Without cards seems to describe something different than null. 

Probably just semantics

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            Action action = () => deck.DrawFromTop();
+
+            action.Should().Throw<OutOfCardsException>();
+        }
+
+        [Theory]
+        [InlineData(-1)]
+        [InlineData(100)]
+        public void DrawAt_OutOfIndex_Throws_IndexOutOfRangeException(int index)
```
jdphenix, line 36: `DrawAt_Given_Out_Of_Index...`

jdphenix (approved):
I honestly only have nitpicks at this point. I think this looks pretty good.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
```diff
@@ -48,9 +48,7 @@ public void DrawAt_OutOfIndex_Throws_IndexOutOfRangeException(int index)
         [Fact]
         public void DrawRandom_52Times_Returns_SameCards_DifferentOrder()
         {
-            List<Card> cards = BuildFullListOfCards();
-            IDeck deck = new Gambling.Deck(cards);
-            List<Card> newCards = new List<Card>();
+            BuildListOfCardsGiveItToDeckAndCreateEmptyListOfCards(out List<Card>  cards, out IDeck deck, out List<Card> newCards);
```
Almantask, line 51: Too long just use a theory and takes out List<Card>  cards, out IDeck deck, out List<Card> newCards as input in test.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: When is for input;
Given is for preconditions;
(we made this convention quite some time ago)

Almantask: Also another convention we made was:
Not using _ after every word, rather after every difference, for example:
Method_Given_Precondition_When_Input_...
Is precondition or input has a few words, no separation will be made- no point- disrupts readability.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: I disagree. When for input, given for preconditions.

Tests/BootCamp.Chapter.Tests/DeckTest.cs
Almantask: More about when given then:
https://martinfowler.com/bliki/GivenWhenThen.html
The convention comes from BDD.

Almantask (approved):

Almantask (approved):

Almantask:
It's up to you if you want to refactor that one test method, the rest looks good and this is definitely good enough to move  on 😄

PR #409, opened 2020-05-22
Chapter3/homework/2

DavidKatona:
Here's the solution for CH3 homework 2. I've spent a fair amount of time researching and trying to imagine how this could be structured.

Eventually, I took a peak at some of the solutions given by others and tried going on a similar route they did. I think that looking at others' solutions is cheating and I do feel a bit disappointed in myself.


Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,13 +1,20 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Data;
+
+namespace BootCamp.Chapter
 {
     // TODO: make a struct and add validation and other needed methods (if needed)
-    public class Credentials
+    public readonly struct Credentials : IEquatable<Credentials>
```
DavidKatona, line 7: Implementing IEquatable was suggested by MihailFox so no props to me. I did my homework though and researched it fairly well so I can understand and reimplement this again in the future.

Src/BootCamp.Chapter/ConsoleUI/ConsoleNavigator.cs
```diff
@@ -0,0 +1,66 @@
...
+    public class ConsoleNavigator
+    {
+        public List<MenuOptions> MenuOptions { get; }
+
+        public ConsoleNavigator()
+        {
+            MenuOptions = Enum.GetValues(typeof(MenuOptions)).Cast<MenuOptions>().ToList();
+        }
+
+        public int RenderMenu()
```
Almantask, line 18: Too much is going on . please split in step functions.

Src/BootCamp.Chapter/Credentials.cs
Almantask: IEquatable is needed for comparing. There is indeed a use case here. So that is fine.

Almantask (approved):
Looks alright

PR #410, opened 2020-05-23
Chapter1/homework/7

mkasarci:
That was kinda tricky one but I've passed all the tests.

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -107,17 +107,27 @@ namespace BootCamp.Chapter
         /// </summary>
         public static string FindMostPoorPerson(string[] peopleAndBalances)
         {
-            decimal lowerBalance = 0;
-            string mostPoorPerson = "";
+            if (isArrayNullOrEmpty(peopleAndBalances))
+                return "N/A.";
+            var ListOfPeople = GetPeople(peopleAndBalances);
+
+            decimal lowerBalance = GetPeople(peopleAndBalances)[0].Balance[^1];
+            List<string> mostPoorPerson = new List<string>();
 
             foreach (Person person in GetPeople(peopleAndBalances))
             {
                 if (person.Balance[^1] < lowerBalance)
                 {
                     lowerBalance = person.Balance[^1];
+                    mostPoorPerson.Clear();
+                    mostPoorPerson.Add(person.Name);
+                }
+                else if (person.Balance[^1] == lowerBalance)
+                {
+                    mostPoorPerson.Add(person.Name);
                 }
             }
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +13,169 @@ public static class BalanceStats
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            if (isArrayNullOrEmpty(peopleAndBalances))
+                return "N/A.";
```
Almantask, line 16: Can be moved to a constant, because it's repeated in multiple places.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +13,169 @@ public static class BalanceStats
...
+
+            decimal highestBalance = 0;
+            List<string> Name = new List<string>();
```
Almantask: Don't use C# collections yet. Use array.

Almantask (changes requested):
Please redo the homework by not using collection or other advanced language features. That will come in chapter 3

mkasarci:
I've changed all the generic lists to array. It took more time than I expected.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -109,73 +128,111 @@ public static string FindMostPoorPerson(string[] peopleAndBalances)
...
             string formattedName = "";
-            if (Name.Count == 1)
-                return Name[0];
 
-            for (int i = 0; i < Name.Count; i++)
+            for (int i = 0; i < nameArray.Length; i++)
             {
-                formattedName += Name[i];
-                if (Name.Count - 2 == i)
+                formattedName += nameArray[i];
```
Almantask, line 199: Instead of concatinating a string like this, use `StringBuilder`. The performance difference, starting from 5 iterations starts changing in multiple times!

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -109,73 +128,111 @@ public static string FindMostPoorPerson(string[] peopleAndBalances)
...
+                if (nameArray[i + 2] == "" || nameArray[i+2] == null)
```
Almantask, line 200: Consistently format your code

Almantask, line 200: Can be simplified. string.IsNullorEmpty(nameArray[i+2])
Also avoid using magic numbers. It's hard for me to figure out what is +2 for without more context.

Almantask (approved):
Overall looks good

PR #411, opened 2020-05-24
Chapter1/homework/5

Dreniak:


Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
         {
-            // ToDo: implement.
-            return array;
+            int[] newArray = new int[array.Length + 1];
+
+            if (IsNullOrEmpty(array))
+            {
+                newArray[0] = number;
+                return newArray;
+            }
```
jdphenix: `int[] newArray = new int[array.Length + 1]` will throw a null reference exception if `array` is null. `InsertAt` is supposed to return a new array with the single new element.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
         }
 
-        /// <summary>
-        /// Inserts a new array element at the specified index.
-        /// </summary>
-        /// <param name="array">Input array.</param>
-        /// <param name="number">Number to be added.</param>
-        /// <param name="index">Index at which array element should be added.</param>
-        /// <returns>A new array with element inserted at a given index. If an array is empty or null, returns new array with number in it.</returns>
         public static int[] InsertAt(int[] array, int number, int index)
```
jdphenix, line 105: If the index is out-of-bounds to the input array, this method should return the array unmodified.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+
+            newArray[0] = number;
+
+            for (int i = 1; i < newArray.Length; i++)
+            {
+                newArray[i] = array[i - 1];
+            }
+
+            return newArray;
         }
```
jdphenix: Consider implementing this method in terms of `InsertAt` - hint: maybe this method could be replaced completely with `InsertAt(0)`

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+
+            for (int i = 0; i < array.Length; i++)
+            {
+                newArray[i] = array[i];
+            }
+
+            newArray[newArray.Length - 1] = number;
+
+            return newArray;
         }
```
jdphenix: Consider implementing this method in terms of `InsertAt`

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
 
-        /// <summary>
-        /// Removes array element at given index.
-        /// </summary>
-        /// <param name="array">Input array.</param>
-        /// <param name="index">Index at which array element should be removed.</param>
-        /// <returns>A new array with element removed at a given index. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveAt(int[] array, int index)
```
jdphenix, line 63: This method should be returning the original array unmodified if the index is out of range. Consider handling that at the beginning of the method.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+
+            int[] newArray = new int[array.Length - 1];
+
+            for (int i = 0; i < newArray.Length; i++)
+            {
+                newArray[i] = array[i + 1];
+            }
+
+            return newArray;
         }
```
jdphenix: Consider implementing this is terms of `RemoveAt`

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+
+            int[] newArray = new int[array.Length - 1];
+            
+            for (int i = 0; i < array.Length - 1; i++)
+            {
+                newArray[i] = array[i];
+            }
+            
+            return newArray;
         }
```
jdphenix: Consider implementing this is terms of `RemoveAt`

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+        }
+
+        internal static bool Sort()
+        {
+            throw new NotImplementedException();
+        }
+
+        private static bool IsNullOrEmpty(int[] array)
+        {
+            if (array.Length == 0 || array == null)
```
jdphenix: Flip this around. `array.Length` will throw a NullReferenceException. Remember that `||` is short-circuiting.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -1,95 +1,183 @@
...
+                newArray[i] = array[i - 1];
+            }
+
+            return newArray;
+        }
+
+        internal static bool Sort()
+        {
+            throw new NotImplementedException();
+        }
```
jdphenix: Can delete this if it's not used.

jdphenix (changes requested):

Dreniak:
I'm having some trouble performing tests because of my Visual Studio. Hopefully by tomorrow i'll have it sorted and will be able to fix the remaining errors.

jdphenix (changes requested):
Several tests are failing, 

- InsertAt_Returns_Array_WithExpectedElements
- InsertLast_Returns_Array_With_ExpectedElements
- RemoveLast_Returns_Array_With_ExpectedElements

These will need to be be passing.

jdphenix:
> 
> 
> I'm having some trouble performing tests because of my Visual Studio. Hopefully by tomorrow i'll have it sorted and will be able to fix the remaining errors.

Okay, that sounds good. Just confirm when you'd like another review.

Dreniak:
Ok, I sorted things out. Worked on the code for some time today. Fixed the error I noticed that was related to RemoveAt and InsertAt usage of index. I tested every situations I could imagine and still don't know where these errors are coming from. I cannot see the array they used for input to debug so I'm quite lost.

jdphenix:
> 
> 
> Ok, I sorted things out. Worked on the code for some time today. Fixed the error I noticed that was related to RemoveAt and InsertAt usage of index. I tested every situations I could imagine and still don't know where these errors are coming from. I cannot see the array they used for input to debug so I'm quite lost.

I would consider reviewing the [Lesson 6 material](https://discordapp.com/channels/475671343463923714/665673287778697248/672174368973127722). I'm also usually on the Discord as @jdphenix if you'd like to poke me in one of the question channels. 

For your still failing tests, I would look at the following cases: 

- What happens when the array is null? Is there a way you might be able to consistently apply null handling behavior in every method?
- For InsertAt, what happens when you try to insert an element outside of the bounds of the array using your methods?

Dreniak:
Finally finished it. I did a lot of newbie's mistakes on the things i had understood, although I was unable to get how I really was supposed to handle empty arrays in InsertAt method, which I only figured out after checking some other Pull Request. I don't think the instructions were clear in this one.
Glad I finished it, really made me improve as a developer.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -2,94 +2,152 @@
...
-            // ToDo: implement.
-            return array;
+            int[] newArray;
+
+            if (array == null)
+            {
+                newArray = new int[1] { number };
+            }
+
+            else if (index < 0 || index > array.Length) //If index is OutOfBounds
```
Almantask, line 114: instead of a comment, make code speaking for itself:
```
bool outOfBounds = index < 0 || index > array.Length;
..
else if(outOfBounds)
{
  return array;
}
```
no comment is needed!
Try looking at the example I made some time ago regarding code speaking for itself.
Code with many comments:
https://github.com/Realm-of-Everlasting-Code/C-Sharp/blob/master/General/CodeSmells/Comments/Src/Comments.Problem/Algorithm%20Without%20Explanation/Better/Sorting.cs
Code that speaks for itself:
https://github.com/Realm-of-Everlasting-Code/C-Sharp/blob/master/General/CodeSmells/Comments/Src/Comments.Problem/Algorithm%20Without%20Explanation/Good/Sorting.cs

Almantask (approved):
Looks good, just try to instead of writing a comment, doing the same but by expressing yourself in code.

PR #412, opened 2020-05-24
Chapter4/homework/2

ghost:
Second attempt I messed up the first one

### DIFF
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
@@ -44,8 +44,8 @@ namespace BootCamp.Chapter.Gambling.Poker
 
             var winners = playersByScore.Where(p => p.Value == highestScore)
                                         .Select(p => p.Key);
-            
-            return winners.ToList();
+
+            return winners;
         }
 
         private PokerCombo FindCombo(IEnumerable<Card> cards)
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
@@ -39,13 +39,13 @@ namespace BootCamp.Chapter.Gambling.Poker
                 playersWithCombos.Add(player, combo);
             }
 
-            var playersByScore = playersWithCombos.OrderByDescending(kvp => kvp.Key);
+            var playersByScore = playersWithCombos.OrderByDescending(kvp => kvp.Value);
             var highestScore = playersByScore.First().Value;
 
             var winners = playersByScore.Where(p => p.Value == highestScore)
                                         .Select(p => p.Key);
             
-            return winners;
+            return winners.ToList();
         }
 
         private PokerCombo FindCombo(IEnumerable<Card> cards)
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
@@ -39,7 +39,7 @@ namespace BootCamp.Chapter.Gambling.Poker
                 playersWithCombos.Add(player, combo);
             }
 
-            var playersByScore = playersWithCombos.OrderByDescending(kvp => kvp.Value);
+            var playersByScore = playersWithCombos.OrderByDescending(kvp => kvp.Key);
             var highestScore = playersByScore.First().Value;
 
             var winners = playersByScore.Where(p => p.Value == highestScore)
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
@@ -28,7 +28,11 @@ namespace BootCamp.Chapter.Gambling.Poker
 
         public override IEnumerable<Player> GetWinners()
         {
-            if (!Players.Any()) return new Player[0];
+            if (!Players.Any())
+            {
+                throw new InvalidGameStateException("There must at least 1 player"); 
+            }
+
 
             var playersWithCombos = new Dictionary<Player, PokerCombo>();
 
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Text;
 
 namespace BootCamp.Chapter.Gambling.Poker
@@ -9,9 +10,9 @@ namespace BootCamp.Chapter.Gambling.Poker
     /// Other combos have a score, which is based on how high the cards make them.
     /// If combo is not existing, score is 0.
     /// </summary>
-    public class PokerCombo
+    public class PokerCombo : IComparable<PokerCombo>, IComparable
     {
-        public enum Sets
+        public enum Sets 
         {
             HighCard,
             Pair,
@@ -51,12 +52,12 @@ namespace BootCamp.Chapter.Gambling.Poker
 
         public static int CalculateRoyalFlushScore(IEnumerable<Card> cards)
         {
-            return 0;
+            return 12;
         }
 
         public static int CalculateStraightFlushScore(IEnumerable<Card> cards)
         {
-            return 0;
+            return 11;
         }
 
         public static int CalculateFourOfAKindScore(IEnumerable<Card> cards)
@@ -97,6 +98,38 @@ namespace BootCamp.Chapter.Gambling.Poker
         public static int CalculateHighCardScore(IEnumerable<Card> cards)
         {
             return 0;
+        }
+
+        public int CompareTo(PokerCombo other)some
+        {
+            if (this.Score < other.Score)
+            {
+                return -1;
+            }
+            
+            if (this.Score == other.Score)
+            {
+                return 0; 
+            }
+
+            return 1; 
+        }
+
+        public override bool Equals(object obj)
+        {
+            return obj is PokerCombo combo &&
+                   Set == combo.Set &&
+                   Score == combo.Score;
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Set, Score);
+        }
+
+        public int CompareTo(object obj)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs
+++ b/Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
-using System.Text;
+using System.Linq;
 
 namespace BootCamp.Chapter.Gambling.Poker
 {
@@ -12,7 +11,7 @@ namespace BootCamp.Chapter.Gambling.Poker
     /// </summary>
     public class PokerCombo : IComparable<PokerCombo>
     {
-        public enum Sets 
+        public enum Sets
         {
             HighCard,
             Pair,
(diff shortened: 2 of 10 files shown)


ghost:
With the poker test , you said in the first PR that it can be done with 3 sentences. I have to think good how to do that or I have to make the extra challenge also.

jdphenix:
> 
> 
> With the poker test , you said in the first PR that it can be done with 3 sentences. I have to think good how to do that or I have to make the extra challenge also.

One way to consider - set up an expectations (an IEnumerable<object>), use `TheoryAttribute`, and feed the test the players and expected winner. 

Remember you'll need to implement the combo logic in the Src/BootCamp.Chapter project to really test your test.

Tests/BootCamp.Chapter.Tests/PokerHandsTests.cs
```diff
@@ -0,0 +1,37 @@
+﻿using BootCamp.Chapter.Gambling;
+using BootCamp.Chapter.Gambling.Poker;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class PokerHandsTests
```
jdphenix, line 10: Poker hands have a maximum size. Consider testing that.

Tests/BootCamp.Chapter.Tests/PokerTests.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class PokerTests
+    {
+        [Fact]
+        public void GetWinners_Given_List_Players_Returns_Winnner()
```
jdphenix: This test should be a Theory.

Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
```diff
@@ -45,7 +45,7 @@ public override IEnumerable<Player> GetWinners()
             var winners = playersByScore.Where(p => p.Value == highestScore)
                                         .Select(p => p.Key);
             
-            return winners;
+            return winners.ToList();
```
jdphenix: This shouldn't need to be changed.

jdphenix (changes requested):
- I don't see any tests for `Deck`. 
  - You'll need to make a couple of changes to `Deck` to fully test it. You need to be able to, from the test, control the output of the randomizer. What about passing in a mocked `IRandomizer`?
- `Card.ToString()` needs to be tested.
- Style nitpick: Watch unnecessary empty lines and member naming conventions. One example is `PokerComboTests` with public static members in `camelCase` instead of `PascalCase`.

jdphenix:
Overall I'm pretty happy with the tests you have. Just a few minor-ish changes needed.

ghost:
Thanks for the feedback. 
I tought about using a theory /memberdata/theorryAtttribute.

But one thing I did not see.
A player has a name and a hand. 
But the hand is always a empty collection.
That is why I use now a lot the addCards.

Must I use a seperate function or so to make this work ?

ghost:
Another question : 

you asks me this :

I don't see any tests for Deck.

    You'll need to make a couple of changes to Deck to fully test it. You need to be able to, from the test, control the output of the randomizer. What about passing in a mocked IRandomizer?
 
but the homework is talking about this : 

Unit test ``/Gambling/Games.``. 

and Deck is not in Games  but in GameComponents. That is why I did not make the tests. 
For a lot of code I did already in HW1 but without the mock. 

If you and Kai decided that it is a part of this homework , then of course I will make it

Src/BootCamp.Chapter/Gambling/Games/Poker/Poker.cs
ghost: sorry, changed it back.

jdphenix:
> 
> 
> Another question :
> 
> you asks me this :
> 
> I don't see any tests for Deck.
> 
> ```
> You'll need to make a couple of changes to Deck to fully test it. You need to be able to, from the test, control the output of the randomizer. What about passing in a mocked IRandomizer?
> ```
> 
> but the homework is talking about this :
> 
> Unit test `/Gambling/Games.`.
> 
> and Deck is not in Games but in GameComponents. That is why I did not make the tests.
> For a lot of code I did already in HW1 but without the mock.
> 
> If you and Kai decided that it is a part of this homework , then of course I will make it

You're correct here. Don't worry about `Deck`. I actually posted my "extra homework" after looking this over.

ghost:
oke, I do not worry, I wanted things to be clear. 
Now I hope I can figure out how to test the GetWinners function with a theory.

Still stuck at the point that I need to use AddCards somehow to make the inputList

jdphenix:
> 
> 
> oke, I do not worry, I wanted things to be clear.
> Now I hope I can figure out how to test the GetWinners function with a theory.
> 
> Still stuck at the point that I need to use AddCards somehow to make the inputList

Yep, I'm going to look at this this evening.

ghost:
Oke,  I see when I get hints to solve this. So no sweat if you cannot help me this evening.

Tests/BootCamp.Chapter.Tests/PokerHandsTests.cs
Almantask: > * I don't see any tests for `Deck`.
>   
>   * You'll need to make a couple of changes to `Deck` to fully test it. You need to be able to, from the test, control the output of the randomizer. What about passing in a mocked `IRandomizer`?
> * `Card.ToString()` needs to be tested.
> * Style nitpick: Watch unnecessary empty lines and member naming conventions. One example is `PokerComboTests` with public static members in `camelCase` instead of `PascalCase`.

You don't have to inject randomized. That was  a tricky bit :) 
- Instead of injecting randomizer, setup a system under test in a way that there is only one outcome.

Tests/BootCamp.Chapter.Tests/PokerTests.cs
Almantask: As of now, it's fine as is. Just the whole setup part should be moved to a single method, that creates a game.

Tests/BootCamp.Chapter.Tests/PokerHandsTests.cs
ghost: moment, do i now to make test for deck or not. The assignment says only the game directory and deck is in the gameobjects file. That is why I did not make any test for it. Deck is already tested in challenge1

Tests/BootCamp.Chapter.Tests/PokerTests.cs
Almantask:
```cs
[Fact]
public void GetWinners_Given_List_Players_Returns_Winnner()
{
  (var game, var expectedWinner) = BuildGameWithAWinner();

  var winners = game.GetWinners();

  winners.Should().Contain(expectedWinner).And().Should().HaveCount(1);
}
```

ghost:
I make test for 1 winner and multiple winners.  I did not make any test for no winners because I think that will never happen. As far as I can see there is a winner or a draw.

Almantask:
https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/412#issuecomment-635772212

No winners means a game is started without players. Maybe that should throw an exception?

Almantask (changes requested):
Your score calculation is off. The point of it was, that same combo can have a different score. 
For example, a flush is a combo with score 1 that will be, if:
1, 2, 3, 4, 5 will happen.
Highest score royal flush will be:
10, J, Q, K, A.

So first part is good- you are able to check if combo is right, but now on top of that you need to check how can the combo be rated. Tip: check the lowest card.

ghost:
Oke

So if I understand you a flush of 1 ,2,3,4,5 gets a score of 1 
and a flush of 2.3.4.5 gets a score of 2 

how do I it then work if one has a flush of 1.2.3.4.5 which should have a score of 1. 
and the other has a full house of 1,1,2,2,2  which also should have a score of 1.
it would be a drawn but that is not according to the rules of poker as I understand it.

Almantask:
> Oke
> 
> So if I understand you a flush of 1 ,2,3,4,5 gets a score of 1
> and a flush of 2.3.4.5 gets a score of 2
> 
> how do I it then work if one has a flush of 1.2.3.4.5 which should have a score of 1.
> and the other has a full house of 1,1,2,2,2 which also should have a score of 1.
> it would be a drawn but that is not according to the rules of poker as I understand it.

Yup. The only combo that always has a score of 1 is Royal flush (same color, 10 to Ace)

Almantask:
> > Oke
> > So if I understand you a flush of 1 ,2,3,4,5 gets a score of 1
> > and a flush of 2.3.4.5 gets a score of 2
> > how do I it then work if one has a flush of 1.2.3.4.5 which should have a score of 1.
> > and the other has a full house of 1,1,2,2,2 which also should have a score of 1.
> > it would be a drawn but that is not according to the rules of poker as I understand it.
> 
> Yup. The only combo that always has a score of 1 is Royal flush (same color, 10 to Ace)

Because think about it: you always check if combo score <> 0. If it's not 0, then that means combo is hit. So on top of that you need to check which of the same combo amounts to a better score.

ghost:
sorry, I understand that if  someone has a flush of 5.6.7.8.9.10  has a a score of 5 and if another player which have 4,,5,6,7,8,9 has a score of 4 so the first person wins. 

but I still do not see it when someone has a flush of 1.2.3.4.5 which should have a score of 1because the lowest card is a 1 
And another person has a three of a kind of 1,1,1,2,2 which has also a score of 1 because the lowest card is a 1. 
How does the code then finds out that the flush must win.

Almantask:
> sorry, I understand that if someone has a flush of 5.6.7.8.9.10 has a a score of 5 and if another player which have 4,,5,6,7,8,9 has a score of 4 so the first person wins.
> 
> but I still do not see it when someone has a flush of 1.2.3.4.5 which should have a score of 1because the lowest card is a 1
> And another person has a three of a kind of 1,1,1,2,2 which has also a score of 1 because the lowest card is a 1.
> How does the code then finds out that the flush must win.
Refer to how combo is found altogether:
```cs
private PokerCombo FindCombo(IEnumerable<Card> cards)
        {
            foreach (var comboCheck in _comboChecks)
            {
                var calculateScore = comboCheck.Value;
                var score = calculateScore(cards);
                var isSuccessfulCombo = score > 0;
                if (isSuccessfulCombo)
                {
                    var set = comboCheck.Key;
                    return new PokerCombo(set, score);
                }
            }

            throw new InvalidGameStateException("Ran out of possible combos.");
        }
```
For every possible combo, we check if it's possible. We start from highest combo. The higher combo wins. The only reason for having score was distinction of same combo, but more powerful cards.

Most important part: **var isSuccessfulCombo = score > 0;**
So first combo <> 0 is the winning combo already.

Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs
```diff
@@ -202,6 +214,56 @@ public static int CalculateHighCardScore(IEnumerable<Card> cards)
             return 0;
         }
 
+        public static int CalculateLowestCardScore(IEnumerable<Card> cards)
+        {
+            var sorted = cards.OrderBy(card => card.Rank).ToArray();
+            var lowestCard = sorted.First().Rank;
+
+            switch (lowestCard)
```
Almantask, line 222: use pattern matching ( 
```
switch lowestCard
{
  Rank.Value1 => score1,
  Rank.Value2 => score2,
  _ => throw..
}

File: Src/BootCamp.Chapter/Gambling/Games/Poker/PokerCombo.cs [hunk 5]
```diff
@@ -202,6 +214,56 @@ public static int CalculateHighCardScore(IEnumerable<Card> cards)
             return 0;
         }
 
+        public static int CalculateLowestCardScore(IEnumerable<Card> cards)
```
Almantask, line 217: Who's going to test what happens when 2 players have the same combos, but different score?

Almantask (changes requested):
Missing tests for people with same combos, but different scores of same combo. I think you need a test for each combo (single theory) for that.

Almantask (approved):
The last part was not very clear. The documentation part was lacking (I could not clearly understand what the tests mean (I understand, but it's not self-documenting).

PR #413, opened 2020-05-24
Chapter 1 Homework 3

Sqelson:
Tests appear to pass and program works as I intended and should fulfil the criteria, found grasping the syntax and where to put things quite difficult at first and a surprising step up from the previous homework but looking at examples of functions and classes online helped a lot.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace BootCamp.Chapter
+﻿namespace BootCamp.Chapter
 {
     /// <summary>
     /// Test class is used to test your implementation.
@@ -16,26 +12,22 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Lesson3.PromptInt(message);
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
+            return Lesson3.PromptString(message);
         }
 
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Lesson3.PromptFloat(message);
         }
 
-        public static float CalculateBmi(float weight, float height)
+        public static double CalculateBMI(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Lesson3.CalculateBMI(weight, height);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Tests/BootCamp.Chapter.Tests/Lesson3Tests.cs
```diff
@@ -16,7 +14,7 @@ public class Lesson3Tests
         [InlineData(100, 10, 1)]
         public void CalculateBmi_With_ValidInput_Returns_Expected(float weightKg, float heightM, float expectedBmi)
         {
-            var bmi = Checks.CalculateBmi(weightKg, heightM);
```
jdphenix: You should leave the test class alone for now. Just make sure the method is called CalculateBmi (note the capitalization).

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static string PromptString(string message)
+        {
+            Console.WriteLine(message);
+            return Console.ReadLine();
+        }
+
+        public static float PromptFloat(string message)
+        {
+            Console.WriteLine(message);
+            return float.Parse(Console.ReadLine());
```
jdphenix: Consider `float.Parse(string, IFormatProvider)` - you can pass in `CultureInfo.InvariantCulture` here.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +12,22 @@ public static class Checks
...
 
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Lesson3.PromptFloat(message);
         }
 
-        public static float CalculateBmi(float weight, float height)
+        public static double CalculateBMI(float weight, float height)
```
jdphenix: Keep this returning `float`

jdphenix (changes requested):
I'm pleased with this, just a couple of minor tweaks. You shouldn't have to change the Tests project.

Tests/BootCamp.Chapter.Tests/Lesson3Tests.cs
Sqelson: Ok, I will leave the tests alone in the future until explicitly told to edit them.

Src/BootCamp.Chapter/Lesson3.cs
Sqelson: Took some time but figured out why this was required, and noticed it was on the wiki to implement it, thanks for the tip!

Src/BootCamp.Chapter/Checks.cs
Sqelson: Changed it back to float.

Sqelson:
Thanks for the feedback, I have implemented the changes recommended, and also added a few comments as I realised I hadn't put any in yet.

jdphenix (approved):
Excellent work, thanks!

PR #414, opened 2020-05-25
Chapter 1/Homework 1

aysenti:
Here's my attempt at homework 1 (the bmi one), and homework 2 which I'm assuming was just forking, branching, and then doing a PR.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -23,10 +23,35 @@ namespace BootCamp.Chapter
 
             Console.Write("Hello world!" + System.Environment.NewLine);
 
-            Console.WriteLine("Tom's age is " + tAge + "and his height is " + tHeight + " and his weight is " + tWeight + " BMI is: " + tBmi);
-            Console.WriteLine("Colby's age is " + cAge + "and his height is " + cHeight + " and his weight is " + cWeight + " BMI is: " + cBmi);
+            Console.WriteLine("Tom's age is " + tAge + " and his height is " + tHeight + " and his weight is " + tWeight + " BMI is: " + tBmi);
+            Console.WriteLine("Colby's age is " + cAge + " and his height is " + cHeight + " and his weight is " + cWeight + " BMI is: " + cBmi);
             Console.WriteLine("Blakes's age is " + bAge + " and his height is " + bHeight + " and his weight is " + bWeight + " BMI is: " + bBmi);
+
+            Console.WriteLine("Press any key to continue...");
+
             Console.ReadLine();
+            Console.Clear();
+
+        Start:
+
+            Console.WriteLine("Please enter your name. ");
+            string yourName = Console.ReadLine();
+            Console.WriteLine("Hello " + yourName);
+
+            Console.WriteLine("Please enter your age now. ");
+            int yourAge = Convert.ToInt32(Console.ReadLine());
+            Console.WriteLine("Please enter your height now. ");
+            double yourHeight = double.Parse(Console.ReadLine());
+            Console.WriteLine("Please enter your weight now. ");
+            int yourWeight = Convert.ToInt32(Console.ReadLine());
+            double yourBMI = yourWeight / yourHeight;
+            Console.WriteLine("Alright " + yourName + ", " + "you are " + yourAge + " years old, and your BMI based off your provided measurements is " + yourBMI);
+
+            Console.WriteLine("Press any key to continue...");
+
+            Console.ReadLine();
+            Console.Clear();
+            goto Start;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,27 @@ class Program
...
+
+            int cWeight = 68;
+            float cHeight = 1.55F;
+            float cBmi = cWeight / cHeight;
+            int cAge = 20;
+
+            int bWeight = 61;
+            float bHeight = 1.41F;
+            float bBmi = bWeight / bHeight;
+            int bAge = 32;
```
jdphenix: Looking good so far! You're still needing to read in values from the console using `Console.ReadLine()`. 

    // getting a string
    Console.WriteLine("Enter a string.");
    string value = Console.ReadLine();

There's still a bit to do for numeric variables. Take a look at the [Finding missing pieces (Console)](https://github.com/csinn/CSharp-From-Zero-To-Hero/tree/Chapter1/Homework/1And2#finding-missing-pieces-console) section for ideas.

jdphenix (changes requested):
Looking good so far. Don't forget the spec asks you to implement prompting for a few key data values and outputting them back out, then repeat it again for another person. 

Ref (https://github.com/csinn/CSharp-From-Zero-To-Hero/tree/Chapter1/Homework/1And2#code)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -23,10 +23,35 @@ static void Main(string[] args)
...
+            Console.WriteLine("Please enter your weight now. ");
+            int yourWeight = Convert.ToInt32(Console.ReadLine());
+            double yourBMI = yourWeight / yourHeight;
+            Console.WriteLine("Alright " + yourName + ", " + "you are " + yourAge + " years old, and your BMI based off your provided measurements is " + yourBMI);
+
+            Console.WriteLine("Press any key to continue...");
+
             Console.ReadLine();
+            Console.Clear();
+            goto Start;
```
jdphenix: Yipe! 

Consider a simple way of asking for two people's data without using `goto`. One way might be: 

    for (int i = 0; i < 2; i++)
    {
      // Ask and output here
    }

If you haven't seen a [for loop](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/for) before, that's fine! Basically, what this does is declare a variable `i`, makes sure that `i < 2`, then runs the code inside the brackets, then adds 1 to `i` and repeats. 

I definitely want to discourage to use of `goto`, in any circumstance.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -23,10 +23,35 @@ static void Main(string[] args)
...
+            Console.WriteLine("Hello " + yourName);
+
+            Console.WriteLine("Please enter your age now. ");
+            int yourAge = Convert.ToInt32(Console.ReadLine());
+            Console.WriteLine("Please enter your height now. ");
+            double yourHeight = double.Parse(Console.ReadLine());
+            Console.WriteLine("Please enter your weight now. ");
+            int yourWeight = Convert.ToInt32(Console.ReadLine());
+            double yourBMI = yourWeight / yourHeight;
+            Console.WriteLine("Alright " + yourName + ", " + "you are " + yourAge + " years old, and your BMI based off your provided measurements is " + yourBMI);
```
jdphenix: Note the spec for this assignment would have you ask for name and surname.

jdphenix (changes requested):

Src/BootCamp.Chapter/Program.cs
aysenti: Alright I reworked it to ask for surname and to also not use goto

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,57 +1,57 @@
...
+                string surname = Console.ReadLine();
+                Console.WriteLine("Hello " + firstName + " " + surname);
+
+                Console.WriteLine("Please enter your age now. ");
+                int yourAge = Convert.ToInt32(Console.ReadLine());
+                Console.WriteLine("Please enter your height now. ");
+                double yourHeight = double.Parse(Console.ReadLine());
+                Console.WriteLine("Please enter your weight now. ");
+                int yourWeight = Convert.ToInt32(Console.ReadLine());
+                double yourBMI = yourWeight / yourHeight;
```
jdphenix, line 44: I should have caught this on the first review. bmi is weight over height squared.

jdphenix (approved):
Overall happy with this program. Keep in mind sticking to what's asked of the spec and coding style (extra empty lines).

Src/BootCamp.Chapter/Program.cs
aysenti: Oh my bad I also should have noticed that when I looked up the formula

PR #416, opened 2020-05-25
Chapter1 homework 10

mkasarci:


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Net.Http;
```
jdphenix, line 3: Not needed.

jdphenix (approved):
Good job, I only have nitpicks. 

As a little bit of a bonus for yourself, consider what you might ought to do if you get a string passed that results in a ciphertext outside the bounds of the alphabet (for example, input "zoo", shift right 3).

PR #417, opened 2020-05-25
Chapter1/homework/8

inz-:


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +13,324 @@ public static class BalanceStats
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            if (IsStringArrayNullOrEmpty(peopleAndBalances))
+                return "N/A.";
```
Almantask, line 17: Can be moved to const

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +13,324 @@ public static class BalanceStats
...
+
+            //Convert to Person-Object
```
Almantask, line 19: Is the comment really needed? The same can be read from code itself.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +13,324 @@ public static class BalanceStats
...
+            Person[] people = DataToObject(peopleAndBalances);
+
+            //Find highest value and number of people with same balance
```
Almantask, line 22: Without understanding OOP, it might be really hard to wrap your head around this, but in near future, starting from chapter 2, blocks like these should really be extracted to functions and multiple values should be returned in objects.

Almantask (approved):
For now okay, but it would be great if you will look back at it and improve it after after chapter 2

PR #419, opened 2020-05-26
Chapter3/homework/3

DavidKatona:
I've added null and other kinds of validations to ctors and most other functions.

Where it was possible, I left these validations to be handled by their corresponding classes, i.e. AddItems inside the Player class could have had a null validation but since it calls out to the Inventory class to do the same, I only added validations to the inventory. This way, null validations are not done twice.

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -1,5 +1,4 @@
 ﻿using BootCamp.Chapter.Items;
-using System.ComponentModel;
 
 namespace BootCamp.Chapter
 {
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -21,14 +21,14 @@ namespace BootCamp.Chapter
             }
             set
             {
-                if (value == null) return;
+                if (value is null) return;
 
-                TotalAttackValue = (value == null) ? 0.0f : value.AttackValue;
+                TotalAttackValue = value.AttackValue;
 
-                var currentWeight = (_weapon == null) ? 0.0f : Weapon.Weight;
-                var newWeight = (value == null) ? 0.0f : value.Weight;
-
+                var currentWeight = (_weapon is null) ? 0.0f : Weapon.Weight;
+                var newWeight = value.Weight;
                 TotalCombinedWeight += newWeight - currentWeight;
+
                 _weapon = value;
             }
         }
@@ -147,13 +147,13 @@ namespace BootCamp.Chapter
 
         private void RefreshArmorStatistics(Armor currentArmor, Armor newArmor)
         {
-            var currentDefenseValue = (currentArmor == null) ? 0.0f : currentArmor.DefenseValue;
-            var newDefenseValue = (newArmor == null) ? 0.0f : newArmor.DefenseValue;
+            var currentDefenseValue = currentArmor?.DefenseValue ?? 0;
+            var newDefenseValue = newArmor?.DefenseValue ?? 0;
 
             TotalDefenseValue += newDefenseValue - currentDefenseValue;
 
-            var currentWeight = (currentArmor == null) ? 0.0f : currentArmor.Weight;
-            var newWeight = (newArmor == null) ? 0.0f : newArmor.Weight;
+            var currentWeight = currentArmor?.Weight ?? 0;
+            var newWeight = newArmor?.Weight ?? 0;
 
             TotalCombinedWeight += newWeight - currentWeight;
         }
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -1,4 +1,5 @@
 ﻿using BootCamp.Chapter.Items;
+using System.ComponentModel;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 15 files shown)


Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,47 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection.Metadata.Ecma335;
```
jdphenix: Watch for unused usings - you shouldn't need this.

Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using BootCamp.Chapter.Items;
+using System.ComponentModel;
```
jdphenix: Unused using.

Src/BootCamp.Chapter/Items/Armor.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
jdphenix, line 3: Unused usings.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,28 +1,22 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Collections.Generic;
```
jdphenix: Unused using.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,28 +1,22 @@
...
-
-        private Inventory _inventory;
+        public Inventory Inventory { get; private set; }
+        public decimal Money { get; private set; }
 
         public Shop()
         {
-
+            Inventory = new Inventory();
         }
```
jdphenix: I don't think this constructor is ever used, consider removing it.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -1,28 +1,22 @@
...
     public class Shop
     {
-        private decimal _money;
-        public decimal GetMoney()
-        {
-            return _money;
-        }
-
-        private Inventory _inventory;
+        public Inventory Inventory { get; private set; }
```
jdphenix: Remove private setter - you never use it.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,47 @@
...
+            }
+
             return new Item[0];
         }
 
         public void AddItem(Item item)
         {
+            if (RemainingSpace == 0) return;
 
+            Items.Add(item ?? throw new ArgumentNullException($"{nameof(item)} cannot be null."));
```
jdphenix, line 42: Consider consistent behavior with your other methods here. One situation that you might see unexpected behavior - 

- Remaining space is 0. 
- `AddItem(null)` is called. 
- Method returns, making no change. 

Compared to 

- RemainingSpace > 0. 
- `AddItem(null)` is called. 
- Throws an `ArgumentNullException`. 

I think this method should *always* throw an `ArgumentNullException` if `item is null`.

Src/BootCamp.Chapter/Inventory.cs
```diff
@@ -1,26 +1,47 @@
...
         public Inventory()
         {
-            _items = new Item[0];
+            Size = defaultInventorySize;
+            RemainingSpace = defaultInventorySize;
+            Items = new List<Item>(defaultInventorySize);
         }
 
-        public Item[] GetItems(string name)
+        public Item[] GetItem(string name)
```
jdphenix: This method needs to either be named `GetItems` and return `Item[]`, or `GetItem` and return `Item`.

jdphenix (changes requested):
Overall good job, just a few nitpicks.

Src/BootCamp.Chapter/Inventory.cs
DavidKatona: Ouch, you're right, I must have added this accidentaly. I've removed all unused namespaces from all classes.

Src/BootCamp.Chapter/Shop.cs
DavidKatona: I've removed the unused ctor from Shop.

Src/BootCamp.Chapter/Shop.cs
DavidKatona: Removed unused private setter for the Inventory component in Shop.

Src/BootCamp.Chapter/Inventory.cs
DavidKatona: You're right on the fact that it should always throw an **ArgumentNullException** if **item is null**.

The 1st behaviour, however, doesn't seem to be right.

- Remaining space is 0. -> returns immediately and never calls AddItem(null).

So in my head, it seemed more logical to check whether the inventory has any available spaces left or not and only worry about the item if there is at least one. Though you might be right, and maybe it would be better to check for null before all.

This option, however, results in a bit less code as I'd have to call something like:
if (item is null) throw new ArgumentNullException(); before the the rest of the method's body.

I'm not sure though, maybe in the long run it would be beneficial to have the null-check first and that would remain consistent with other methods that do null-checking as the first step.

Src/BootCamp.Chapter/Inventory.cs
DavidKatona: Agree!

I've refactored this method so it returns an Item now. It lead to easier to read code in other classes that call this method.

I think the confusion came from the fact that some of the tests that are written for the original GetItems(string name) method in the Player class, expected an array as a result. It doesn't make much sense to be honest as it was designed to return a single item.

Anyways, this should be corrected now. GetItem() can return null as well but since there are plenty of null validations the code is still functional. :)

Almantask (approved):
Great!

PR #420, opened 2020-05-26
Chapter2/Homework/3

mkasarci:


Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -5,11 +5,49 @@
...
-    public class MsFactory
+    public class MsFactory : PcFactory
     {
-        public DesktopComputer Assemble()
+        private Body _msBody;
+        private Cpu _msCpu;
+        private Gpu _msGpu;
+        private HardDisk _msHardDisk;
+        private Motherboard _msMotherboard;
+        private Ram _msRam;
```
jdphenix: Remove these fields.

Src/BootCamp.Chapter/PcFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+            AssembleRam();
+            return new DesktopComputer(body, cpu, gpu, disk, board, ram);
+        }
+
+        public abstract void AssembleBody();
+        public abstract void AssembleCpu();
+        public abstract void AssembleGpu();
+        public abstract void AssembleHardDisk();
+        public abstract void AssembleMotherboard();
+        public abstract void AssembleRam();
```
jdphenix: Naming nitpick - what exactly are these methods assembling? Maybe `PrepareXXX` might be better? 

Also, make these `protected`. As a rule, don't make something `public` until you've got a proven use case where it's necessary.

jdphenix (changes requested):
One important aspect of this homework is to demonstrate understanding of how derived types can be used. Make these changes, 

(1) Remove the private fields from `MsFactory` and `MacFactory`. 
(2) Change the design of `MsFactory` and `MacFactory` to set up derived types of the components. For example, `MacFactory.AssembleMotherboard()` should be setting up a new `MacMotherboard`, while `MsFactory` should be setting up a new `MsMotherboard`.

mkasarci:
Thanks for all the reviews that you've made it helps me a lot. I've changed the program as you requested. I hope I understood you right. Can you check the code again.

Src/BootCamp.Chapter/PcFactory.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using BootCamp.Chapter.Computer;
+
+namespace BootCamp.Chapter
+{
+    public abstract class PcFactory
+    {
+        protected Body body;
```
Almantask: IMagine that your factory is really popular and you have multiple assembles happening at once. Do you really want to afford member fields storing info on what is being assembled? Preparing each part means creating it and attaching it to a PC. The flow is defined and known for both factories. Just individual steps are different.

Src/BootCamp.Chapter/PcFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+        protected Cpu cpu;
+        protected Gpu gpu;
+        protected HardDisk disk;
```
Almantask: move fields to local scope

Src/BootCamp.Chapter/PcFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+        protected Motherboard board;
+        protected Ram ram;
+
+        public DesktopComputer Assemble()
+        {
+            PrepareBody();
```
Almantask: Attach (Assemble) a part to the pc.

Almantask: Right now it kinda does it but you need to either:
A) return a created part
B) pass a PC and install the created part in the passed empty PC

Almantask (changes requested):

Src/BootCamp.Chapter/PcFactory.cs
mkasarci: I changed the methods and return the parts

Src/BootCamp.Chapter/PcFactory.cs
mkasarci: I've copletely deleted them hope it is right :D

jdphenix (approved):
Overall pretty happy with your changes.

Almantask (commented):
The illustration could be slightly better (mac and PC having different individual parts, like GeforceGpu vs MacGpu) but the idea of template method is done (though impractical in your illustration)

PR #422, opened 2020-05-27
Chapter3/Homework/1

mkasarci:
Refactored code and used Generic lists and properties. Thanks God I don't have to use arrays anymore to add something to list.

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -7,58 +7,78 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
+        private float _totalWeight;
+        private float _totalDefense;
+        private float _totalAttack;
+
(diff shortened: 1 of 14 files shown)


Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,58 +7,78 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
+        private float _totalWeight;
+        private float _totalDefense;
+        private float _totalAttack;
+
+        //I haven't change those to properties cause it has so many if statements.
```
Almantask, line 14: It's not a very good argument though. The only time for method vs property when it should not be a property, is if it does too much (takes long too). This doesn't do too much.

Almantask (approved):
Legit looks good!

jdphenix (approved):

PR #425, opened 2020-05-28
Chapter3/Homework/2

mkasarci:
I had weird problem on the tests with the default .txt files. So I've created my own files to test. I've got directory not found exception with default files but I've pass all the tests with the files that I've created on desktop.

### DIFF
--- a/Src/BootCamp.Chapter/Credentials.cs
+++ b/Src/BootCamp.Chapter/Credentials.cs
@@ -1,22 +1,54 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace BootCamp.Chapter
 {
     // TODO: make a struct and add validation and other needed methods (if needed)
-    public class Credentials
+    public readonly struct Credentials
     {
-        public string Username;
-        public string Password;
+        public readonly string Username;
+        public readonly string Password;
 
         public Credentials(string username, string password)
         {
+            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(password)) throw new ArgumentException("user name and password not be empty.");
+
             Username = username;
             Password = password;
         }
 
+        //Credentials.TryParse(credentialsString, out var credentials);
+
         // TODO: Implement properly.
         public static bool TryParse(string input, out Credentials credentials)
         {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,22 +1,54 @@
...
+    public readonly struct Credentials
     {
-        public string Username;
-        public string Password;
+        public readonly string Username;
+        public readonly string Password;
 
         public Credentials(string username, string password)
         {
+            if (string.IsNullOrEmpty(username) || string.IsNullOrEmpty(password)) throw new ArgumentException("user name and password not be empty.");
```
Almantask, line 14: It's one hell of a line 😄 
Try to be more specific at what is wrong, don't make people guess. An improved version of this would be:

```cs
if(string.IsNullOrEmpty(username) throw new ArgumentException(...);
if(string.IsNullOrEmpty(password) throw new ArgumentException(...);
...
```

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -1,24 +1,93 @@
...
+            return Encoding.Unicode.GetString(byteArray); ;
+        }
+
         // TODO: store credentials in credentials file.
         public void Register(Credentials credentials)
         {
+            if (!Credentials.credentialsNotNull(credentials)) throw new ArgumentException("Credentials not be empty.");
+
+            StringBuilder stringBuilder = CreateRegisterString(credentials);
+            //D:\Desktop\CSharp-From-Zero-To-Hero\Src\BootCamp.Chapter\CredentialsManager.cs
```
Almantask, line 67: Not sure what those comments mean, burn it with fire!

Almantask (approved):
Nothing gets in my way. Just try to be more consistent with the way you write var.
For example, use it nearly always, unless it's not possible to tell what the type is.

Src/BootCamp.Chapter/Credentials.cs
mkasarci: Haha you are right I'll be more carefull about it.

mkasarci:
Thanks for the review. You guys are great!

PR #427, opened 2020-05-29
Homework 3 from chapter 1

joshwu8:
Attempt at Homework 3.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -1,14 +1,62 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Transactions;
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,14 +1,62 @@
...
 namespace BootCamp.Chapter
 {
     class Program
+
+
     {
         static void Main(string[] args)
         {
+
+            BmiCalculator();
```
Almantask, line 16: CalculateBmi is a better name. 
Methods are verbs, not nouns.

Also, be careful with random empty lines all over the place. Just remove them. Code looks really dirty with them.

Almantask (approved):
Overall is good enough 😄

PR #428, opened 2020-05-29
Chapter3/Homework/3

mkasarci:
Code refactored. Used new operators I've learnt.

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -12,78 +12,89 @@ namespace BootCamp.Chapter
         private float _totalDefense;
         private float _totalAttack;
 
-        public Weapon Weapon { get; set; }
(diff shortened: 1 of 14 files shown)


Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,58 +8,82 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
-        private Weapon _weapon;
+        private float _totalWeight;
+        private float _totalDefense;
+        private float _totalAttack;
+
+        public Weapon Weapon { get; set; }
```
Almantask: You have duplicate implementation of every setter. Instead of using auto properties, move setter method logic to setter property.

Almantask (changes requested):

Almantask (approved):
Looks good now

PR #429, opened 2020-05-30
Homework 1 & 2

mbmaeso:
Results of homeworks 1 and 2


Almantask (changes requested):
Please read all input from console.
Refer to:
`Console.ReadLine()`
`int.Parse(input)`
`double.Parse(input)`

Almantask (approved):
Good stuff 😄 You can proceed!

PR #430, opened 2020-05-30
Homework 3

mbmaeso:
Solution for homework 3

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -7,15 +7,29 @@ namespace BootCamp.Chapter
     {
         public static void Demo()
         {
-            string name = StringPrompt("Please introduce a name:");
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            string name = StringPrompt("Please introduce a name:");
```
Almantask: 2nd person is missing;
Your demo could also call another method: PromptPersonBmi()

Almantask (changes requested):

Almantask (approved):
Better, but you could have simplified it further by extracting logic in a function and just calling it twice, instead of looping 😄

PR #431, opened 2020-05-30
Chapter 1/Homework 3

arthur-schevey:
A little confused on the `Check.cs` file but hopefully it's correct.
I named my functions in `Program.cs` the same as the functions in `Checks.cs`

In `Program.cs` I made four functions that display a message while requesting a value.
I used these functions in `Main` and assigned each of the requested values to variables such as name, age, height, etc.

I then used these variables to calculate BMI and print the information back to the user.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,22 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Program.PromptInt(message);
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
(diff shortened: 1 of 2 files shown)


Almantask:
> A little confused on the `Check.cs` file but hopefully it's correct.
> I named my functions in `Program.cs` the same as the functions in `Checks.cs`
> 
> In `Program.cs` I made four functions that display a message while requesting a value.
> I used these functions in `Main` and assigned each of the requested values to variables such as name, age, height, etc.
> 
> I then used these variables to calculate BMI and print the information back to the user.

Checks file was used for 2 reasons:
1) Automatically test your functions
2) Give you freedom for defining functions the way you want. 

I cannot achieve both 1 and 2 without the checks file. I did not want to enforce special function rules on you, other than just putting what you did in checks. 

You did well.

Almantask (approved):
Looks good!

PR #432, opened 2020-05-31
Chapter3/homework/4

mkasarci:
I think it works as it should but, I couldn't find the place where I should use the in and out keywords.

resim


Almantask (approved):
Semantically you did the job, practically, not much use. Though the point of this homework was not that (thought it should have been). Indeed, there are no applicable places where in/out can be used, unless you split interfaces until a single method is there- then you can apply in for input and out for output generic.

PR #433, opened 2020-06-01
Chapter3/homework/5

mkasarci:


Src/BootCamp.Chapter/MostCommonLetterFinder.cs
```diff
@@ -1,10 +1,51 @@
...
+
+namespace BootCamp.Chapter
 {
     public static class MostCommonLetterFinder
     {
         public static char Find(string sentence)
         {
-            return 'z';
+            if (string.IsNullOrWhiteSpace(sentence)) throw new ArgumentNullException();
+                Dictionary<char, int> testCollection = StringToCollection(sentence);
```
Almantask, line 13: Try to get into habit of using var.

Src/BootCamp.Chapter/ToggleableGrid2d.cs
```diff
@@ -1,16 +1,47 @@
...
+
+        private void PrintArray()
+        {
+            _gridClearer.Clear();
+
+            for (int i = 0; i < _boolArray.GetLength(0); i++)
+            {
+                for (int j = 0; j < _boolArray.GetLength(1); j++)
+                {
+                    Console.Write(_boolArray[i, j] == true ? "■" : " ");
```
Almantask, line 37: these magic strings should go as constants all the way at the top of the class. It's likely to change, but you don't want to look for every place where it was used.

Almantask (approved):
Good job, I like what I see.

PR #435, opened 2020-06-01
Chapter1/homework/4

joshwu8:
Chapter 1 Homework 4 attempt.


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -0,0 +1,113 @@
...
+            var input = Console.ReadLine();
+            float number;
+            var floatCheck = float.TryParse(input, out number);
+            if (!floatCheck)
+            {
+                Console.WriteLine(number);
+                Console.WriteLine(input + " is not a valid number");
+                return -1;
+            }
+            if (float.Parse(input) == 0)
```
Almantask, line 70: Not sure why:
```
            if (float.Parse(input) == 0)
            {
                return 0;
            }
            if (floatCheck)
            {
                return float.Parse(input);
            }
```
can't be just
```
return number;
```
...because you already parsed it.

Almantask (changes requested):
Remove redundant ifs in the Prompt functions.

PR #436, opened 2020-06-02
Chapter3/homework/6

mkasarci:


Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,48 @@
...
+                    string[] personInfo = streamReader.ReadLine().Split(',');
+
+                    GetPersonInfoAndAddPersonList(personInfo);
+                }
+            }
+        }
+
+        private void GetPersonInfoAndAddPersonList(string[] personInfo)
+        {
+            Person person = new Person(personInfo[0], personInfo[1], formatAndParseDateTime(personInfo[2]), personInfo[3], personInfo[4], personInfo[5], personInfo[6]);
```
jdphenix: Consider eliminating the magic numbers for column order here. Naming constants instead is much more readable, e.g. 

    const int FirstNameField = 0;
    // later
    Person person = new Person(
        personInfo[FirstNameField],
        personInfo[SurnameField], ...

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,48 @@
...
+
+        private void LoadList(string peopleFile)
+        {
+            if (string.IsNullOrEmpty(peopleFile)) throw new ArgumentNullException();
+            if (!File.Exists(peopleFile)) throw new FileNotFoundException();
+            using (StreamReader streamReader = new StreamReader(peopleFile))
+            {
+                if (string.IsNullOrEmpty(streamReader.ReadLine())) throw
+                new ArgumentNullException("The file hasn't got any data.");
+                bool firstIndexFlag = true;
```
jdphenix: Unneeded, remove.

jdphenix (changes requested):
Good evening, 
This is pretty good so far! I've got a couple of specific comments I've noted, and one comment about the design of the demo that will need to be addressed. 

It is expected the demo is implemented using handling the specified events, 

a) DemoStarted
b) Example selected (from 2) a, b, c)
c) DemoEnded
d) Application closed

I would start with creating a `Demo` class with event handlers you can subscribe to, like `Started`, `SelectionMade`, etc, so they could be used like `demo.Started += Demo_Started;`

Src/BootCamp.Chapter/ContactsCenter.cs
mkasarci: I got exception when I deleted this. Because first line of the .csv file is contain header. I've just skip the first line which contains header when I check streamReader.ReadLine() is empty. And I start my while loop at 2nd line of he .csv file with that code. I couldn't workaround this issue without these codes. Do you have any advice?

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,34 @@
...
     {
+        public static bool flag = true;
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            
+            while (flag)
+            {
+                Demo.Demo_Started += Demo.Started;
+                Demo.Demo_Started?.Invoke(Selection(), EventArgs.Empty);
```
Almantask, line 18: The whole point of events is that the implementation for handler comes from outside. The Demo.Started delegate should be a delegate outside of demo itself. Else, you could just call Demo.Started, right?

`Demo_Started` is good, if you had the right name for it `DemoStarted` is a better name.

Almantask: Your demo started, should actually call the selection in that case. right now, you manually invoke an extra method on top of event. Be more organic:
`Demo.DemoStarted += Selection;`

Almantask (changes requested):

Almantask:
EXpired

PR #437, opened 2020-06-02
Chapter1/homework/1 and2 jimmy

PunchyJ:
Removed Console.WriteLine("Hello World");
Added code to display name, age, height and weight.

Calculates BMI.

Added range indicator for BMI because as my first code I wanted to see if I was able to do that.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -1,12 +1,32 @@
 ﻿using System;
 
-namespace BootCamp.Chapter
+namespace BootCamp
 {
-    class Program
+    class BMI1
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string Name = "James Ferguson";
+            int Age = 30;
+            var Sex = "his";
+
+            double Height = 195.0;
+            var heightMetres = Height / 100.0;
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,32 @@
...
+            double Weight = 96.4;
+            double BMI = Weight / heightMetresMultiplied;
+            BMI = Convert.ToInt32(BMI);
+
+            Console.WriteLine($"{Name} is {Age} years old, {Sex} weight is {Weight} kg and {Sex} height is {Height}cm");
+
+            if (BMI < 18.5)
+            { Console.WriteLine($"Your BMI is {BMI}. Which is Underweight. Go have a piece of cake :("); }
+            if (BMI >= 18.5 & BMI <= 24.9)
+            { Console.WriteLine($"Your BMI is {BMI}. Which is Normal. Good stuff just a small piece of cake should do :)"); }
```
Almantask, line 25: By convention, it's easier to read when you put brackets in separate lines.
```cs
{
 // code
}
```
Input should be read from console. Refer to
Console.ReadLine and int.Parse, float.Parse

Almantask (changes requested):

PR #438, opened 2020-06-03
Chapter1/homework/4

arthur-schevey:
### What I did: 
In all the methods I added conditions to check if the inputs were actually valid. If not they would return an error code (`-1 or -`) as instructed.

The `BmiPrompt()` function then checks for these error codes while executing other code. If the error codes are present, error text will be presented to the user. Otherwise, it will function as normal.

### Issues:
Having significant issues failing checks even though it's seemingly correct in the console. I don't really know what I am doing wrong, the test says that it expected a particular response but received nothing.

Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`

Currently I get `"-1" is not a valid number.` because the prompt function returns -1 and not the original input.

All I can think of is doing is calling the variable from a prompt function "UserInput" `var UserInput = Console.ReadLine()` from the main class but I am pretty sure this is not possible.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,25 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            return Program.PromptInt(message);
         }
 
         public static string PromptString(string message)
         {
             // To do: call your implementation. 
-            return "";
+            return Program.PromptString(message);
         }
 
         public static float PromptFloat(string message)
         {
             // To do: call your implementation. 
-            return 0;
+            return Program.PromptFloat(message);
         }
 
         public static float CalculateBmi(float weight, float height)
         {
             // To do: call your implementation. 
-            return 0;
+            return Program.CalculateBmi(weight, height);
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -43,9 +43,9 @@ namespace BootCamp.Chapter
             // check for errors during BMI calculation; otherwise state the BMI
             if (determinedBmi == -1)
             {
-                Console.Write("\nFailed calculating BMI. Reason: ");
-                if (height <= 0) Console.WriteLine("Height cannot be less than zero, but was " + height + ". \n");
-                if (weight <= 0) Console.WriteLine("Weight cannot be equal or less than zero, but was " + weight + ". \n");
+                Console.WriteLine("\nFailed calculating BMI. Reason: ");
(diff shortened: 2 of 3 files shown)


Almantask:
> Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`

Printing to a console and returning a value are two different things.
'ffav is not a valid number' is what will be printed, while -1 is what will be returned. 😄

Almantask:
> > Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`
> 
> Printing to a console and returning a value are two different things.
> 'ffav is not a valid number' is what will be printed, while -1 is what will be returned. 😄

So print to console from the function itself, not outside of it.

arthur-schevey:
> 
> 
> > Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`
> 
> Printing to a console and returning a value are two different things.
> 'ffav is not a valid number' is what will be printed, while -1 is what will be returned. 😄

I'm such a fool, you're such a genius. Thank you so much, I don't know why my brain completely blocked me out from doing that.

I still seem to be failing checks even though I believe it's correct in console. Is there something I have done wrong?

Almantask:
> > > Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`
> > 
> > 
> > Printing to a console and returning a value are two different things.
> > 'ffav is not a valid number' is what will be printed, while -1 is what will be returned. 😄
> 
> I'm such a fool, you're such a genius. Thank you so much, I don't know why my brain completely blocked me out from doing that.
> 
> I still seem to be failing checks even though I believe it's correct in console. Is there something I have done wrong?

Yup, error messages are all wrong. Sometimes you don't print it at all, sometimes you do WrtieLine where Write is needed, etc. Be more careful the way you do it.

Almantask:
> > > > Also, I am completely lost on how to return an error code while also returning the original user input. For example user inputs `Age: ffav` --> `"ffav" is not a valid number.`
> > > 
> > > 
> > > Printing to a console and returning a value are two different things.
> > > 'ffav is not a valid number' is what will be printed, while -1 is what will be returned. 😄
> > 
> > 
> > I'm such a fool, you're such a genius. Thank you so much, I don't know why my brain completely blocked me out from doing that.
> > I still seem to be failing checks even though I believe it's correct in console. Is there something I have done wrong?
> 
> Yup, error messages are all wrong. Sometimes you don't print it at all, sometimes you do WrtieLine where Write is needed, etc. Be more careful the way you do it.

You can see all the reasons why it failed if you click the checks link.

Almantask:
Expired

PR #441, opened 2020-06-04
Chapter3/homework/4

DavidKatona:
I think I understand the use of generics and covariance, though I cannot fully wrap my head around contravariance.

From your lessons, I've understood that "in" -> contravariance, is used to go from something more general to something more specific, but I'm having a hard time visualizing the practical use of this.

Nevertheless, here's what I've done for this homework.


Src/BootCamp.Chapter/Schools/ISchool.cs
```diff
@@ -0,0 +1,23 @@
...
+        string Location { get; }
+        List<TStudent> StudentList { get; }
+
+        TStudent this[int index]
+        {
+            get;
+            set;
+        }
+
+        void RegisterStudent(TStudent student);
```
DavidKatona, line 19: Maybe i'm completely off here, but since we're using TStudent as input here, shouldn't ISchool look something like this?

> ISchool '<'in TStudent'>' (the quotes are there because markdown hid the < signs)

 ?

Though, since i'm also using TStudent as something that's returned, it doesn't work as I get errors upon compilation, hence i've left the variance modifiers out of this interface.

Almantask:
> I think I understand the use of generics and covariance, though I cannot fully wrap my head around contravariance.
> 
> From your lessons, I've understood that "in" -> contravariance, is used to go from something more general to something more specific, but I'm having a hard time visualizing the practical use of this.
> 
> Nevertheless, here's what I've done for this homework.

Refer to this example: https://github.com/csinn/CSharp-From-Zero-To-Hero/blob/Chapter3/Lesson/Generics/Src/BootCamp.Chapter/Demos/Contravariance/ContravarianceDemo.cs

Janitor is Employee
Programmer is Employee
IManager<TEmplyee> where TEmployee : Employee
Programmer can have its' own manager, Janitor can have its' own manager.

Imagine if it worked logically (like covariance).
```
IManager<Employee>[] = {janitorManager, programmerManager}
```
However let's try to consume it:
```
janitorManager.Manager(new Programmer());
```
Huh? Janitors aren't programmers, this shouldn't be allowed! But if it worked like covariance, it would be legal. Therefore, you need to specify which managers can handle Programmers or more generalized forms.
So:
```
IManager<Programmer>[] = {programmerManager, generalEmployeeManager}
```
Programmer manager can handle Programmers, because it's made for it.
GeneralEmployeeManager can handle Programmers, also Janitors and any other employee type.

DavidKatona:
This:

Imagine if it worked logically (like covariance).
```
IManager<Employee>[] = {janitorManager, programmerManager}
```
However let's try to consume it:
```
janitorManager.Manager(new Programmer());
```

So you're saying that with contravariance, I can define what kind of specific form a generic paramater can be consumed in a given context? So in:

`interface IManager<in TEmployee> where TEmployee : Employee`

Here, we create a generic IManager interface with TEmployee specified as contravariant. Then all the other methods which have been declared in this interface will accept TEmployee as input.

So for example:
`IManager<Janitor> janitorsManager` -> This supports casting Employee to a more specific type
Then I do something like:
`janitorsManager.Manage(new Janitor());`

Using a more "generalised" form in the above example will result in a compile time error right?
So I won't be able to do something like:
`janitorsManager.Manage(new Employee());`

Covariance would allow this but since we've used the contravariant modifier in the original interface, we cannot do something like this, like in polymorphism right?

Also, If our IManager interface would look something like this:
`interface IManager<out TJanitor> where TJanitor : Janitor`
I could write something like this:
`IManager<Employee> janitorsManager`
Right? This should be legal since we've used the covariant form and Employee is a more "generalised" form of Janitor.

Src/BootCamp.Chapter/Schools/ISchool.cs
Almantask: Yes, the problem is that T here is both input and output, so no modifier is needed.

Src/BootCamp.Chapter/Schools/University.cs
```diff
@@ -0,0 +1,9 @@
+﻿using BootCamp.Chapter.Students;
+
+namespace BootCamp.Chapter.Schools
+{
+    public class University<TStudent> : School<TStudent> where TStudent : IStudent
```
Almantask: Can specify that this accepts only UniversityStudents.

Src/BootCamp.Chapter/Simulation.cs
```diff
@@ -0,0 +1,47 @@
...
+using BootCamp.Chapter.Teachers;
+using System;
+
+namespace BootCamp.Chapter
+{
+    public class Simulation
+    {
+        public static void Run()
+        {
+            ISchool<UniversityStudent> university = new University<UniversityStudent>("Massachusetts Institute of Technology", "Cambridge, Massachusetts, United States");
```
Almantask, line 13: University shouldn't accept middleSchoolStudent, but as of now it can accept them.

Almantask (changes requested):
One last bit to fix: add constraints for specific schools that they only accept-school specific students. It should not be provided during school creation, but should always be known (class level).

Src/BootCamp.Chapter/Schools/University.cs
DavidKatona: Fixed!

Src/BootCamp.Chapter/Simulation.cs
DavidKatona: Changed so schools only accept their own type of students.

DavidKatona:
I think I get it now. I spent a huge amount of time playing with generic interfaces a couple of days ago in realtion to the next homework, and I used them there as well.

Spending hours of grinding them was definitely a plus! :)

PR #444, opened 2020-06-10
Chapter2/homework/4

matthew-hole:
I hadn't done the homework 1And2 so just created a quick 'Calculate' class to validate the implementation.

Hopefully the interfaces are done correctly. I did initially attempt at having the 'Calculate' class implement the ILogger interface directly but it didnt seem clean - guidance on this would be great.

### DIFF
--- a/Src/BootCamp.Chapter/Calculate.cs
+++ b/Src/BootCamp.Chapter/Calculate.cs
@@ -1,30 +1,91 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 
 namespace BootCamp.Chapter
 {
     class Calculate
     {
-        public double Add(params int[] numbers)
-        {
-            var sum = 0;
-            foreach(var x in numbers)
-            {
-                sum += x;
-            }
+        private ILogger _logger;
 
-            return sum;
+        public Calculate()
+        {
+            _logger = InitialiseLogger();
+            _logger.ActionLog("Calculation Program Started");
+        }
+        public double Add(params double[] numbers)
+        {
+            try
+            {
+                var sum = 0.0;
+                foreach (var x in numbers)
+                {
+                    if(x.GetType() == typeof(double))
+                    {
+                        sum += x;
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"{x} is not a valid value");
+                    }
+                    
+                }
+                _logger.ActionLog($"{string.Join(" + ", numbers)} = {sum}");
+                return sum;
+            }
+            catch(Exception ex)
+            {
+                _logger.ErrorLog(ex);
+                return 0;
+            }
         }
 
-        public double Subtract(params int[] numbers)
+        public double Subtract(params double[] numbers)
         {
-            var sum = numbers[0];
-            for(int i = 1; i < numbers.Length; i++ )
+            try
             {
-                sum -= numbers[i];
+                var sum = numbers[0];
+                for (int i = 1; i < numbers.Length; i++)
+                {
+                    if (numbers[i].GetType() == typeof(double))
+                    {
+                        sum -= numbers[i];
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"{numbers[i]} is not a valid value");
+                    }
+                }
+                _logger.ActionLog($"{string.Join(" - ", numbers)} = {sum}");
+                return sum;
             }
-            return sum;
+            catch(Exception ex)
+            {
+                _logger.ErrorLog(ex);
+                return 0;
+            }
+        }
+
+        private static ILogger InitialiseLogger()
+        {
+            while (true)
+            {
+                Console.WriteLine("Would you like to create a Log File (1) or Log in Console (2) to track activity?");
+                var option = Int32.Parse(Console.ReadLine());
+
+                switch (option)
+                {
+                    case 1:
+                        return new FileLogger();
+
+                    case 2:
+                        return new ConsoleLogger();
+
+                    default:
+                        break;
+                }
+            }
         }
     }
 }
--- a/Src/BootCamp.Chapter/Calculate.cs
+++ b/Src/BootCamp.Chapter/Calculate.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Calculate
+    {
+        public double Add(params int[] numbers)
+        {
+            var sum = 0;
+            foreach(var x in numbers)
+            {
+                sum += x;
+            }
+
+            return sum;
+        }
+
+        public double Subtract(params int[] numbers)
+        {
+            var sum = numbers[0];
+            for(int i = 1; i < numbers.Length; i++ )
+            {
+                sum -= numbers[i];
+            }
+            return sum;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Calculate
```
Almantask, line 8: Horrible name.
1) should be a noun
2) should logically named. Even demo would be better. Maybe Calculator? Etc...

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        private ILogger _logger;
+
+        public Calculate()
+        {
+            _logger = InitialiseLogger();
```
Almantask, line 14: Wrong. You are missing a point, you should be using the implementation of logger across the whole application consistently, now you just hardcoded it.

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+            _logger.ActionLog("Calculation Program Started");
+        }
+        public double Add(params double[] numbers)
+        {
+            try
```
Almantask, line 19: Your try catch obscured the rest of the program. Please remove it.

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                var sum = 0.0;
+                foreach (var x in numbers)
+                {
+                    if(x.GetType() == typeof(double))
```
Almantask, line 24: This is also wrong. You have an array of doubles. Every individual number is bound to be double

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+                    {
+                        sum += x;
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"{x} is not a valid value");
```
Almantask, line 30: This can be removed, it gets never hit.

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+                return sum;
+            }
+            catch(Exception ex)
+            {
+                _logger.ErrorLog(ex);
+                return 0;
+            }
+        }
+
+        public double Subtract(params double[] numbers)
```
Almantask, line 44: This is copy-pasta. You could have reused the previous function.

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+                return sum;
+            }
+            catch(Exception ex)
+            {
+                _logger.ErrorLog(ex);
+                return 0;
+            }
+        }
+
+        private static ILogger InitialiseLogger()
```
Almantask, line 70: This part is good.

Src/BootCamp.Chapter/Calculate.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            while (true)
+            {
+                Console.WriteLine("Would you like to create a Log File (1) or Log in Console (2) to track activity?");
+                var option = Int32.Parse(Console.ReadLine());
+
+                switch (option)
+                {
+                    case 1:
```
Almantask, line 79: ..though you should prefer to use enums over magic numbers. Enums or constants, but not magic numbers. Really hard to understand what does it mean.

Src/BootCamp.Chapter/ConsoleLogger.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class ConsoleLogger : ILogger
```
Almantask, line 7: This is fine.

Src/BootCamp.Chapter/FileLogger.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class FileLogger : ILogger
```
Almantask, line 8: This is fine.

Src/BootCamp.Chapter/ILogger.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    interface ILogger
```
Almantask, line 7: Well done for putting comments.

Src/BootCamp.Chapter/ILogger.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Used to log the Action at the current Date and Time
+        /// </summary>
+        /// <param name="action"></param>
+        void ActionLog(string action);
```
Almantask, line 13: If you simply swap Log and Action, you get LogAction. However, it should be simplified. It should be LogAction(string message). Action is too ambigous.

Src/BootCamp.Chapter/ILogger.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        /// <summary>
+        /// Used to log any errors or exceptions raised at the current Date and Time
+        /// </summary>
+        /// <param name="error"></param>
+        void ErrorLog(Exception error);
```
Almantask, line 19: If you rename it to LogError, then all good. However, you probably expect an exception as input, because error = exception?

Almantask (changes requested):

Src/BootCamp.Chapter/Calculate.cs
matthew-hole: Agreed - change to be made.

Src/BootCamp.Chapter/ILogger.cs
matthew-hole: Agreed.

Src/BootCamp.Chapter/ILogger.cs
matthew-hole: Agreed.

Src/BootCamp.Chapter/Calculate.cs
matthew-hole: Agreed.

Src/BootCamp.Chapter/Calculate.cs
matthew-hole: Agreed

Src/BootCamp.Chapter/Calculate.cs
matthew-hole: Agreed

Src/BootCamp.Chapter/Calculate.cs
matthew-hole: Agreed, the entire application should utilise the same 'logger' everything the application to run. Initialisation of logger is now generic and instantiated within the Program.cs file.

PR #445, opened 2020-06-10
Chapter4/homework/3

ghost:
make a lot of small tests so I can solve this one the TDD way


Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+
+public class Solution
+{
+    public int MyAtoi(string str)
```
Almantask, line 5: 😞 Why is there no task of what you're doing? How am I supposed to know. What is Atoi, what are you even trying to do...?

Almantask: Please learn how to name things properly. MyAtoi is a meaningless name to me.
It's the 100th time I repeat myself to name methods as VERBS and not NOUNS!

Almantask: No commnents, no nothing and I am left clueless.

Almantask (changes requested):
Please have common sense to add descirption of a problem you're solving.

Almantask:
> make a lot of small tests so I can solve this one the TDD way

This is not how we describe a PR. I don't really care the fact that you wrote tests and especially not about the fact that it's in TDD way.
I care about what actual problem you were trying to solve. Which is not described here.

Please don't be lazy and try to think about the reviewer who is going to check code. You, as a reviewer, would not be able to understand such PR. It's just code for hell knows what...

ghost:
im sorry
here the whole description from the leetcode site 

Implement atoi which converts a string to an integer.

The function first discards as many whitespace characters as necessary until the first non-whitespace character is found. Then, starting from this character, takes an optional initial plus or minus sign followed by as many numerical digits as possible, and interprets them as a numerical value.

The string can contain additional characters after those that form the integral number, which are ignored and have no effect on the behavior of this function.

If the first sequence of non-whitespace characters in str is not a valid integral number, or if no such sequence exists because either str is empty or it contains only whitespace characters, no conversion is performed.

If no valid conversion could be performed, a zero value is returned.

Note:

    Only the space character ' ' is considered as whitespace character.
    Assume we are dealing with an environment which could only store integers within the 32-bit signed integer range: [−231,  231 − 1]. If the numerical value is out of the range of representable values, INT_MAX (231 − 1) or INT_MIN (−231) is returned.

Example 1:

Input: "42"
Output: 42

Example 2:

Input: "   -42"
Output: -42
Explanation: The first non-whitespace character is '-', which is the minus sign.
             Then take as many numerical digits as possible, which gets 42.

Example 3:

Input: "4193 with words"
Output: 4193
Explanation: Conversion stops at digit '3' as the next character is not a numerical digit.

Example 4:

Input: "words and 987"
Output: 0
Explanation: The first non-whitespace character is 'w', which is not a numerical 
             digit or a +/- sign. Therefore no valid conversion could be performed.

Example 5:

Input: "-91283472332"
Output: -2147483648
Explanation: The number "-91283472332" is out of the range of a 32-bit signed integer.
             Thefore INT_MIN (−231) is returned.

Almantask:
> im sorry
> here the whole description from the leetcode site
> 
> Implement atoi which converts a string to an integer.
> 
> The function first discards as many whitespace characters as necessary until the first non-whitespace character is found. Then, starting from this character, takes an optional initial plus or minus sign followed by as many numerical digits as possible, and interprets them as a numerical value.
> 
> The string can contain additional characters after those that form the integral number, which are ignored and have no effect on the behavior of this function.
> 
> If the first sequence of non-whitespace characters in str is not a valid integral number, or if no such sequence exists because either str is empty or it contains only whitespace characters, no conversion is performed.
> 
> If no valid conversion could be performed, a zero value is returned.
> 
> Note:
> 
> ```
> Only the space character ' ' is considered as whitespace character.
> Assume we are dealing with an environment which could only store integers within the 32-bit signed integer range: [−231,  231 − 1]. If the numerical value is out of the range of representable values, INT_MAX (231 − 1) or INT_MIN (−231) is returned.
> ```
> 
> Example 1:
> 
> Input: "42"
> Output: 42
> 
> Example 2:
> 
> Input: " -42"
> Output: -42
> Explanation: The first non-whitespace character is '-', which is the minus sign.
> Then take as many numerical digits as possible, which gets 42.
> 
> Example 3:
> 
> Input: "4193 with words"
> Output: 4193
> Explanation: Conversion stops at digit '3' as the next character is not a numerical digit.
> 
> Example 4:
> 
> Input: "words and 987"
> Output: 0
> Explanation: The first non-whitespace character is 'w', which is not a numerical
> digit or a +/- sign. Therefore no valid conversion could be performed.
> 
> Example 5:
> 
> Input: "-91283472332"
> Output: -2147483648
> Explanation: The number "-91283472332" is out of the range of a 32-bit signed integer.
> Thefore INT_MIN (−231) is returned.

should be in PR description an not a comment...

Almantask:
So Atoi is `ToSpecialInteger(string a)`

ghost:
Could be, I took the name that the challenge is given so I could easily check my code

Almantask:
> Could be, I took the name that the challenge is given so I could easily check my code

And how is an outside person supposed to know that the method name is from a challenge?

ghost:
he/she could not or I have to mentioned that,

Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        int result = 0;
+        int sign = 1;
+
+        if (!string.IsNullOrEmpty(str) && str.Length > 0)
```
Almantask, line 10: Instead of two conditions just use IsNullOrWhitespace

Almantask: To reduce nesting, invert if statement and return.

Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            char[] chars = str.ToCharArray();
```
Almantask, line 12: And be simplified:
step 1: remove all whitespaces from string. str.Replace(" ", "")

Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
...
+                        break;
+                    }
+                    else
+                    {
+                        continue;
+                    }
+                }
+                else if (chars[i] == '-' || chars[i] == '+')
+                {
+                    if (i + 1 < chars.Length && IsDigit(chars[i + 1]) &&
```
Almantask, line 31: `i + 1 < chars.Length`- why are you doing this? this is checked inside a loop.

Src/BootCamp.Chapter/Solution.cs
Almantask, line 5: This code is extremely hard to read. We learned in chapter 1 that we should split big functions into smaller ones, where did all that go?

Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        if (!string.IsNullOrEmpty(str) && str.Length > 0)
+        {
+            char[] chars = str.ToCharArray();
+
+            double num = 0;
+
+            for (int i = 0; i < chars.Length; i++)
+            {
+                if (chars[i] == ' ')
```
Almantask, line 18: This is not needed if you strip the string from whitespaces.

Src/BootCamp.Chapter/Solution.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    if (i > 0 && chars[i - 1] >= '0' && chars[i - 1] <= '9')
```
Almantask, line 20: Makes no sense.

Almantask, line 20: You have IsDigit function but you suddenly don't use it here. I don't get it.

Almantask (changes requested):
Why are you jumping in the loop back and forth (i+1, i-1). Just work with one character at a time. One of your worst written code. Please rewrite cleaner.

Break down the problem:
```
1) Strip whitespace
2) Get multiplier (-1 or 1)
3) Get digits until you hit a non-digit
4) Evaluate result
```

Almantask:
Expired

PR #450, opened 2020-06-15
Chapter1/homework/5

arthur-schevey:
**Array Operations Homework:**

Sort operation was completed using the simple bubble sort algorithm I learned from [this YouTube video.](https://www.youtube.com/watch?v=xli_FI7CuzA)

The reverse operation works by comparing the beginning and end of an array and swapping the two values. It then shifts the beginning index up by one and the end index down by one and swaps those two new values. It continues during a while loop until the start and end index surpasses each other (so values don't get swapped back).

To save time/work I called `RemoveAt()` in both `RemoveFirst()` and `RemoveLast()`.
I also did this by calling `InsertAt()` in both `InsertFirst()` and `InsertLast()`.


Almantask:
Merging to wrong branch (master). You should be merging to the same hw branch.

Almantask (approved):
Yup, what you did is exactly what we wanted 😄

PR #451, opened 2020-06-15
Chapter3/homework/5

DavidKatona:
Here's the solution for the exercises. I tried reusing knowledge we've learned throughout this chapter (generics, etc.).

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using BootCamp.Chapter.Renderer;
+using System;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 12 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,15 @@
...
+﻿using BootCamp.Chapter.Demos;
+using System;
 
 namespace BootCamp.Chapter
 {
     class Program
     {
         static void Main(string[] args)
         {
-
```
DavidKatona, line 9: These are really lazy demos to be honest :D

Instead of spending time on creating an interactive menu like we did in previous exercises, I tried focusing on using high level language features such as generics, contravariant generic interfaces, and other sugar.

Almantask:
Looks good

PR #452, opened 2020-06-15
Completed homework 1 and 2.

itsjustkevin:


Almantask (approved):
Looks good but next time be more mindful to the requirements: second person is missing.

PR #454, opened 2020-06-20
Chapter1/homework/4

mbmaeso:
Finished homework 4 of chapter 1. All tests passed.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -55,13 +55,13 @@ namespace BootCamp.Chapter
             {
                 return number;
             }
-            else if (input.Length == 0)
+            else if (input == null || input?.Length == 0)
             {
                 return 0;
             }
             else
             {
-                Console.WriteLine($"\"{input}\" is not a valid number.");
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +11,118 @@ class Program
     {
         static void Main(string[] args)
         {
+            // Demo();
```
Almantask, line 14: Call actual demo please.

Almantask: Twice :)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +11,118 @@ class Program
...
+
+            Console.Write(CalculateBMI(-100, 50));
+        }
+
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
```
Almantask, line 21: Loop is fine, but since you need to call it twice, it would be slightly more easy to
```cs
public static void Demo()
{
  ProcessPersonBmi();
  ProcessPersonBmi();
}
```

Almantask (approved):
Looks good overal.

PR #455, opened 2020-06-21
Jude - Chapter 1 Lesson 3 Homework Completed

JudeRV:
Completed Chapter 1 Lesson 3 (functions) homework. All tests in `Lesson3Tests.cs` pass correctly.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -20,27 +20,27 @@ namespace BootCamp.Chapter
 
         public static string GetName()
         {
-            Console.Write("Please enter name: ");
+            Console.Write("Testing");
             return Console.ReadLine();
         }
         public static int GetAge()
         {
-            Console.Write("Please enter age: ");
+            Console.Write("Testing");
             return int.Parse(Console.ReadLine());
         }
         public static float GetWeight()
         {
-            Console.Write("Please enter weight (in kg): ");
+            Console.Write("Testing");
             return float.Parse(Console.ReadLine());
         }
         public static float GetHeight()
         {
-            Console.Write("Please enter height (in cm): ");
+            Console.Write("Testing");
             return float.Parse(Console.ReadLine());
         }
         public static float GetBMI(float weight, float height)
         {
-            return weight / ((height / 100) * (height / 100));
+            return weight / (height * height);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        static string name = GetName();
+        static int age = GetAge();
+        static float weight = GetWeight();
+        static float height = GetHeight();
+        static float bmi = GetBMI(weight, height);
```
Almantask: CalculateBmi is a better name. Getting refers to retrieving something that is complete, not computing.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        static string name = GetName();
```
Almantask: GetX is a fine name, but you don't just get a name, you also request for it. PromptX sounds better, because it refers to requesting in console and getting it.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,46 @@
...
+        static int age = GetAge();
+        static float weight = GetWeight();
+        static float height = GetHeight();
+        static float bmi = GetBMI(weight, height);
+
+        public static void Demo()
+        {
+            Console.WriteLine(name + " is " + age + " years old, their weight is " + weight + " kg, and their height is " + height + " cm.");
```
Almantask, line 17: Getting input and calculating it should be done in demo function and not the class level. Especially since you also need to do the same for person2. Right now, you would loose bmi of person1.

Almantask (changes requested):
Would look okay, but logic should be in function, not class level. It being in a class means more default value rather than actual demo/calulation.

Src/BootCamp.Chapter/Lesson3.cs
JudeRV: That's a very good point, I hadn't even considered that. Alright, duly noted and code fixed.

Almantask (approved):
Looks almost perfect, except that in Demo function you could call 
```
ProcessBmi();
ProcessBmi();
```
(twice, as requested).

PR #459, opened 2020-06-27
Created Encrypt and Decrypt while using Caesar cipher

Laura582:


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -7,16 +7,37 @@ namespace BootCamp.Chapter
...
+                Console.WriteLine(messageEncrypt);
+            }
+
+            return new string(encryptString.ToArray());
+
+        }
         public static string Decrypt(string message, byte shift)
         {
-            return message;
+            return EncryptString(message, 0 - shift);
```
Almantask, line 40: The fact that you spotted that you can reuse EncryptString puts you ahead of many. Well done.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -7,16 +7,37 @@ namespace BootCamp.Chapter
...
+            if (message == null)
+            {
+                return null;
+            }
+
+            List<char> encryptString = new List<char>();
+
+            for (int j = 0; j < message.Length; j++)
+            {
+                var messageToInt = (int)message[j];
```
Almantask, line 28: You could be more clear with the way you name variable. It's not a message to int, it's a character to int. Character to int is asciiCode.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -7,16 +7,37 @@ namespace BootCamp.Chapter
...
+                var messageToIntAddShift = (messageToInt + shift);
```
Almantask, line 29: simplify: 
```
var shiftedChar = (char)(messageToint + shift)
```

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -7,16 +7,37 @@ namespace BootCamp.Chapter
...
+                var messageEncrypt = (char)messageToIntAddShift;
+                encryptString.Add(messageEncrypt);
+                Console.WriteLine(messageEncrypt);
```
Almantask, line 32: This is not good, because there was no requirement for it. Maybe it was for you for debugging, but in actual code it shouldn't be there.

Almantask: It shouldn't be there, because it's a side effect of a function and this function now does 2 things.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -7,16 +7,37 @@ namespace BootCamp.Chapter
...
-    public static class CaesarCipher
+    public class CaesarCipher
     {
         public static string Encrypt(string message, byte shift)
         {
-            return message;
+            return EncryptString(message, shift);
         }
 
+        private static string EncryptString(string message, int shift)
```
Almantask, line 17: Put private methods below public. Because private is a detail you might not care about. Public gives us a clear understanding of what we can do with the class, private gives a better understanding how does it work under the hood.

Almantask (approved):

Almantask:
I think you did good. Carry on, I recommend doing Chapter 2, homework 1 next.

PR #460, opened 2020-06-27
Homework 3 Initial Submission

LarryBoxler:
Homework 3 submission - all check test passed

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,7 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Text;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Globalization;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            string firstName = InputString("Enter First Name: ");
```
Almantask: No need a blank line between every statement. Use a blank line to separate unrelated blocks of code. For example: Reading is one thing, calculating is another, printing final result is third. (so separate each by 3 blanks lines in total).

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            float personsHeight = InputFloat("Enter Height in m:");
+
+            Console.WriteLine(firstName + " " + lastName + " is " + personsAge + " years old, their weight is "
+                                + personsWeight + " and their height is " + personsHeight + " m.");
+
+            Console.WriteLine("Their BMI is " + string.Format("{0:0.#}", CalculateBMI(personsWeight, personsHeight)));
+
+            Console.WriteLine();
+            Console.WriteLine("Please Enter a second persons information");
```
Almantask: instead of copy-pasting code from second person- reuse it via a function. Put the way you work with first person to a function and only change the person prompt (which you will have to pass from an input).

Almantask (changes requested):

Src/BootCamp.Chapter/Lesson3.cs
LarryBoxler: Good point - thanks for this kind of feedback.  I spent most of my programming writing code for myself so these kind of points help me to make my code more readable for someone who might take it on later.

Src/BootCamp.Chapter/Lesson3.cs
LarryBoxler: Yes - missed that makes sense to have that as a function.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            //Enter information for 2 people
```
Almantask, line 10: Instead of doing a comment like this, why don't you just:
```
public static void Demo()
{
  PromptPerson(1);
  PromptPerson(2);
  CloseProgram();
}
```
```
private static void PromptPerson(int index)
{
            Console.WriteLine("Please enter the information for person "+ index);
            NewPerson();
}
```

Almantask (changes requested):

Src/BootCamp.Chapter/Lesson3.cs
Almantask: Code should speak for itself 😄

LarryBoxler: Ok yes - understood !  👌

PR #461, opened 2020-06-28
Chapter1/homework/1 and2

Laura582:
I have created a program which is able to calculate BMI from input data

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,24 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            Console.WriteLine("Enter Name");
+            string name = Console.ReadLine();
+            Console.WriteLine("Enter age ");
+            string age = Console.ReadLine();
+            int.Parse(age);
+            Console.WriteLine("Enter weight ");
+            string weight = Console.ReadLine();
+            double weightInDouble = double.Parse(weight);
+            Console.WriteLine("Enter height in cm ");
+            string height = Console.ReadLine();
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,30 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            Person();
```
Almantask: A function name should be a verb, not a noun, because a function does something. It is not something.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,30 @@ class Program
...
+            Person();
+        }
+
+        static void Person()
+        {
+            Console.WriteLine("Enter name");
```
Almantask, line 14: I call such code a blob. Try to split code into logical sections, using an empty line. For example: releated to reading name, related to reading age, printing, etc.

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,30 +6,53 @@ class Program
...
+            double heightInCentemeters = double.Parse(height);
+            return heightInCentemeters;
+        }
+        static double CalculateBMI(double weight, double height)
+        {
+            double heightInMeters = height / 100.0;
+            double BMI = weight / (heightInMeters * heightInMeters);
+            return BMI;
+        }
+        static void PrintPersonInformation()
```
Almantask, line 45: As mentioned before- higher level methods should be above lower level. The details (how it works), should be under the method.

Almantask (approved):

PR #463, opened 2020-06-29
Chapter1/homework/3

Laura582:
- I have added a new file "Lesson3.cs";
- I have reduced code duplication;


Almantask (approved):
Looks good.

Just a small nit: instead of `Read` use prompt. It doesnt just read, it also asks to input.

PR #464, opened 2020-06-29
Chapter3/homework/6

DavidKatona:
Took me quite a while to put this together. Also the CSV reader has a very lazy implementation and could be rewritten in a much better format.


Src/BootCamp.Chapter/Person.cs
```diff
@@ -1,7 +1,75 @@
...
+        }
+
+        public static bool TryParse(string input, out Person person)
+        {
+            person = default;
+
+            if (string.IsNullOrWhiteSpace(input)) return false;
+
+            var personalData = input.Split(',');
+
```
DavidKatona, line 61: This looks bad in my opinion but I didn't want to spend too much time on just parsing the data.

Src/BootCamp.Chapter/Demo/DemoDelegatesAndEvents.cs
```diff
@@ -0,0 +1,55 @@
+﻿using BootCamp.Chapter.Events;
+using System;
+using System.Transactions;
+
+namespace BootCamp.Chapter.Demo
+{
+    public static class DemoDelegatesAndEvents
+    {
+        private static event EventHandler<ApplicationStateEventArgs> ApplicationStateHandler;
```
Almantask: It's not a handler, is an event. OnApplicationStarted for example. You will call this event, all those that subscribed to it will be hit.

Src/BootCamp.Chapter/Demo/DemoDelegatesAndEvents.cs
```diff
@@ -0,0 +1,55 @@
...
+        private static event EventHandler<InputPressedEventArgs> InputHandler;
+
+        public static void Run()
+        {
+            // Subscribe for events so we can catch input presses and application state changes
+            ApplicationStateHandler += ApplicationStateEvents.OnApplicationStateChange;
```
Almantask: Normally, it should be subscribed to from the outside. If it's nested like that, it could simply be a normal method.

Almantask (changes requested):
Events part could be improved.

DavidKatona:
I did some heavy refactoring on the codebase. Events should be much better now. I've left some comments in the codebase because they're helpful to me when I'll review this homework later.

PS: not sure why that codacy check keeps failing.

Src/BootCamp.Chapter/Demo/DemoDelegatesAndEvents.cs
DavidKatona: Indeed. This should be fixed now.

Src/BootCamp.Chapter/Demo/DemoApplication.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
+using BootCamp.Chapter.ApplicationState;
+using BootCamp.Chapter.UserInput;
+
+namespace BootCamp.Chapter.Demo
+{
+    public class DemoApplication
+    {
+        public IApplicationStateController ApplicationStateController { get; }
```
Almantask: Why don't you just put `ApplicationStarted` event here?

Almantask: I'd remove the state controller.

Almantask (changes requested):

Src/BootCamp.Chapter/Demo/DemoApplication.cs
DavidKatona: I've removed the ApplicationStateController class as a whole. It makes more sense for the DemoApplication as an object to manage/monitor its internal state.

For the second point though, I have decided to avoid adding events to the ctor. I wanted to work with a single ApplicationStateChanged event instead of using multiple. If I'd want to fire the event inside the ctor, I'd have to make it static and initialize it before any objects are instantiated from the outside. Otherwise, if I'd keep them as non-static members, I'd have to subscribe to them inside the ctor, otherwise, they'd be just null upon initialization.

Maybe I got that one wrong there but that's what worked out for me. :)

Almantask (approved):
Looks good now. Normally, for simplicity sake you will see specific events (because in your case each state is just a different event) which enable you subscribing to only state changes you care about. Maybe I want to do something when application opened, nothing when application closed. But now you are forcing me to care about that in the handler (to also switch on state).

DavidKatona:
> Looks good now. Normally, for simplicity sake you will see specific events (because in your case each state is just a different event) which enable you subscribing to only state changes you care about. Maybe I want to do something when application opened, nothing when application closed. But now you are forcing me to care about that in the handler (to also switch on state).

That's a really good point. I really tried to reduce the amount of code I write; hence this "reusable" event but I understand your point. When I worked with some WinForms apps in the past, I remember them having all kinds of specific events for App started, closed, minimized, etc. I'll keep this in mind when I'll work with events in the future.

Almantask:
> > Looks good now. Normally, for simplicity sake you will see specific events (because in your case each state is just a different event) which enable you subscribing to only state changes you care about. Maybe I want to do something when application opened, nothing when application closed. But now you are forcing me to care about that in the handler (to also switch on state).
> 
> That's a really good point. I really tried to reduce the amount of code I write; hence this "reusable" event but I understand your point. When I worked with some WinForms apps in the past, I remember them having all kinds of specific events for App started, closed, minimized, etc. I'll keep this in mind when I'll work with events in the future.

Having one function that does everything is actually going to result into more code.

PR #466, opened 2020-07-01
Chapter1/homework/4

Laura582:
- Added Logical flow to check if entered values are valid


Almantask (approved):
Looks good, though PrintFullMessage is poorly named, it should be PrintPersonWithBmi.

PR #469, opened 2020-07-02
Application calculates BMI for as many people as desired until the...

daman92:
…"q" key is entered.  Types simplified to use double for all numerics.


Almantask:
Looks good, maybe even an overkill for hw1and2

PR #470, opened 2020-07-03
Chapter3/homework/5

matthew-hole:
This homework took a lot longe than I initially thought. Struggled to understand the requirements and the principle behind having the arrays of type 'bool'.

After that I found it fairly straight forward. Reinforced my understanding of generics (which did make the solution more complex and added to the amount of time taken to complete this task).

All tests pass 100%. 

Cheers!


Almantask (approved):
It looked weird at first, but I cannot really pick on anything. You can split bigger functions into smaller steps, but that's nitpicking. Carry on 😄

PR #472, opened 2020-07-04
Chapter 1 HW3

daman92:
Everything works, but I think I messed up the checks part. Any suggestions are appreciated.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -8,31 +8,11 @@ namespace BootCamp.Chapter
     {
 		public static void Demo()
 		{
-			/*2) Take homework 2 code, place it in the forked branch and refactor it using functions.There should be as little duplicate code as possible (there should be functions for:
-				-Calculating BMI(weight comes in kg, height comes in meters),
-			-Prompt for input and converting it to int(print message for request, read console input and return converted input to int), 
-			-Prompt for input and converting it to string(print message for request, read console input and return input),
-			-Prompt for input and converting it to float(print message for request, read console input and return converted input to float).
-			3) Put all the function you made into the right places in Checks.cs class. Run tests, make sure you pass all the tests
-			4) Put all program.cs logic to Lesson3.cs.Call it from main function. Program class should look like this:
-			*/
-
 			string quit = "";
 
 			while (quit != "q" && quit != "Q")
 			{
-				int age = getUserInt("age");
-				string firstName = getUserString("first");
-				string lastNmae = getUserString("last");
-				float weight = getUserFloat("weight(Kg)");
-				float height = getUserFloat("height(cm)");
-
-
-				Console.WriteLine($"{firstName} {lastNmae} is {age} years old, his weight is {weight} kg and his height is {height} cm.");
-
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public class Lesson3
+    {
+		public static void Demo()
+		{
+			/*2) Take homework 2 code, place it in the forked branch and refactor it using functions.There should be as little duplicate code as possible (there should be functions for:
```
Almantask: Remove such comments from code.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,97 @@
...
+			-Prompt for input and converting it to float(print message for request, read console input and return converted input to float).
+			3) Put all the function you made into the right places in Checks.cs class. Run tests, make sure you pass all the tests
+			4) Put all program.cs logic to Lesson3.cs.Call it from main function. Program class should look like this:
+			*/
+
+			string quit = "";
+
+			while (quit != "q" && quit != "Q")
+			{
+				int age = getUserInt("age");
```
Almantask: Use a function that tells what is being done. You need to process person info, might as well call the contents of the loop like that.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,97 @@
...
+				quit = Console.ReadLine();
+			}
+
+		}
+
+		public static float calculateBMI(float weight, float height)
+		{
+			float heightInMeters = height / 100f;
+			return (weight / (heightInMeters * heightInMeters));
+
```
Almantask, line 27: No need a blank line before ending a function.

Almantask: It's actually slightly more clean putting a blank line before return.

Almantask (changes requested):
Be more careful about how you do things, because this is more than is asked and therefore the tests are failing. You can still make extra effort and make it pass with the same idea that you have now, but don't- you ain't gonna need it.

A+ for effort though 😄

Almantask:
> Everything works, but I think I messed up the checks part. Any suggestions are appreciated.

Tests are failing, because you did more than is asked. In a tests, for this homework, no validation is needed and you just pretend to do a happy path.

Src/BootCamp.Chapter/Lesson3.cs
daman92: refactored per instruction.

Src/BootCamp.Chapter/Lesson3.cs
daman92: removed per instructions. (sorry that was just for me.)

Src/BootCamp.Chapter/Lesson3.cs
daman92: made formatting uniform, per comment.

Almantask:
Looks good

PR #477, opened 2020-07-10
Chapter1/Homework/Lesson3

agakesik:
Finished homework for lesson 3: refactor previous code using funcions

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            Console.WriteLine("Person 1:");
+            PersonInformation();
+
+            // Person 2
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Console.WriteLine("Person 1:");
+            PersonInformation();
+
+            // Person 2
+            Console.WriteLine("Person 2:");
+            PersonInformation();
+        }
+
+        private static void PersonInformation()
```
Almantask, line 19: Methods should be verbs, not nouns. What do we do with Personinformation? We Prompt it, ask it, process it, etc. PromptPersonInformation would be a better name. I would also include the writelines within the function and just pass int personIndex (because the number of a person is the only difference between 1 and 2). 

Comments are not needed as well, because it is clear what you do. Code speaks for itself 😄

Almantask (approved):
Well done 😉

Src/BootCamp.Chapter/Lesson3.cs
agakesik: True, true and true 😅 thank you!

PR #479, opened 2020-07-11
Chapter3/homework/6

matthew-hole:
This for me was a turning point in properly understanding both delegates and events; but also afirming knowledge of generics and interfaces. Also learn about ConsoleKey which was cool :)

I have tried to add comments to help the reader understand why certain methods were used. Any advice on how I can make this better (or what some best practices are) that would be great!

### DIFF
--- a/Src/BootCamp.Chapter/ApplicationController.cs
+++ b/Src/BootCamp.Chapter/ApplicationController.cs
@@ -1,0 +1,150 @@
+﻿using System;
+using System.Reflection;
+
+namespace BootCamp.Chapter
+{
+    class ApplicationController
+    {
+        // implement into a method that invokes this event and has the termination logic tirggered?
+        public event EventHandler<ApplicationModeEventArgs> ApplicationModeChanged;
+        public IInputMonitor<ConsoleKey> InputMonitor { get; }
+
+        private readonly Filters Filters = new Filters();
+
+        // Passing the instatiated IInputMonitor object such that the correct event is called in method InitializeApplication()
+        public ApplicationController(IInputMonitor<ConsoleKey> inputMonitor)
+        {
+            InputMonitor = inputMonitor;
+        }
+
+        /// <summary>
+        /// Changes mode to APPLICATION_STARTED: Starts the application and continually checks for User Keyboard Inputs at Main Menu
+        /// </summary>
+        private void InitializeApplication()
+        {
+            ChangeMode(ApplicationModes.APPLICATION_STARTED);
+            while (true)
+            {
+                InputMonitor.MonitorInput(MainMenuInput());
+            }       
+        }
+
+        /// <summary>
+        /// Chnages mode to DEMO_STARTED: Runs the pre-created Demo
+        /// </summary>
+        private void StartDemo()
+        {
+            ChangeMode(ApplicationModes.DEMO_STARTED);
+            Demo demo = new Demo();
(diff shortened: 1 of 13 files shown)


matthew-hole:
There are also instances within the program where I should of added null checks (person information, before Predicates etc etc) which I can add if deemed necessary. However,  I do believe this solution satisfies the requirements.

Src/BootCamp.Chapter/ApplicationController.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+using System.Reflection;
+
+namespace BootCamp.Chapter
+{
+    class ApplicationController
+    {
+        // implement into a method that invokes this event and has the termination logic tirggered?
+        public event EventHandler<ApplicationModeEventArgs> ApplicationModeChanged;
```
Almantask, line 9: Instead of having one event with arguments for each state, have multiple events, without any arguments. That way, you can freely subscribe only to the state of application that you are interested in. Also, you won't have to do an if statement to check what state it is in. Do something when it Start, when it Ends. Not always.

Src/BootCamp.Chapter/InputMonitor.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    class InputMonitor : IInputMonitor<ConsoleKey>
+    {
+        public event EventHandler<InputPressedEventArgs<ConsoleKey>> InputPressed;
+
+        public void MonitorInput(ConsoleKey input)
```
Almantask, line 9: MonitorInput name is redundant. Monitor is enough (because the context and args on top of that explain what you are monitoring.

Src/BootCamp.Chapter/InputMonitor.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            OnInputPressed(input);
+            
+        }
+
+        protected virtual void OnInputPressed(ConsoleKey keyPressed)
```
Almantask, line 15: I think this function is not needed and you can call it in the function above direclty (withuot this function).

Almantask (approved):
Looks very clean. The only pick I have is the fact that ApplicationModeChange event is too big. Normally it's split into very specific events (ApplicationStarted, ApplicationEnded), so that you don't have to switch what mode it is and just handle the event you care about.

Almantask:
> This for me was a turning point in properly understanding both delegates and events; but also afirming knowledge of generics and interfaces. Also learn about ConsoleKey which was cool :)
> 
> I have tried to add comments to help the reader understand why certain methods were used. Any advice on how I can make this better (or what some best practices are) that would be great!

Use less of if statements. Do less in functions. Be more specific.
If you combo all 3, it becomes natural 😄

matthew-hole:
> 
> 
> > This for me was a turning point in properly understanding both delegates and events; but also afirming knowledge of generics and interfaces. Also learn about ConsoleKey which was cool :)
> > I have tried to add comments to help the reader understand why certain methods were used. Any advice on how I can make this better (or what some best practices are) that would be great!
> 
> Use less of if statements. Do less in functions. Be more specific.
> If you combo all 3, it becomes natural 😄

Thank you that review was so quick! Yes that is something I need to work on definitely, I will be sure to be more specific in the next homework! See you then :)

PR #482, opened 2020-07-13
Chapter3/Homework/7And8

matthew-hole:
When creating the custom Linq Extensions for Shuffle, I attempted to make it as generic as I could by accepting any IEnumerable<T>.

This homework was quite fun but also challenging and allowed me to learn more about the inner workings of the language. As always, please point out anyways I can improve - whether its structure of the code of simplification of classes/methods I have created.

Once again, thanks for the great lessons and homeworks!

### DIFF
--- a/Src/BootCamp.Chapter/CollectionWriter.cs
+++ b/Src/BootCamp.Chapter/CollectionWriter.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace BootCamp.Chapter
+{
+    public static class CollectionWriter
+    {
+        public static void WriteCollectionToConsole<T>(IEnumerable<T> collection)
+        {
+            foreach (var element in collection)
+            {
+                Console.WriteLine(element);
+            }
+        }
+    }
+}
--- a/Src/BootCamp.Chapter/LINQExtensions.cs
+++ b/Src/BootCamp.Chapter/LINQExtensions.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BootCamp.Chapter
+{
+    public static class LINQExtensions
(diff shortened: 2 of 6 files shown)


Src/BootCamp.Chapter/CollectionWriter.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace BootCamp.Chapter
+{
+    public static class CollectionWriter
+    {
+        public static void WriteCollectionToConsole<T>(IEnumerable<T> collection)
```
Almantask, line 8: Simplify name: Print()

Src/BootCamp.Chapter/LINQExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public static class LINQExtensions
+    {
+        /// <summary>
+        /// Generic Shuffling Algorithm which will accept a collection of type T and return a new shuffled List<T>.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="collection"></param>
+        /// <returns></returns>
+        public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> collection)
```
Almantask, line 15: This works, a bit inneficient though. A more efficient way would be to randomize a number, then keep shifting by 1 until you hit a number that wsan't used yet.

Almantask (approved):
I think overall you did a good job. Well done 😄

Src/BootCamp.Chapter/LINQExtensions.cs
matthew-hole: Ah okay, see I initially did that but my thinking was that for a large dataset that method would be inefficient, I guess I was wrong haha. Thanks for the tip

PR #483, opened 2020-07-13
Chapter1/homework/4

agakesik:
This was double tricky, first I had to go a little bit back and forth with tests, before I got the expected results, then something weird happened when I try commiting and creating pull request... Bu I hope it works and is good now

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            PromptPersonInformation("1");
+
+            PromptPersonInformation("2");
+        }
+
+        private static void PromptPersonInformation(string person)
+        {
+            Console.WriteLine("Person " + person + ":");
+            string name = PromptStringInput("Please provide your name: ");
+            string surname = PromptStringInput("Please provide your surname: ");
+            int age = PromptIntInput("Please provide your age: ");
+            int weight = PromptIntInput("Please provide your weigth in kg: ");
+            float height = PromptFloatInput("Please provide your height in m: ");
+
+            Console.WriteLine(name + " " + surname + " is " + age + " years old, their weight is " + weight + " kg and their height is " + height + "cm.");
+            float bmi = CalculateBmi(weight, height);
+            Console.WriteLine("Their BMI is: " + bmi);
+        }
+
+        public static float CalculateBmi(float weight, float height)
+        {
+            if (weight <= 0 && height <= 0)
+            {
+                Console.WriteLine("Failed calculating BMI. Reason:");
+                Console.WriteLine("Weight cannot be equal or less than zero, but was " + weight + ".");
+                Console.WriteLine("Height cannot be less than zero, but was " + height + ".");
+                return -1;
+            }
+            else if (weight <= 0)
+            {
+                Console.WriteLine("Failed calculating BMI. Reason:");
+                Console.WriteLine("Weight cannot be equal or less than zero, but was " + weight + ".");
+                return -1;
+            }
+            else if (height <= 0)
+            {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,113 @@
...
+            int age = PromptIntInput("Please provide your age: ");
+            int weight = PromptIntInput("Please provide your weigth in kg: ");
+            float height = PromptFloatInput("Please provide your height in m: ");
+
+            Console.WriteLine(name + " " + surname + " is " + age + " years old, their weight is " + weight + " kg and their height is " + height + "cm.");
+            float bmi = CalculateBmi(weight, height);
+            Console.WriteLine("Their BMI is: " + bmi);
+        }
+
+        public static float CalculateBmi(float weight, float height)
```
Almantask, line 30: Extract pieces of code regarding input validation to places where it belongs (PromptHeight, PromptWeight, etc)

Almantask (changes requested):

Src/BootCamp.Chapter/Lesson3.cs
agakesik: Could you elaborate, please? 
I've been rereading task from hw description and looking through other people's homeworks to see what I should be going for, but I just can't figure out by myself which pieces are out of place

Almantask: What I mean is that calculate bmi doesn't just calculate it, it also validates height, weight and age.

Wouldn't it make sense that you would have functions like PromptAge, PromptHeight, PromptHeight, which would do the validation?

CalculateBmi would just do that one thing and so would the other function would do just one thing.

Almantask: What I mean is that calculate bmi doesn't just calculate it, it also validates height, weight and age.

Wouldn't it make sense that you would have functions like PromptAge, PromptHeight, PromptHeight, which would do the validation?

CalculateBmi would just do that one thing and so would the other function would do just one thing.

agakesik: okay, I think I get it, thanks!

agakesik: Okay, so this is really tricky, because of what specific outputs tests are expecting.
Checks.CalculateMbi is expecting this error messages from validating, so when I move it somewhere I get failed tests... 
Doesn my latest commit (passing all tests) is a step in right direction? Or maybe at this point I shouldn't focus to much about tests and refactor the code to make sense by itself?

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -23,34 +23,53 @@ private static void PromptPersonInformation(string person)
...
+            
+        }
+
+        private static float CalculateBmi(float weight, float height)
+        {
+            float bmi = weight / (height * height);
+            return bmi;
+        }
+
+        private static bool ValidateWeightAndHeight(float weight, float height)
```
Almantask: You should not validate seperately, do it in the same function where you get weight, in another function where you get height.
This function should be removed.

Src/BootCamp.Chapter/Lesson3.cs
Almantask: This homework is hard, but for the wrong reasons (as you mentioned).

And yes, you are in the right direction, this is exactly what I wanted. Now you need to do the crappy part- adapt to make the tests pass.

agakesik: Thank you for your patience, but I think I'll have to pass on this one :v I'm probably too much in my head and just can't stop going around in circles, but I just can't figure out how to pass test without putting error messages into CalculateBmi function. I might come back to it with a clear head somewhere in future, but for now I think I won't go any further.
Nonetheless, thank you for all the help!

Almantask: Would you like to get on a call so I can help you through it? I believe you can do it!

Almantask:
Well done ;)

PR #486, opened 2020-07-15
Program for homework 1 from chapter 1

EnochRobokoff:
Program.cs takes in user input for a person's name, age, weight, and height, and displays this info back to the user. It then calculates the BMI of this person and displays that to the user. It continues to a second person's information.


Almantask (approved):
Looks good.

PR #489, opened 2020-07-16
Chapter1/homework/7

agakesik:


Almantask:
Please target the homework branch and not master.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            string personWithHighestBalanceEver = "not found";
```
Almantask: nit: can be a constant.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
+            double? highestBalance = null;
```
Almantask: Instead of making it nullable, use the smallest possible balance that there can be. For example - double.MaxValue. Like that, every number will be bigger than this one 😄

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
         }
 
         /// <summary>
         /// Return name and loss of a person with a biggest loss (balance change negative).
         /// </summary>
         public static string FindPersonWithBiggestLoss(string[] peopleAndBalances)
         {
-            return "";
+            string personWithBiggestLoss = "not found";
+            double? biggestLoss = null;
```
Almantask: min double instead of nullable.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
+
+            foreach (string personInformation in peopleAndBalances)
+            {
+                string[] personInformationSplit = personInformation.Split(", ");
+
+                for (int i = 1; i <= personInformationSplit.Length - 2; i++)
```
Almantask: You could use a constant to indicate the reason why the loop starts at 1. For example: 
```
const int firstBalanceIndex = 1;
```

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
+                {
+                    bool isPreviousBalanceNumber = double.TryParse(personInformationSplit[i], out double previousBalance);
```
Almantask: this is not previous balance, this is current balance.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
         }
 
         /// <summary>
         /// Return name and current money of the richest person.
         /// </summary>
         public static string FindRichestPerson(string[] peopleAndBalances)
         {
-            return "";
+            string richestPerson = "not found";
+            double? highestCurrentBalance = null;
```
Almantask: use min double instead of null

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +9,128 @@ public static class BalanceStats
...
         }
 
         /// <summary>
         /// Return name and current money of the most poor person.
         /// </summary>
         public static string FindMostPoorPerson(string[] peopleAndBalances)
         {
-            return "";
+            string poorestPerson = "not found";
+            double? lowestCurrentBalance = null;
```
Almantask: use max double instead of null

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +37,60 @@ public static class TextTable
...
+
+            sb.Append(AddBorder(message, padding));
+            sb.Append(AddPaddingLines(message, padding));
+            sb.Append(AddMessageLines(message, padding));
+            sb.Append(AddPaddingLines(message, padding));
+            sb.Append(AddBorder(message, padding));
+
+            return sb.ToString();
+        }
+
```
Almantask, line 50: nit: consider reusing similar parts of code:
whatever is similar should go under a function;
whatever is different should come as an argument.

Almantask (changes requested):
Everything looks quite good, except for finding the min/max. Consider using a very big/small initial value for the first comparison, instead of null.

Src/BootCamp.Chapter/BalanceStats.cs
agakesik: oh, that a cool trick!

Almantask (approved):
Beautiful 😄

PR #492, opened 2020-07-19
Homework 3

Tailor-AI:
Homework 3, using functions.

/Tailor

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Globalization;
+namespace BootCamp.Chapter
+{
+    public class Lesson3
+    {
+        // Had to implement to make test work with float (. instead of , in test number).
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class Lesson3
+    {
+        // Had to implement to make BootCamp.Chapter.Test work with float ("." instead of "," in test number).
+        static CultureInfo invC = CultureInfo.InvariantCulture;
+
+        internal static void Demo()
+        {
+            // Calling the functions and combining output.
+            string name = PrintMessageAndReturnString("Please enter your name: ");
```
Almantask, line 15: In order to understand what your demo does, I need to read through all of those lines.
However, if you extracted the function body, I could simply read a single line (ProcessPersonBmi or similar) and immediately understand what demo does in high level. If I want to understand how ProcessPersonBmi, then I will refer to the detail- the other function.

Almantask: nit: Second person is missing.

Almantask (approved):
Overall looks good enough.

PR #494, opened 2020-07-19
Chapter1/homework/1 and i couldn't find 2

Zacc806:
I tried the try catch errors, but cleared it all, for it is the first homework. I figured that would come later

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Globalization;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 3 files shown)


Almantask:
> I tried the try catch errors, but cleared it all, for it is the first homework. I figured that would come later

First homework is to write code.
Second homework is to just get github setup.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,36 @@
...
+            surtitle = textInfo.ToTitleCase(surtitle); 
+
+            Console.Write("Input Age: ");
+            int age = int.Parse(Console.ReadLine());
+            Console.Write("Input Weight (in kg): ");
+            var weight = double.Parse(Console.ReadLine());
+            Console.Write("Input Height (in cm): ");
+            var height = double.Parse(Console.ReadLine());
+
+            var BMI = weight / ((height*0.01)*(height*0.01));
```
Almantask, line 28: conventional name, even for abreviations, is to use camelCase.

Almantask: so bmi for varialbe name, Bmi for class name, etc.

Almantask (approved):
Looks good 😄

PR #496, opened 2020-07-19
Chapter1/homework/4

LarryBoxler:
Initial submission.  All tests passing.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -1,0 +1,118 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static void Demo()
+        {
+            PromptPerson(1);
+            PromptPerson(2); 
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,118 @@
...
+            int personsAge = InputInt("Enter Age: ");
+            float personsWeight = InputFloat("Enter Weight in kg: ");
+            float personsHeight = InputFloat("Enter Height in m: ");
+
+            //Echo data to console
+            Console.WriteLine(firstName + " " + lastName + " is " + personsAge + " years old, their weight is "
+                                + personsWeight + " and their height is " + personsHeight + " m.");
+
+            //Calculate and Display BMI
+            Console.WriteLine("Their BMI is " + string.Format("{0:0.#}", CalculateBMI(personsWeight, personsHeight)));
```
Almantask, line 98: since you're using string format, I would recommend just using string interpolations.
instead of:
"a = " + a " and b = " + b;
you can just:
$"a = {a} and b = {b}";

Almantask (approved):
Looks great!

PR #499, opened 2020-07-20
Chapter1/homework/9

Laura582:
This was fun! 

- The program converts from decimal to binary and from binary to decimal

- Using WASD (both capital and non capital) keys program prints arrows

### DIFF
--- a/Src/BootCamp.Chapter/BinaryConverter.cs
+++ b/Src/BootCamp.Chapter/BinaryConverter.cs
@@ -8,12 +8,26 @@ namespace BootCamp.Chapter
     {
         public static long ToInteger(string binary)
         {
-            return 0;
+            try
+            {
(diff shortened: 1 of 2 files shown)


Almantask (approved):
Looking clean! 👍

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -8,12 +8,26 @@ public static class BinaryConverter
...
         public static long ToInteger(string binary)
         {
-            return 0;
+            try
+            {
+                if (string.IsNullOrEmpty(binary))
+                {
+                    return 0;
+                }
+                var decimalNumber = Convert.ToInt32(binary, 2);
```
Almantask, line 17: I actually wasn't aware of this. Nailed it!

PR #500, opened 2020-07-20
Chapter1/homework/3

Zacc806:
Reverse engineering the testfile proved to work for me. Sorry, if that wasn't intended


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+
```
Almantask: Missing something here 😄

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            
+        }
+        public static int PromptInt(string message)
+            {
+                Console.Write(message);
+                int intedMessage = int.Parse(Console.ReadLine());
```
Almantask: You're not getting a stringed message- you're getting a number.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+                return intedMessage;
+            }
+            
+        public static string PromptString(string message)
+            {
+                Console.Write(message);
+                var stringedMessage = Console.ReadLine();
```
Almantask: You're not getting a stringed message- you're getting a string.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+                return stringedMessage;
+            }
+            
+        public static float PromptFloat(string message)
+            {
+                Console.Write(message);
+                float floatedMessage = float.Parse(Console.ReadLine());
```
Almantask: You're not getting a stringed message- you're getting a number.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,43 @@
...
+                return floatedMessage;
+            }
+
+            
+        public static float calculateBmi(float W, float H)
```
Almantask: please don't use one character variable names, as it makes code unclear. Also use camelCase and not PascalCase, even for abreviations, when passing variables.

Almantask (changes requested):
Please note to what branch are you trying to merge. You should have aimed at Chapter1/Homework/3 and not master (to the same one you are at, except the original repo)

Src/BootCamp.Chapter/Lesson3.cs
Zacc806: I can't get my head around it. Aren't i getting a message (in string) for requesting to enter a number and then parsing the input into int?

Almantask: Console.ReadLine()- this is getting the user input
int.Parse() - this is turned that to an int
The result?
You have a number.

Zacc806: image

Zacc806: > Console.ReadLine()- this is getting the user input
> int.Parse() - this is turned that to an int
> The result?
> You have a number.
My code does exactly that. 
Pormpting int(age) asks for age, int.parse parses the readlined number and it does give number.

Src/BootCamp.Chapter/Lesson3.cs
Zacc806: what shall i write here? Messages prompting for name, surname, age and height/weight?

Almantask: Yes, as the name implies, demo that the method works.

Zacc806: Fixed it and all the other problems, Should be fine now

Almantask (approved):
Looks good, just be careful to what branch are you merging.

PR #502, opened 2020-07-21
Chapter1/homework/9

agakesik:
Finished homework for chapter 1/lesson 9

- create convers: binary to decimal, and decimal to binary
- create function that prints arrows when WSAD keys are used


Almantask (approved):
Great!

PR #504, opened 2020-07-21
Took the original console app from Homework 1And2 and refactored it.

hyp-do:
I feel very strongly that the global variables are likely a no-no. Lets give it a shot.


Almantask (approved):
Looks good. Just please avoid having that weird formatting that you have in main method.

PR #505, opened 2020-07-23
Chapter3/homework/9

matthew-hole:
I believe the solution to the *described* problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.

I found this task tricky ( and *very* time consuming) but I think that was down to me over complicating it and getting stuck on simple things.

Going through it I do see that there is a lot of scope for potential code reuse through inheritance but I opted to stay away from that to try and make everything decoupled - this probably added to the ammount of time this took.

I would still ask that you please review this and as always suggest improvements. I am not so proud of this work, I could of made it better by adding try catch statements for dealing with the expressions, but I want to move onto the rest of the C# sugar.

Thanks again and I appreciate the time you guys spend creating the homeworks and lessons.

Thanks and see you in homework 10!


Almantask:
> I believe the solution to the _described_ problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.

I used guid as file names to fake console. That was the only reason why the files were being created- to create a unique file each time to which console out would be swapped with a file.

matthew-hole:
> > I believe the solution to the _described_ problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.
> 
> I used guid as file names to fake console. That was the only reason why the files were being created- to create a unique file each time to which console out would be swapped with a file.
> 
> 

Right I understand. Potentially I misread the test but it did seem that it was creating a file of Guid but as it didn't exist my solution kept throwing exceptions. 

I should of checked the unittests first. Thank you for clarifying 😊

matthew-hole:
> > I believe the solution to the _described_ problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.
> 
> I used guid as file names to fake console. That was the only reason why the files were being created- to create a unique file each time to which console out would be swapped with a file.
> 
> 

If this approach is okay with you, currently I am doing Homework 11 which implements the code from homework 9 - I propose to implement these changes there. If you agree then could you provide comments and close this PR and I will reference the changes specified here in HW 11?

Almantask (changes requested):
There are many parts that repeat themselves, functions do too many thigns and should be split.

matthew-hole:
> 
> 
> > I believe the solution to the _described_ problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.
> 
> I used guid as file names to fake console. That was the only reason why the files were being created- to create a unique file each time to which console out would be swapped with a file.

@Almantask Can you please elaborate as I think there is some confusion. As I understand it, the output file should be a valid directory where the user wishes to save the output? In the unittests, it doesnt make sense to me why the genuine output file would be of a random Guid string? This would make sense to me if you were checking of an illegitimate file location sure, but in this context I am confused. 

Can you please give me some context I have implemented check for this i.e check that output file location is valid?

Almantask:
> > > I believe the solution to the _described_ problem is complete. As seen, not unit tests are passing. This is due to the either the potential ambiguity in the requirements (homework spec) or that the Unit Tests implemented it differently. By this I mean using a Guid as an output file (this always broke my solution as it was implemented to check for a valid output location/directory) as well as different file naming convention (and how the info was structured within the file), in addition to the unit test seemingly not checking the produced file. I know usually the unit tests **_are_** the requirements so I guess this is on me not checking them first.
> > 
> > 
> > I used guid as file names to fake console. That was the only reason why the files were being created- to create a unique file each time to which console out would be swapped with a file.
> 
> @Almantask Can you please elaborate as I think there is some confusion. As I understand it, the output file should be a valid directory where the user wishes to save the output? In the unittests, it doesnt make sense to me why the genuine output file would be of a random Guid string? This would make sense to me if you were checking of an illegitimate file location sure, but in this context I am confused.
> 
> Can you please give me some context I have implemented check for this i.e check that output file location is valid?

Those files with guids are for console output. When I do a writeline, I write it not to a console, but to a file, so I can later verify the input. I use guid for a name, because I want it to be unique for every test and not risk overlapping names.

Almantask (approved):
I believe that the homework is hard, but the essense of it is completed in this PR. Carry on 😄

matthew-hole:
> I believe that the homework is hard, but the essense of it is completed in this PR. Carry on 😄 

@Almantask Yes it definitely is the biggest so far. I have been caught up in other things but I've been refactoring the whole solution and acting on your comments for HW11. 

I will be interested in seeing what you think when it's done! Thanks for reviewing!

PR #506, opened 2020-07-24
Chapter2/homework/1

agakesik:
I think it took me way longer than it should xd

But I've refactored code from chapter1/lesson8 to be object oriented and made sure it passed all tests


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -9,31 +10,53 @@ public static class Checks
...
         public static void Clean(string file, string outputFile)
         {
+            var fileWithBalances = new FileWithBalances(file);
+            string cleanedContent = fileWithBalances.GetContent();
+            File.WriteAllText(outputFile, cleanedContent);
+        }
+
+        private static PeopleAndBalances CreatePeopleObjectFromArrayInput(string[] peopleAndBalances)
+        {
+            if (peopleAndBalances == null) peopleAndBalances = new string[0];
```
Almantask: This line does 2 different thing, should be split into another line.
Please use brackets in order to explicitly say what happens if the if condition is true.

Src/BootCamp.Chapter/FileWithBalances.cs
```diff
@@ -0,0 +1,99 @@
...
+            try
+            {
+                content = File.ReadAllText(pathToFile);
+            }
+            catch (Exception ex)
+            {
+                throw new ArgumentException("Path to source file does not exist", ex);
+            }
+
+            if (String.IsNullOrEmpty(content))
```
Almantask: Please clarify the purpose of this. Looks redundant.

Src/BootCamp.Chapter/FileWithBalances.cs
```diff
@@ -0,0 +1,99 @@
...
+        private static bool IsContentValid(string contentToValidate)
+        {
+            
+            var differentPeople = contentToValidate.Split(Environment.NewLine);
+
+            foreach (var person in differentPeople)
+            {
+                var nameAndBalances = person.Split(",");
+                if (!ISNameValid(nameAndBalances[0])) return false;
+                if (nameAndBalances.Length > 1 && !AreBalancesValid(nameAndBalances[1..^0])) return false;
```
Almantask: replace magic numbers with named constants.

Src/BootCamp.Chapter/FileWithBalances.cs
```diff
@@ -0,0 +1,99 @@
...
+            return true;
+        }
+
+        private static bool ISNameValid(string name)
+        {
+            var nameRegex = new Regex(@"[a-zA-Z]+ [a-zA-Z]+'?-?[a-zA-Z]+[.]?$");
+
+            if (!nameRegex.IsMatch(name, 0))
+            {
+                Console.WriteLine(name);
```
Almantask: Side effect- should be removed.

Almantask: Can be simplified:
```cs
private static bool ISNameValid(string name)
{
            var nameRegex = new Regex(@"[a-zA-Z]+ [a-zA-Z]+'?-?[a-zA-Z]+[.]?$");
            var isMatch = nameRegex.IsMatch(name, 0);
            return isMatch ;
}
```

Src/BootCamp.Chapter/FileWithBalances.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            return true;
+        }
+
+        private static string Clean(string content)
+        {
+            var contentToClean = content;
+            contentToClean = contentToClean.Replace("_", "");
+
+            if (!IsContentValid(contentToClean)) throw new InvalidBalancesException("Balances file is still invalid after cleaning up the file");
```
Almantask: add brackets {}, do not inline throwing exceptions.

Src/BootCamp.Chapter/Person.cs
```diff
@@ -0,0 +1,96 @@
...
+                    highestBalance = balance;
+                }
+            }
+
+            return highestBalance;
+        }
+
+        public double GetBiggestLoss()
+        {
+            if (_balances.Count <= 1) return double.MinValue;
```
Almantask: why does it get compared 1

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -0,0 +1,145 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class TextTable
+    {
+        /*
```
Almantask: Comment can be remvoed, was just an example

Almantask (changes requested):
Looks good but could be better.

Src/BootCamp.Chapter/FileWithBalances.cs
agakesik: The idea was to not do validation and/or cleaning if input is null (it's the first time checking if content is null/empty), but actually I think I have a better way to structure it

Almantask (approved):
Good 😄

PR #507, opened 2020-07-24
Chapter1/Homework 4.

Tailor-AI:
Code for homework 4. 
Refactored code from Lesson 3 with some error/input checking.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -6,11 +6,7 @@ using System.Globalization;
 namespace BootCamp.Chapter
 {
     public class Lesson3
-    {
-        // Used with float.TryParse to make tests pass ("." instead of "," in tests).
-        static CultureInfo invC = CultureInfo.InvariantCulture;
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    public class Lesson3
+    {
+        // Used with float.TryParse to make tests pass ("." instead of "," in tests).
+        static CultureInfo invC = CultureInfo.InvariantCulture;
```
Almantask: Why static?

Tailor-AI: I'm not entirly sure how that object works but it was to make the object visible in the function. 
Now I moved it down and put it in the function and it works without static.

My understanding was that you read it once as a global (static) variable/setting and then use it thruout the program?

Almantask (approved):
Looks good

PR #508, opened 2020-07-25
Chapter1/homework/4

Zacc806:
Test for height and weight for being 0 was wrong, so i tried fixing it, now it works well


Almantask (approved):
Looks good

PR #511, opened 2020-07-27
Chapter3/homework/10

matthew-hole:
Solution to Homework 10. 

Initially overcomplicated the creating of the TextTable but after that it was fairly straightforward. Used your example from the live coding part of the lesson 'AttributeGetter' as I was initially unsure how I was to retrieve the 'attributes'.

Nevertheless, homework is done!

### DIFF
--- a/Src/BootCamp.Chapter/TextBoxBuilder.cs
+++ b/Src/BootCamp.Chapter/TextBoxBuilder.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public class TextBoxBuilder
+    {
+        private readonly string _message;
+        private readonly TextTableAttribute _textTableAttributes;
+        private readonly int _messageLength;
+        private readonly int _numberOfLines;
+        private readonly int _numberOfMainLine;
+        private readonly string _paddingString;
+        private StringBuilder sb = new StringBuilder();
+        private const int DivideByTwo = 2;
+        private const int AddOne = 1;
+
+        public TextBoxBuilder(TextTableAttribute textTableAttributes, string message)
+        {
+            _textTableAttributes = textTableAttributes;
+            _message = message;
+            _messageLength = _message.Length + (_textTableAttributes.Padding * DivideByTwo);
+            _numberOfLines = 3 + (_textTableAttributes.Padding * DivideByTwo);
+            _numberOfMainLine = (_numberOfLines + AddOne) / DivideByTwo;
+            _paddingString = new string(' ', _textTableAttributes.Padding);
+        }
+
+        public string BuildTextBox()
+        {
+            string topAndBottom = BuildTopAndBottomLines();
+            string mainLine = BuildMainLine();
+            string mainLineWithPadding = BuildMainLineWithPadding();
+            string paddingLine = BuildPaddingLine();
+
+            for (int i = 1; i < _numberOfLines + AddOne; i++)
+            {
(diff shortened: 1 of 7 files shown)


Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,58 @@
...
+                    sb.AppendLine(paddingLine);
+                }
+                else if (i == numberOfMainLine && textTableAttributes.Padding == 0)
+                {
+                    sb.AppendLine(mainLine);
+                }
+                else if (i == numberOfMainLine && textTableAttributes.Padding != 0)
+                {
+                    sb.AppendLine(mainLineWithPadding);
+                }
```
matthew-hole: All of this above could probably be improved through refactoring into its own private method to reduce the complexity in the Print() method. Let me know if you spot any other ways of achieving this.

Src/BootCamp.Chapter/TextBoxPrinter.cs
```diff
@@ -1,12 +1,58 @@
...
+
+            var table = GenerateTable(getTextTableAttribute, obj.ToString());
+            Console.WriteLine(table);
+
+            return GenerateTable(getTextTableAttribute, obj.ToString());
+        }
+
+        private static string GenerateTable(TextTableAttribute textTableAttributes, string message)
+        {
+            int messageLength = message.Length + (textTableAttributes.Padding * 2);
```
Almantask: Get rid of magic numbers 😄

Src/BootCamp.Chapter/TextBoxPrinter.cs
Almantask: Maybe not only private methods, maybe a class of its' own. Builder pattern and fluent interface sounds like a good fit.

Also, a lot of redundancies are spotted.

Almantask (changes requested):

Src/BootCamp.Chapter/TextBoxPrinter.cs
matthew-hole: Unsure how to remove the magic numbers when they are fundamental to the logic used? What is the best practice for achieving this?

Almantask: Instead of magic numbers, used named constants.

matthew-hole:
@Almantask Please see commit "f5fc7c1" for changes relating to your comments. 

Note that there is still some complexity with multiple else if statements. I spent a while researching state-machines/state-pattern and now understand its uses and deemed it was not sensible to be used here. 

PR should be good to go! Thanks.

Src/BootCamp.Chapter/TextBoxBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class TextBoxBuilder
+    {
+        private readonly string _message;
+        private readonly TextTableAttribute _textTableAttributes;
+        private readonly int _messageLength;
+        private readonly int _numberOfLines;
+        private readonly int _numberOfMainLine;
+        private readonly string _paddingString;
+        private StringBuilder sb = new StringBuilder();
+        private const int DivideByTwo = 2;
```
Almantask, line 15: DivideByTwo or Addone does not make anything better. You need to describe those constants in a logical why (what are they used for?). 

You did good with `NumberOfLinesWithoutPadding`

Almantask (approved):

Almantask (approved):
Yeah, I don't see much to pick on. Carry on 👍

PR #512, opened 2020-07-27
homework Chapter2/lesson3

agakesik:
Completed homework for lesson 3 in chapter 2.

Created abstract class with Asseble method and structure it in a way, that child classes only change components in a method

*FYI: readme for this homework is blank, so I used discord to find tasks. Not sure if it's intended or not


Almantask:
Not intended

Almantask (approved):
Yup, this was exactly what i wanted.

PR #514, opened 2020-07-28
Chapter2/homework/4

agakesik:
Completed hw for chapter2/lesson4

- Created ILogger interface with 2 child loggers
- Recreated logic from hw for chapter1/lesson1&2 using objects
- Created program that
  a) let user choose prefered logger
  b) gathers personal information
  c) logs personal information and information on BMI to chosen logger

### DIFF
--- a/Src/BootCamp.Chapter/Person.cs
+++ b/Src/BootCamp.Chapter/Person.cs
@@ -1,0 +1,125 @@
+﻿using BootCamp.Chapter.Logger;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Person
+    {
+        private string _name;
+        private int _age;
+        private float _weight;
+        private float _height;
+
+        public void SetInformation(ILogger logger)
+        {
+            SetName(logger);
+            SetAge(logger);
+            SetWeight(logger);
+            SetHeight(logger);
+        }
+
+        public void SetName(ILogger logger)
+        {
+            while (true)
+            {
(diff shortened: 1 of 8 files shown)


Almantask:
Holy crap! Well done 🙂 I mean so fast and dedicated, you are an inspiration

agakesik:
Ah damn, thanks! 
I just have lots of free time 😅 The thing is your resources and help make it reeeeally easy to stay focused and motivated, and I just go with it because I'm constantly curious what's gonna be next 😃

Src/BootCamp.Chapter/Person.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            SetName(logger);
+            SetAge(logger);
+            SetWeight(logger);
+            SetHeight(logger);
+        }
+
+        public void SetName(ILogger logger)
+        {
+            while (true)
```
Almantask, line 25: quite scare to see such a loop condition. Try to design that access is given explicitly and not broken explicitly. Having it this way you risk an infinite loop more.

Almantask (approved):
Looks good

Src/BootCamp.Chapter/Person.cs
agakesik: funny, because I always thought so, but I've seen it A LOT in different python projects. And I just figured it's a standard.
Will  avoid it in future

Almantask: Because people don't care. People are result focused.

Subtle details like this don't require much effort to change, but significantly improve readability and maintainability.

PR #515, opened 2020-07-30
Chapter3/homework/1

agakesik:
Refactored code from homework for chapter2/lesson2 to use Properties (instead of fields), Lists (instead of arrays) and Indexer (instead of seperate method)

### DIFF
--- a/Src/BootCamp.Chapter/Equipment.cs
+++ b/Src/BootCamp.Chapter/Equipment.cs
@@ -7,63 +7,19 @@ namespace BootCamp.Chapter
     /// </summary>
     public class Equipment
     {
-        private Weapon _weapon;
-        public void SetWeapon(Weapon weapon)
-        {
-            _weapon = weapon;
-        }
-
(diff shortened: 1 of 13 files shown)


Src/BootCamp.Chapter/Equipment.cs
```diff
@@ -7,85 +7,60 @@ namespace BootCamp.Chapter
...
-        {
-            return 0;
+            return totalDefence;
         }
 
-        /// <summary>
-        /// Returns damage done by weapon.
-        /// </summary>
-        /// <returns></returns>
         public float GetTotalAttack()
```
Almantask, line 56: Try to forget about get methods, unless they have rich logic. 
```cs
public float TotalAttack
{
  get
  {
                if (Weapon == null)
            {
                return 0;
            }

            return Weapon.Attack;
  }
}
```
is totally fine

Almantask (approved):
Looks great!

PR #517, opened 2020-07-31
Chapter3/homework/2

agakesik:
Finished homework for chapter3/lesson2

Sooo.. this was quite a bit of headache 😅 
I kinda couldn't wrap my head around it, so I started with tests and just tried to write code that would make them pass one by one.
In the end all tests pass, but I'm not sure if it's written well or not :v 

The biggest problem I had was with comparing two credentials: is overriding Equals method [to compare (by value) properties of objects, instead of comparing 2 objects by reference] a bad practice?

### DIFF
--- a/Src/BootCamp.Chapter/Credentials.cs
+++ b/Src/BootCamp.Chapter/Credentials.cs
@@ -1,22 +1,64 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Linq;
+using System.Text;
+
+namespace BootCamp.Chapter
 {
     // TODO: make a struct and add validation and other needed methods (if needed)
     public class Credentials
     {
-        public string Username;
-        public string Password;
+        public readonly string Username;
+        public readonly string Password;
 
         public Credentials(string username, string password)
         {
+            if (string.IsNullOrEmpty(username))
+            {
+                throw new ArgumentException("You must provide username");
+            }
+            if (string.IsNullOrEmpty(password))
+            {
+                throw new ArgumentException("You must provide password");
+            }
             Username = username;
-            Password = password;
+            var encodedPassword = new EncodedPassword(password);
+            Password = encodedPassword.PasswordAfterEncoding;
         }
 
         // TODO: Implement properly.
         public static bool TryParse(string input, out Credentials credentials)
         {
             credentials = default;
-            return false;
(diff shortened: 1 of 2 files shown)


Almantask:
Overriding equals is a good practice

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,22 +1,64 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Linq;
+using System.Text;
+
+namespace BootCamp.Chapter
 {
     // TODO: make a struct and add validation and other needed methods (if needed)
```
Almantask, line 7: Missed this TODO :) simply change class to struct.

Src/BootCamp.Chapter/Credentials.cs
```diff
@@ -1,22 +1,64 @@
...
+
+            var CredentialsParts = input.Split(",");
+            if (CredentialsParts.Length != 2) return false;
+
+            if (string.IsNullOrEmpty(CredentialsParts[0]) || string.IsNullOrEmpty(CredentialsParts[1])) return false;
+
+            credentials = new Credentials(CredentialsParts[0], CredentialsParts[1]);
+            return true;
+        }
+        public readonly struct EncodedPassword
```
Almantask, line 42: I think this nested class is not needed. Essentially, not having it is the same as just having a string. Right now, it's a roundtrip with no value.

Almantask: All that is needed is already encapsulated in Credentials class.

Src/BootCamp.Chapter/CredentialsManager.cs
```diff
@@ -1,24 +1,53 @@
...
+            }
             _credentialsFile = credentialsFile;
         }
 
         // TODO: load credentials and check for equality.
         public bool Login(Credentials credentials)
         {
+            string credentialsToFind = credentials.GetCredentials();
+
+            string fileContent = File.ReadAllText(_credentialsFile);
```
Almantask, line 24: Can be simplified: File.ReadAllLines(...)

Almantask (approved):
Yes, the good still looks good! Believe in yourself 😉

PR #518, opened 2020-07-31
Chapter1/homework/1 and2

Rickolajten:
Added everything according to instructions


Almantask:
The input should be read from console and not hardcoded (with constant values). Refer to Console.ReadLine, int.Parse, float.Parse, etc.

Almantask (changes requested):
Also wrong branch is being targeted. You should target the same branch as the homework

Almantask (approved):
Well done! Beautiful 👍

PR #519, opened 2020-08-04
Chapter1/homework/5

hyp-do:
We were preparing for a natural disaster that (thankfully) didn't happen. Seems my original pull request expired a few hours ago.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -1,0 +1,119 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Reflection.PortableExecutable;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static float globalHeight = 0f;
+        public static float globalWeight = 0f;
+        public static string bmiError = "Failed calculating BMI. Reason:";
+
+        public static string Information()
+        {
+            Console.WriteLine("Testing");
+            string firstName = Console.ReadLine();
+            string errorMsg = "Name cannot be empty.";
+
+            if (String.IsNullOrEmpty(firstName))
+            {
+                Console.WriteLine(errorMsg);
+                return errorMsg;
+            }
+            else
+            {
+                return firstName;
+            }
+        }
+
+        public static int AskAge()
+        {
+            Console.WriteLine("Testing");
+            string sAge = Console.ReadLine();
+
+            int age;
+            bool isNumber = int.TryParse(sAge, out age);
+
+            if (!isNumber) return -1;
+            if (age == 0) return 0;
+            if (age >= 0) return age;
+
+            return 0;
+
+        }
+
+        public static float Weight()
+        {
+            Console.WriteLine("What is your weight in kgs?");
+            string sWeight = Console.ReadLine();
+
+            bool isNumber = float.TryParse(sWeight, out globalWeight);
+
+            if (!isNumber)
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+using System.Reflection.PortableExecutable;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static float globalHeight = 0f;
```
Almantask, line 11: avoid using global, please move this to local scope. Global can be changed from anywhere and in any more complex cases than just this class you will hit yourself to the wall just because you made something global.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static float globalWeight = 0f;
+        public static string bmiError = "Failed calculating BMI. Reason:";
+
+        public static string Information()
```
Almantask, line 15: Methods are actions and therefore should be named after verbs. Information is not informative enough. Please rename to PromptBmi or similar

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            if (!isNumber) return -1;
+            if (age == 0) return 0;
+            if (age >= 0) return age;
+
+            return 0;
+
+        }
+
+        public static float Weight()
```
Almantask, line 48: Keep it consistent. AskAge was a good name. AskWeight, AskHeight should be how you name the other methods too, because that's what they do.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,119 @@
...
+                Console.WriteLine("\"" + sHeight + "\"" + " is not a valid number.");
+                return -1;
+            }
+            if (globalHeight == 0) return 0;
+            if (globalHeight >= 0) return globalHeight;
+
+            return 0;
+        }
+
+        public static float BMIFormula(float weight, float height)
```
Almantask, line 84: Change noun to a verb. CalculateBmi. note that I didn't write Bmi, becasue we use PascalCase even for abreviations (not BMI but Bmi)

Almantask (changes requested):
You did everything good, just the naming and style changes. I know you want to learn, so I would appreciate if you could change those.

Src/BootCamp.Chapter/Lesson4.cs
hyp-do: Thank you for the feedback. I renamed the methods to be more clear and concise as part of reworking the solution.

Src/BootCamp.Chapter/Lesson4.cs
hyp-do: I recently went through and re-worked the solution to remove global variables.

Src/BootCamp.Chapter/Lesson4.cs
hyp-do: Thank you for the feedback, renamed this method to PromptBmi

Src/BootCamp.Chapter/Lesson4.cs
Almantask: Did you push? It seems to be the same

Almantask (changes requested):

PR #521, opened 2020-08-06
Homework chapter 1 and lesson 1

mayank2441:
I am still new to this thing.

I tried the homework i would love to know how i did.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text=auto
+
+###############################################################################
+# Set default behavior for command prompt diff.
+#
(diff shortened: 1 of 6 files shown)


Almantask:
Hey! Thanks for the homework, but there are a few issues:
1) don't hardcode input. Use console to read user input (Console.ReadLine) then convert tha to int or float (int.Parse)
2) select the same target branch as teh source brach. This is hw 1, yet you merge to master.

mayank2441:
Woah thanks for the superfast reply.
I might sound dumb...but i am not really familiar with the term hardcoding input and i am having difficulty to understand how to do that...

Almantask:
Hardcoding- using constant values, instead of passing them as parameters or reading them as user input.

If you want to discuss about this, in 4h time you can come on Discord and we can have a talk. Else, ask 😉

PR #522, opened 2020-08-06
1 and2 homework mayank

mayank2441:
I hope i did the right thing with the merging.

I did the softcoding too. Please tell if there is any other mistake.

And also thanks for all the help.


Almantask:
Close. If you could undo changes to all other files, except program.cs that would be great

1and2/1and2/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace _1and2
+{
+    class Program
+    {
+        static void Main(string[] args)
+        { // PRess enter or type anything and the the code will appear.
+            string name = Console.ReadLine();
+
+            String TomName = "Tom Jefferson";
+             int TomAge = 19;
```
Almantask, line 16: hardcoded values. Please use Console.Readline to read user input. Use int.Parse, Float.Parse to convert user input to numbers where applicable.

Almantask:
You did it almost correct. Now just the last part- using console.

Almantask:
Closed for being duplicate
https://github.com/csinn/CSharp-From-Zero-To-Hero/pull/524

PR #526, opened 2020-08-09
Homework: Ch1, 1and2 Submission

malamunyx:
My homework submission

Any improvements/criticism to the code would be very appreciated

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,43 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name, surname, ageInput, weightInput, heightInput;
+            int age;
+            double weight, height, bodyMassIndex;
+
+            Console.Write("Name: ");
+            name = Console.ReadLine();
+
+            Console.Write("Surname: ");
+            surname = Console.ReadLine();
+
+            Console.Write("Age: ");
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,48 @@
...
+            ageInput = Console.ReadLine();
+
+            Console.Write("Weight: ");
+            weightInput = Console.ReadLine();
+
+            Console.Write("Height: ");
+            heightInput = Console.ReadLine();
+
+            age = int.Parse(ageInput);
+            // Convert age string input weight into int
```
Almantask, line 29: Comments are good, but normally the convention is to put them above (and not below) the line that you explain. They are meant to explain, so it is nicer to read them before reading the actual line of code.

Almantask (approved):
Well done. I like the comments for learning 🛩️

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,48 @@
...
-}
+﻿using System;
+
+namespace BootCamp.Chapter
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            string name, surname, ageInput, weightInput, heightInput;
```
Almantask, line 9: It's recommended to just declare a variable right before you use it and not all in one place. Like that, you will have the variable at the place where you need to see it and not scroll up and down.

Almantask (approved):

PR #527, opened 2020-08-09
Chapter 1-3

YoRHa-A5:
My solution to Chapter 1-3


Almantask (approved):
As simple as that. Well done!

PR #528, opened 2020-08-10
Chapter1/homework/4

YoRHa-A5:
My solution to Chapter 1-4

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,93 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            var i = Console.ReadLine();
+            var isNumeric = int.TryParse(i, out int number);
+            if (String.IsNullOrEmpty(i))
+            {
+                return 0;
+            }
+            else if (!isNumeric)
+            {
+                Console.Write($"\"{i}\" is not a valid number.");
+                return -1;
+            }
+            else if (isNumeric && number == 0)
+            {
+                Console.Write($"\"{i}\" is not a valid number.");
+                return 0;
+            }
+            return number;
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
-        }
+            Console.WriteLine(message);
+            var s = Console.ReadLine();
+            if (String.IsNullOrEmpty(s))
+            {
+                Console.Write("Name cannot be empty.");
+                return "-";
 
+            }
+            else
+            {
+                return s;
+            }
+        }
         public static float PromptFloat(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            var f = Console.ReadLine();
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
...
+            if (String.IsNullOrEmpty(i))
+            {
+                return 0;
+            }
+            else if (!isNumeric)
+            {
+                Console.Write($"\"{i}\" is not a valid number.");
+                return -1;
+            }
+            else if (isNumeric && number == 0)
```
Almantask, line 31: You checked the numeric in if above, you don't need to check it again.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            var i = Console.ReadLine();
+            var isNumeric = int.TryParse(i, out int number);
+            if (String.IsNullOrEmpty(i))
```
Almantask, line 22: I would recommend separating the if statement of the value that you are checking with a blank line form the rest of code.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
...
+            return number;
         }
 
         public static string PromptString(string message)
         {
-            // To do: call your implementation. 
-            return "";
-        }
+            Console.WriteLine(message);
+            var s = Console.ReadLine();
```
Almantask, line 42: Avoid using variable names with a single character. Calling it "input" is even better than calling it string (you would need to add @ symbol to allow variable names after keywords)

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
...
+                Console.Write("Name cannot be empty.");
+                return "-";
 
+            }
+            else
+            {
+                return s;
+            }
+        }
         public static float PromptFloat(string message)
```
Almantask, line 54: Separate methods with a blank line in betweeen

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
...
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            var f = Console.ReadLine();
```
Almantask, line 57: Instead of f name it number.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,93 @@ public static class Checks
...
+                return -1;
+            }
+            else if (!isNumeric)
+            {
+                Console.Write($"\"{f}\" is not a valid number.");
+                return -1;
+            }
+            else if (isNumeric && number == 0)
+            {
+                //Console.WriteLine($"\"{f}\" is not a valid number.");
```
Almantask, line 75: Remove commented out code.

Almantask (approved):
Overall I like it, looks good 😄

PR #530, opened 2020-08-11
Chapter1/homework/1 and2

ACHYUT001:
Completed Code to read and print user info.

Note: Used VS Code because my team needs me to use VS Code, therefore, this will serve as practice 😊


Almantask (changes requested):
You don't calculate Bmi, second person is missing (doing the same thing twice). vscode files aren't really needed.

PR #531, opened 2020-08-11
Completed homework 3

syed87:
homework 3 completed

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Dynamic;
 using System.Text;
 
 namespace BootCamp.Chapter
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,31 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation.
```
Almantask, line 20: All comments can be removed.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,31 @@ public static class Checks
...
+            return float.Parse(Console.ReadLine());
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation.
+            float mSquared = height * height;
+            float bmi = weight / mSquared;
```
Almantask, line 43: Should simply have multiplication and division in the same line, because they represent the same thing.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            static void bmiCalculator()
+            {                
+                string firstName = Checks.PromptString("What is your first name?");
+
```
Almantask, line 14: No need a blank line after every statement

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            static void bmiCalculator()
+            {                
+                string firstName = Checks.PromptString("What is your first name?");
```
Almantask, line 13: Call PromptString function in checks. PromptString function should be in YOUR OWN static class. Checks class is there to simply allow you to name your function and place them in whatever classes you want.

Almantask (changes requested):

PR #532, opened 2020-08-16
Made all the stuff work, but tests are unsuccessful.

JohnHNewland:
So yeah, refactored into the functions, it functions, but when I do tests, I fail everything.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,26 +16,26 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            Program.AskingAndConvertingINT();
-            return 0;
+            int integer = Program.AskingAndConvertingINT();
+            return integer;
         }
 
         public static string PromptString(string message)
         {
-            Program.AskingAndConvertingSTRING();
-            return "";
+            string thePrompt = Program.AskingAndConvertingSTRING();
+            return thePrompt;
         }
 
         public static float PromptFloat(string message)
         {
-            Program.AskingAndConvertingFLOAT();
-            return 0;
+            float number = Program.AskingAndConvertingFLOAT();
+            return number;
         }
 
         public static float CalculateBmi(float weight, float height)
         {
-            Program.CalculateBMI(weight, height);
-            return 0;
+            float theBmi = Program.CalculateBMI(weight, height);
+            return theBmi;
         }
     }
 }
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,19 +16,20 @@ namespace BootCamp.Chapter
     {
         public static int PromptInt(string message)
         {
-            int integer = Program.AskingAndConvertingINT();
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,25 +16,25 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
+            Program.AskingAndConvertingINT();
             return 0;
```
Almantask: All tests are failing because you are not reusing the returns of Program.AskingAndConverting... functions. Return the result instead of default values (0, "", etc)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +11,55 @@ class Program
...
+            Console.WriteLine(secondName + " " + secondSurname + " is " + secondAge + " years old, his weight is " + secondWeight + " kg and his height is " + secondHeight + " cm.");
+
+            float secondBmi = CalculateBMI(secondWeight, secondHeight);
+
+            Console.WriteLine("His BMI is " + secondBmi + ".");
+
+            Console.ReadLine();
+
+        }
+        public static float CalculateBMI(float weight, float height)
```
Almantask, line 49: Don't have those methods in Program class. Program class is just for bootstrapping. These should be in their own class.

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +11,55 @@ class Program
...
+        {
+            float bmi = weight / ((height / 100) * (height / 100));
+            return (float)Math.Round(bmi, 1);
+        }
+        public static int AskingAndConvertingINT()
```
Almantask: Follow the PascalCaseNotations. Based on the notation the name should be:
`AskingAndConvertingInt`. Also, the name can be simplified: `PromptString`. Same goes for all other names, hard to read.

Src/BootCamp.Chapter/Checks.cs
JohnHNewland: Owww, yeah, I am stupid, thanks!

Src/BootCamp.Chapter/Program.cs
JohnHNewland: First of all, is this a question of style? Or it is a requirement? Why would I take a class just for functions, like a bank?

Src/BootCamp.Chapter/Program.cs
JohnHNewland: Yeah, I have no idea, why did I upper case the type, my fault.

Src/BootCamp.Chapter/Program.cs
Almantask: It's a style, it's not a requirement. As  I mentioned, having logic in Program class is out of place, it should just kickstart the application. No biggie for the first chapter though.

Almantask (approved):
Use conventional naming, constant string values can be in the method you use them, instead of all being at the top.

PR #535, opened 2020-08-21
Complete Homework 1 and 2

kungfuphil:
Read name, surename, age, weight (in kg) and height (in cm) from console.

Print all the info based on the example message below:
Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm.

Calculate and print body-mass index (BMI)

Do 1 and 2 for another person.


Almantask (approved):
All good, but comments can be removed when you are done 😉

PR #536, opened 2020-08-23
Chapter2/homework/2

Laura582:
This took me longer than I thought it would. Didn't do that extra challenge. :( 

This program has an inventory where player can buy or sell items from or to a shop. The shop can sell or buy items also.


Almantask (approved):
Looks good 😄

PR #538, opened 2020-08-23
Chapter 1 Homework 1 and 2

Alkhriz:
Completed the assigned tasks 1-3 for Chapter 1 homework 1-2. Feedback would be greatly appreciated.

### DIFF
--- a/Homework/Homework 1 and 2.cs
+++ b/Homework/Homework 1 and 2.cs
@@ -36,7 +36,6 @@ namespace BootCamp.Chapter
             decimal otherHeight = 162.6m;
             string otherName = "Rafa";
             string otherSurname = "Erende";
-
             Console.WriteLine(otherName + " " + otherSurname + " " + "is " + otherAge + " " +"years old " + "and weighs " + otherWeight + " " + "kilograms " + "and measures " + otherHeight + " " + "centimeters.");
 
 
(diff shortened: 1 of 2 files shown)


Homework/Homework 1 and 2.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            int myAge = 25;
+            decimal myWeight = 77m;
+            decimal myHeight = 167.6m;
+            string myName = "Al";
+            string mySurname = "Serna";
+            Console.WriteLine( "My name is " + myName + " " + mySurname + " " + "I am " + myAge + " " + "years old, " + "I weigh " + myWeight + " " + "kilograms " + "and I measure " + myHeight + " " + "centimeters." );
+            
+
+
```
Almantask, line 21: Absolutely unnecessary to leave so many blank lines- hits to readability.

Almantask: It's good that you seperate, but 1 line is enough 👍

Homework/Homework 1 and 2.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace BootCamp.Chapter
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            
+
+            int myAge = 25;
+            decimal myWeight = 77m;
```
Almantask, line 14: Read input from console. Refer to:
Console.ReadLine()
int.Parse, float.Parse

Almantask, line 14: weight should be either float or double (I think float suits better). You don't need high precision nor big number for weight, therefore we don't care if 1/100 of mg will be dropped.

Either way, decimal is a bad choice for datatype, because it is used for high precision calculations, for example finance or scientific calculations.

Almantask (changes requested):

PR #540, opened 2020-08-26
Chapter1/homework/1 and2

claytonellison:
Please review Chapter 1 Homework 1 and 2. User is prompted to enter: first name, surname (last name), age, weight, and height. Using this information two sentences are generated detailing the information provided and the user's BMI.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,16 +7,20 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm.");
+            string TomsName = "Tom Jefferson";
+            int TomsAge = 19;
             int TomsWeight = 50;
             float TomsHeight = 156.5f;
             float TomsBMI = ((TomsWeight / TomsHeight)/TomsHeight)*10000;
-            Console.WriteLine("Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm. His BMI is " + TomsBMI + ".");
+            Console.WriteLine(TomsName + " is " + TomsAge + " years old, his weight is " + TomsWeight + " kg, his height is " + TomsHeight + " cm, and his BMI is " + TomsBMI + ".");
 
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,32 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string FirstName;
```
Almantask: Don't declare all the variable in one place. Declare them as close as possible to their usage.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,32 @@ class Program
...
+            string AgeEntered;
+            int Age;
+            string WeightEntered;
+            double Weight;
+            string HeightEntered;
+            double Height;
+            double BMICalculated;
+            double BMIRounded;
+            Console.WriteLine("What is your first name?");
+            FirstName = Console.ReadLine();
```
Almantask: Split different parts of code:
Read 1 thing

Read another thing

Read yet another thing

Bmi calculation

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,32 @@ class Program
...
+            Console.WriteLine("Enter your age as a number please.");
+            AgeEntered = Console.ReadLine();
+            Age = int.Parse(AgeEntered);
+            Console.WriteLine("What is your weight in kilograms?");
+            WeightEntered = Console.ReadLine();
+            Weight = double.Parse(WeightEntered);
+            Console.WriteLine("What is your height in centimeters?");
+            HeightEntered = Console.ReadLine();
+            Height = double.Parse(HeightEntered);
+            BMICalculated = (Weight/Height/Height)*10000;
```
Almantask: variables should be named using camelCase notation (first non-capital letter and then every new word capitlized). Also, only the first letter in abreviation (if it's the second word) should be capitalized. In your case, since it's the first word for a variable, it's not needed to be capitalized.

Almantask (changes requested):
Did what is asked, but I would appreciate if you could clean up a bit. Right now, code looks like one big blog. Use blank lines and renamed slightly more conventional, as suggested.

Almantask (approved):
Much better. Thanks 🥇

PR #534, opened 2020-08-18
Chapter3/homework/12

matthew-hole:
Homework Complete

- Operator Overloading has been demonstrated and Unit Tests are passing.
- Potential to implement better NullOrEmpty checking as well as making the TryParse check more sophisticated (by checking the legitiamacy of the address - could be used through a RoyalMail checker if addresses are real)
- Potential scope to create a 'Demo' although as the homework is small and functionality simple I deemed this unnecessary.

See you in the next chapter!

PR #541, opened 2020-08-27
Chapter3/homework/12

matthew-hole:
Updates were made to the initial PR #534. Comments have been addressed and tests are passing.

Apologies for the not addressing #534 sooner, I was away. 

Cheers!


Src/BootCamp.Chapter/PostOfficeHq.cs
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        }
+
+        private bool DoesPostOfficeExist(Address address)
+        {
+            return ListOfPostOffices.Any(x => x.Location == address.Town);
+        }
+
+        private PostOffice ReturnTownPostOffice(string town)
```
Almantask: FindPostOfficeAt sounds better

Almantask (changes requested):
Still missing duplciates

matthew-hole:
> Still missing duplciates

Can you elaborate please as I believe the program does that? When run there are no duplicates in a post offices 'list' of addresses and correctly increments the duplicate count? Maybe I am missing the objective slightly?

Src/BootCamp.Chapter/PostOffice.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class PostOffice
+    {
+        public string Location;
+        public List<Address> PostAddresses;
+        public int DuplicateCount;
```
Almantask: Where is the duplicate assignment? I don't see it being incremented.

Almantask (changes requested):
Also, all that you have public is a field. NEVER expose fields publicly, because it introduces more ways in which the application could break. Use properties instead.

Almantask:
> > Still missing duplciates
> 
> Can you elaborate please as I believe the program does that? When run there are no duplicates in a post offices 'list' of addresses and correctly increments the duplicate count? Maybe I am missing the objective slightly?

Maybe I didn't notice, but I don't see where do you actually calculate the duplicates count. I see the variable, but not its' use, nor do I see it in demo.

Src/BootCamp.Chapter/PostOfficeHq.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                PostOffice tempPostOffice = ReturnTownPostOffice(address.Town);
+
+                if (tempPostOffice.PostAddresses.All(x => x != address))
+                {
+                    tempPostOffice.AddPostalAddress(address);
+                }
+                else
+                {
+                    tempPostOffice.DuplicateCount++;
```
matthew-hole, line 77: @Almantask Here is the usage of DuplicateCount. It is incremented when an already stored address in the PostOffice Addresses is found (using operator overloading).

Please let me know if there is a more elegant or best practice way of achieving this.

Src/BootCamp.Chapter/PostOfficeHq.cs
matthew-hole: Agree, fits the use case more.

matthew-hole:
> 
> 
> Also, all that you have public is a field. NEVER expose fields publicly, because it introduces more ways in which the application could break. Use properties instead.

I agree, this was a mistake of mine. I will implement the changes and bear them in mind for future projects.

Src/BootCamp.Chapter/PostOfficeHq.cs
Almantask: Yeah, it's very hidden. I think it would be better to:
1) read all offices
2) out of all offices, get duplicates count (always computed, not stored)

Src/BootCamp.Chapter/PostOfficeHq.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        private void AddAddressesToPostOffice()
+        {
+            foreach (var address in _addresses)
+            {
+                PostOffice tempPostOffice = ReturnTownPostOffice(address.Town);
+
+                if (tempPostOffice.PostAddresses.All(x => x != address))
```
Almantask, line 71: LINQ All will check all elements in the collection.
Invert All and use Any, which will stop at first element that fails the condition.

Almantask (approved):

PR #542, opened 2020-08-30
Chapter1/Homework/9

Tailor-AI:
First draft for homework 9.


Almantask (approved):
I like what I see. The only nitpick is to split if statements and for blocks with a blank line.

PR #543, opened 2020-08-30
Chapter2/homework/3

Laura582:
I haven't used polymorphism so probably I will need to make changes but it works...! 

This program creates 2 mac factories, 2 win factories and builds 100 pcs in each


Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -9,7 +9,17 @@ public class MsFactory
     {
         public DesktopComputer Assemble()
         {
-            return new DesktopComputer();
+            DesktopComputer desktopComputer = new DesktopComputer(new Body(), new Ram(), new Cpu(), new Gpu(), new HardDisk(), new Motherboard());
```
Almantask: Computer is created with empty slots, it starts from a body.

Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -9,7 +9,17 @@ public class MsFactory
...
+
+            desktopComputer.GetBody();
```
Almantask: Instead of ``Get``, use ``Install``. We are supposed to install different parts into a computer.
And each factory installs their own brand parts.

Almantask (changes requested):

Src/BootCamp.Chapter/Computer/Installation.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using BootCamp.Chapter.Computer;
+
+namespace BootCamp.Chapter.Computer
+{
+    public abstract class Installation
```
Almantask: It's a factory, not installation.

Src/BootCamp.Chapter/MacFactory.cs
```diff
@@ -1,15 +1,48 @@
...
 namespace BootCamp.Chapter
 {
-    public class MacFactory
+    public class MacFactory : Installation
     {
-        public DesktopComputer Assemble()
+
+        public override Body InstallBody()
+        {
+            Body body = new Body();
```
Almantask: should create MacBody, MacCpu, MacRam, etc...

Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -5,11 +5,44 @@
 
 namespace BootCamp.Chapter
 {
-    public class MsFactory
+    public class MsFactory : Installation
```
Almantask: should create MsBody, MsCpu, MsRam, etc...

Almantask (changes requested):

Almantask (approved):
Looks good

PR #544, opened 2020-09-02
Chapter1/Homework 10

Tailor-AI:
First draft of homework 10.

### DIFF
--- a/Src/BootCamp.Chapter/CaesarCipher.cs
+++ b/Src/BootCamp.Chapter/CaesarCipher.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Reflection.Metadata.Ecma335;
 using System.Text;
 
 namespace BootCamp.Chapter
--- a/Src/BootCamp.Chapter/CaesarCipher.cs
+++ b/Src/BootCamp.Chapter/CaesarCipher.cs
@@ -9,10 +9,10 @@ namespace BootCamp.Chapter
     {
         public static string Encrypt(string message, byte shift)
         {
-            return string.IsNullOrEmpty(message) ? message : EncryptedMessage(message, shift, true); 
+            return string.IsNullOrEmpty(message) ? message : EncryptedMessage(message, shift); 
         }
 
-        private static string EncryptedMessage(string message, byte shift, bool shiftRight)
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -11,12 +10,46 @@ public static class CaesarCipher
...
+            return new string(asciiChars);
         }
 
         public static string Decrypt(string message, byte shift)
         {
-            return message;
+            return string.IsNullOrEmpty(message) ? message : DecryptedMessage(message, shift);
+        }
+
+        private static string DecryptedMessage(string message, byte shift)
```
Almantask: Decrypt is the same as encrypt (except the shift is different). Please refactor

Almantask (changes requested):

Src/BootCamp.Chapter/CaesarCipher.cs
Tailor-AI: Thanks for the feedback. I added a bool parameter to EncryptedMessage to check if it should encrypt or decrypt.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -10,39 +9,29 @@ public static class CaesarCipher
     {
         public static string Encrypt(string message, byte shift)
         {
-            return string.IsNullOrEmpty(message) ? message : EncryptedMessage(message,shift); 
+            return string.IsNullOrEmpty(message) ? message : EncryptedMessage(message, shift, true); 
         }
 
-        private static string EncryptedMessage(string message, byte shift)
+        private static string EncryptedMessage(string message, byte shift, bool shiftRight)
```
Almantask: No no 😄 
Have both decrypt and encrypt, no need an if statement.
But decrypt should reuse encrypt, that's all.

Tailor-AI: Ok :)
Hopefully now I got it right?

Almantask (approved):
Yup, well done now 😄

Src/BootCamp.Chapter/CaesarCipher.cs
Almantask: Yes

PR #545, opened 2020-09-02
Chapter1/challenges/hangman

Tailor-AI:
First draft of hangman challenge.

### DIFF
--- a/Src/BootCamp.Chapter1/Hangman.cs
+++ b/Src/BootCamp.Chapter1/Hangman.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 
 namespace BootCamp1.Chapter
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter1/Hangman.cs
```diff
@@ -23,18 +24,76 @@ public static class Hangman
         /// <returns>True if game was won or false if game was lost.</returns>
         public static bool Play(int lives, string wordsFile, int difficulty)
         {
-            var wordToGuess = WordsBank.PickRandomWord(@$"Words/{wordsFile}", difficulty);
-            // ToDo: finish the game!
-            var playersGuess = GuessCharacter();
+            if(lives < 1 || difficulty < 3)
+            {
+                throw new SystemException();
```
Almantask, line 29: Throw your own exception, always. This is bad for 2 reasons:
1) there is nothing wrong with system. Basic exception class is better suited.
2) why did it fail? You should give all the details needed (lives and difficulty) to explain what went wrong.

Src/BootCamp.Chapter1/Hangman.cs
```diff
@@ -23,18 +24,76 @@ public static class Hangman
...
+                    lives--;
+                }
+
+                if (lives == 0)
+                {
+                    Console.WriteLine("Too bad! You lost.");
+                    Console.WriteLine($"The correct word was {wordToGuess}.");
+                    return false;
+                }
+
```
Almantask, line 54: no need a blank line

Src/BootCamp.Chapter1/Hangman.cs
```diff
@@ -23,18 +24,76 @@ public static class Hangman
...
+                else if (isVisible.All(x => x == true))
```
Almantask, line 55: Invert condition and use Any(). All will check all entries to be true, Any will check if any is false. Any is more optimal.

Almantask: More optimal, because a single missmatch in condition will stop it.

Almantask (approved):
Yup, looks good

PR #547, opened 2020-09-08
Chapter1/homework/4

EnochRobokoff:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,14 +1,10 @@
...
-using System.Collections.Generic;
-using System.IO;
-using Console = System.Console;
-
-namespace BootCamp.Chapter
+﻿namespace BootCamp.Chapter
 {
     class Program
     {
-        static void Main(string[] args)
```
Almantask, line 10: Just wondering: do you know the impact of removing `args` parameter?

Almantask (approved):

Almantask:
Looks good 👍

Src/BootCamp.Chapter/Program.cs
EnochRobokoff: My understanding of it is that it's used for executing a program from the command line. Beyond that I'm not really sure. Should I not get into a habit of removing this for console applications?

PR #549, opened 2020-09-08
Complete homework 4 for chapter 1.

kungfuphil:
Completed Chapter 1, homework 4.

I did change one of the lines in the Lesson3Tests.cs file, though, because the wording was inconsistent between the first test and last one.

[InlineData(100, 0, "Height cannot be **equal or** less than zero, but was 0.")]
[InlineData(0, 50, "Weight cannot be equal or less than zero, but was 0.")]
[InlineData(-100, 50, "Weight cannot be equal or less than zero, but was -100.")]
~~[InlineData(0, 0, "Weight cannot be equal or less than zero, but was 0.Height cannot be less than zero, but was 0.")]~~
[InlineData(0, 0, "Weight cannot be equal or less than zero, but was 0.Height cannot be **equal or** less than zero, but was 0.")]

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static void Demo()
+        {
+            GetInfoAndPrint();
+            GetInfoAndPrint();
+        }
+
+        static void GetInfoAndPrint()
+        {
+            Console.Write("Name: ");
+            string name = Console.ReadLine();
+
+            var age = PromptForInt("Age: ");
+
+            var weight = PromptForFloat("Weight (in kg): ");
+
+            var height = PromptForFloat("Height (in m): ");
+
+            double bmi = CalculateBmi(weight, height);
+
+            Console.WriteLine($"{ name } is { age } years old, his weight is { weight } kg and his height is { height } m. { name }'s BMI is { bmi }.");
+        }
+
+        public static float CalculateBmi(float weight, float height)
+        {
+            bool isValidWeight = true;
+            bool isValidHeight = true;
+
+            string errorMessage = "{0} cannot be equal or less than zero, but was {1}.";
+
+            if (weight <= 0)
+            {
+                isValidWeight = false;
+            }
+
+            if (height <= 0)
+            {
+                isValidHeight = false;
+            }
+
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,113 @@
...
+            GetInfoAndPrint();
+        }
+
+        static void GetInfoAndPrint()
+        {
+            Console.Write("Name: ");
+            string name = Console.ReadLine();
+
+            var age = PromptForInt("Age: ");
+
```
Almantask, line 20: No need to have a blank line after related pieces of code. I would separate calculation and printing, but the rest seems related.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,113 @@
...
+            var invC = CultureInfo.InvariantCulture;
+
+            Console.WriteLine(message);
+            string input = Console.ReadLine();
+            bool isNumber = float.TryParse(input, NumberStyles.Float, invC, out float number);
+            
+            if (!isNumber && !string.IsNullOrEmpty(input))
+            {
+                Console.Write($"\"{input}\" is not a valid number.");
+                return -1;
```
Almantask, line 78: -1 is a magic arbitary number. Please use a constant for that to explain its' meaning to me and other readers.

Almantask (approved):
Overall looks good!

Almantask:
> Completed Chapter 1, homework 4.
> 
> I did change one of the lines in the Lesson3Tests.cs file, though, because the wording was inconsistent between the first test and last one.
> 
> [InlineData(100, 0, "Height cannot be **equal or** less than zero, but was 0.")]
> [InlineData(0, 50, "Weight cannot be equal or less than zero, but was 0.")]
> [InlineData(-100, 50, "Weight cannot be equal or less than zero, but was -100.")]
> ~[InlineData(0, 0, "Weight cannot be equal or less than zero, but was 0.Height cannot be less than zero, but was 0.")]~
> [InlineData(0, 0, "Weight cannot be equal or less than zero, but was 0.Height cannot be **equal or** less than zero, but was 0.")]

Yeah, it's a good choice and I am aware of inconsistencies- it annoy me as well. Never have the time to actually go and make the changes myself 😢 

Well done for spotting it!

kungfuphil:
Thank you for the comments! I can submit a pull request with just the update to the InlineData where the comment is different, if that works for you.

PR #551, opened 2020-09-11
Completed homework 1and2

bsaranga:
Please review and merge my first homework completion :)

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,39 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            int numPeople = 2;
+
+            while(numPeople > 0)
+            {
+                Console.Write("Enter name: ");
+                string name = Console.ReadLine();
+
+                Console.Write("Enter surname: ");
+                string surname = Console.ReadLine();
+
+                Console.Write("Enter age: ");
+                int age = int.Parse(Console.ReadLine());
+
+                Console.Write("Enter weight (in kg): ");
+                double weight = double.Parse(Console.ReadLine());
+
+                Console.Write("Enter height (in cm): ");
+                double height = double.Parse(Console.ReadLine());
+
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,39 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            int numPeople = 2;
```
Almantask, line 9: Use const modifier.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,39 @@ class Program
...
+
+                BMIPrinter(weight, height);
+
+                numPeople--;
+            }
+
+            Console.ReadKey();
+        }
+
+        static void BMIPrinter(double weight, double height)
```
Almantask: Instead of BMI name Bmi. Pascal case notation.

Almantask: This actually doesn't just print Bmi, it also calculates it.
I would say it's better to split it into two functions (maybe just one)
But don't couple printing with calculating- the two are very different things.

Almantask (changes requested):
All looks great except the tiny nitpick of coupling.

Src/BootCamp.Chapter/Program.cs
bsaranga: Agreed.

bsaranga: Noted

Src/BootCamp.Chapter/Program.cs
bsaranga: When I use the const modifier I cannot decrement it in the while loop.

bsaranga:
Requested changes have been made

Src/BootCamp.Chapter/Program.cs
Almantask: Which is fine- you don't need to decrement numPeople in the loop. You need to increment the index until it reaches numPeople constant.

Almantask (approved):

PR #558, opened 2020-09-17
Chapter 1, Lesson 1 and 2 homework.

vytautas-sablinskas:
I read the information that's inputted about the first person, then make variables for BMI. Once I do that I write in console his name, surname, age, weight (in kg), and height (in cm), then calculate his BMI and print it. Then I repeat the same thing with second person while adding _2 to every variable.


vytautas-sablinskas:
I've updated the code for it to be more readable.

Almantask (approved):
Looks good.
A few nitpicks:
- don't use underscores for naming.
- for local variables use camelCase

PR #560, opened 2020-09-17
Chapter1/homework/4

Nothing-Too-Fancy:


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,110 @@
...
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
+            {
+                Console.WriteLine("What is your first name?");
```
Almantask, line 15: These are not needed. You can use a prompt message for that.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,110 @@
...
+                var name = Console.ReadLine();
```
Almantask, line 16: Use promptX

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Net.NetworkInformation;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
```
Almantask, line 13: Whatever you have in a loop, split in a method of its' own:
`ProcessPersonInfo()`
or smth like that.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            return number;
+        }
+
+        public static float CalculateBMI(float weight, float height)
+        {
+            var bmiError = "Failed calculating BMI. Reason:";
+
+            if (height <= 0 && weight <= 0)
```
Almantask, line 87: If you look closely, you will see repeating yourself:
```
                Console.WriteLine(bmiError);
                // ... 
                return -1;
```
If something changes for the returned value or the way error is printed, you will have to change 3 places.
Instead, you can work with the error messages:
```
string error = "";
if(height <= 0)
{
  error += "....";
}
else if(height <=)
{
  error +="....";
}

if(string.IsNullOrEmpty(error))
{
  Console.WriteLine(error);
  return -1;
}
else
{
  // NOTE: it's cleaner to ALWAYS include the else statement. It adds continuation to the logical flow.
  return (weight / (height * height));
}
```

Almantask (approved):
Well done 😸

PR #561, opened 2020-09-18
Chapter1 Lesson 3 homework.

vytautas-sablinskas:


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
+            {
+                Console.WriteLine("Hello first person, what's your first name?");
```
Almantask, line 16: Extract the body of for loop in a function. And 
A: just call that function twice (loop becomes not needed for this specific case)
B: call the function in the loop (adds clarity)

Almantask (approved):
Looks good 🥇

PR #562, opened 2020-09-18
Chapter1/homework/5

Nothing-Too-Fancy:
You have contradicting test cases for InsertAt_Returns_Array_WithExpectedElements.

Case 2, 3, and 4 have an empty input array. In Case 2 and 4, the expected output is an empty array. In Case 3, the expected output is an array of size 1.

### DIFF
--- a/Src/BootCamp.Chapter1/ArrayOperations.cs
+++ b/Src/BootCamp.Chapter1/ArrayOperations.cs
@@ -189,6 +189,8 @@ namespace BootCamp.Chapter1
                     }
                     return new int[0];
                 }
+                if (index < 0 || index > array.Length)
+                    return new int[0];
             }
             return new int[] { number };
         }
--- a/Src/BootCamp.Chapter1/ArrayOperations.cs
+++ b/Src/BootCamp.Chapter1/ArrayOperations.cs
@@ -1,4 +1,7 @@
-﻿namespace BootCamp.Chapter1
+﻿using System;
+using System.Linq;
+
+namespace BootCamp.Chapter1
 {
     public static class ArrayOperations
     {
@@ -9,7 +12,8 @@
         /// <param name="array">Input array in a random order.</param>
         public static void Sort(int[] array)
         {
-            // ToDo: implement.
+            if (array.Length != 0 || array != null)
+                Array.Sort(array);
         }
 
         /// <summary>
@@ -19,7 +23,8 @@
         /// <param name="array">Input array in a random order.</param>
         public static void Reverse(int[] array)
         {
-            // ToDo: implement.
+            if(array.Length != 0 || array != null)
+            Array.Reverse(array);
         }
 
         /// <summary>
@@ -29,7 +34,15 @@
         /// <returns>A new array with the last element removed. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveLast(int[] array)
         {
-            // ToDo: implement.
+            if (array.Length != 0 && array != null)
+            {
+                var lastRemoved = new int[array.Length - 1];
+                for (int i = 0; i < array.Length - 1; i++)
+                {
+                    lastRemoved[i] = array[i];
+                }
+                return lastRemoved;
+            }
             return array;
         }
 
@@ -39,7 +52,14 @@
         /// <returns>A new array with the first element removed. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveFirst(int[] array)
         {
-            // ToDo: implement.
+            if (array.Length != 0 && array != null)
+            {
+                var firstRemoved = new int[array.Length - 1];
+                for (int i = 1; i < array.Length; i++)
+                {
+                    firstRemoved[i - 1] = array[i];
+                }
+            }
             return array;
         }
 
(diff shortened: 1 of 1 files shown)


Almantask:
> You have contradicting test cases for InsertAt_Returns_Array_WithExpectedElements.
> 
> Case 2, 3, and 4 have an empty input array. In Case 2 and 4, the expected output is an empty array. In Case 3, the expected output is an array of size 1.

No, only case 1 and 2 (when array is out of bounds, we return the same array).
Could you elaborate more?
here are the test cases:
```cs
            // Index out of bounds- returns original array.
            yield return new object[] { Arrays.Empty, 1, Arrays.Empty };
            yield return new object[] { Arrays.Empty, -1, Arrays.Empty };

            // Null or empty- create a new array with that element.
            yield return new object[] { Arrays.Empty, 0, new[] { ElementToBeInserted } };
            yield return new object[] { Arrays.Null, 0, new[] { ElementToBeInserted } };
            
            yield return new object[] { Arrays.TwoElementsDesc, 1, new[] { 1, ElementToBeInserted, 0 } };
```

Nothing-Too-Fancy:
Ah ok you are right. My brain didn't process that index 1 was out of bounds. What is confusing to me still is that it seems an out of bounds index takes precedent over an empty array, in terms of which "error" value should be returned. Is that right? 

The <returns> portion of the function "docs" states:
<returns>A new array with element inserted at a given index. **If an array is empty or null, returns new array with number in it.**</returns>

In order to get the tests to pass, it seems you have to first do a null check, then an index bounds check, and finally the empty array check. Am I on the right track there?

Almantask:
Actually, when it is null, it does nothing - just returns the same array

Nothing-Too-Fancy:
According to the instructions it should return a new array with the desired number and also this test case:
yield return new object[] { Arrays.Null, 0, new[] { ElementToBeInserted } };

Almantask:
> According to the instructions it should return a new array with the desired number and also this test case:
> yield return new object[] { Arrays.Null, 0, new[] { ElementToBeInserted } };

Yes, that's the requirements. I mean it might sound silly, but it's a requirement 😄

Nothing-Too-Fancy:
But what I am saying is that an empty array falls under the same requirements as stated in the function docs, yet the test cases do not reflect that. Maybe I'm not explaining correctly or I am confused...

Nothing-Too-Fancy:
I will modify the code, but I'm just saying the instructions do not state anything about how out of bounds indexes are handled.

Almantask:
> I will modify the code, but I'm just saying the instructions do not state anything about how out of bounds indexes are handled.

I can miss things 

> But what I am saying is that an empty array falls under the same requirements as stated in the function docs, yet the test cases do not reflect that. Maybe I'm not explaining correctly or I am confused...

The requirements might be poorly worded compared to tests. Tests is #1 source of truth.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -9,7 +13,11 @@ public static class ArrayOperations
         /// <param name="array">Input array in a random order.</param>
         public static void Sort(int[] array)
         {
-            // ToDo: implement.
+            if (array != null)
+            {
+                if(array.Length != 0)
+                    Array.Sort(array);
```
Almantask, line 19: Write a custom implementation of this.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -19,7 +27,11 @@ public static void Sort(int[] array)
         /// <param name="array">Input array in a random order.</param>
         public static void Reverse(int[] array)
```
Almantask, line 28: Custom implmentation.

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -29,7 +41,18 @@ public static void Reverse(int[] array)
         /// <returns>A new array with the last element removed. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveLast(int[] array)
```
Almantask, line 42: Reuse removeAt

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -39,7 +62,18 @@ public static int[] RemoveLast(int[] array)
         /// <returns>A new array with the first element removed. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveFirst(int[] array)
```
Almantask, line 63: Reuse removeAt

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -63,8 +119,20 @@ public static int[] RemoveAt(int[] array, int index)
         /// <returns>A new array with element added at a given index. If an array is empty or null, returns new array with number in it.</returns>
         public static int[] InsertFirst(int[] array, int number)
```
Almantask, line 120: Reuse InsertAt

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -75,8 +143,20 @@ public static int[] InsertFirst(int[] array, int number)
         /// <returns>A new array with element added in the end of array. If an array is empty or null, returns new array with number in it.</returns>
         public static int[] InsertLast(int[] array, int number)
```
Almantask, line 144: Reuse InsertAt

Almantask (changes requested):

Nothing-Too-Fancy:
I hope you know I am not trying to criticize, only to understand. I will use the tests themselves as the source of truth from now on.

Almantask:
> I hope you know I am not trying to criticize, only to understand. I will use the tests themselves as the source of truth from now on.

Please do say criticism, that is good, but mentioning it once is enough. I understand that things are imperfect and I am on my best effort of refining it.

Actually, I have paused the lessons a week ago and have been going through all the lessons from day 1 and updating the material, double-checking tests, adding wikis if there weren't any.

Criticism leads to perfection. Please don't stop that.

Nothing-Too-Fancy:
Will do. Appreciate the work you've done here.

PR #563, opened 2020-09-19
Chapter 1, Lesson 4 homework.

vytautas-sablinskas:
Checker in VS doesn't work, so I'll check it by using PR.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -9,22 +9,7 @@ namespace BootCamp.Chapter
     {
         public static void Demo()
         {
-            for (int i = 0; i < 2; i++)
-            {
-                Console.WriteLine("Hello first person, what's your first name?");
-                string firstName = Console.ReadLine();
-                Console.WriteLine("What's your second name?");
-                string secondName = Console.ReadLine();
-                Console.WriteLine("What's your age?");
-                int age = int.Parse(Console.ReadLine());
-                Console.WriteLine("What's your weight? (In kg)");
-                float weight = float.Parse(Console.ReadLine());
-                Console.WriteLine("What's your height? (In cm)");
-                float height = float.Parse(Console.ReadLine());
-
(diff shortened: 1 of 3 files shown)


vytautas-sablinskas:
I have no clue honestly what to do right now, probably will watch Lesson 5 and then try to comeback because I'm stuck on this for long time already and Float checks don't show if they're correct or wrong.

vytautas-sablinskas:
Done all checks, I think it should be alright now.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson4
+    {
+        public static void Demo()
+        {
+            for (int i = 0; i < 2; i++)
+            {
```
Almantask: The body of for loop could be moved into a function. That would leave you with two options then:
A- just call the same function twice
B- do what you do, but in a body call a function instead.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,106 @@
...
+                return -1;
+            }
+            return number;
+        }
+        public static float CalculateBmi(float weight, float height)
+        {
+            bool isValidWeight = true;
+            bool isValidHeight = true;
+
+            if (height <= 0)
```
Almantask: Can be simplified:
```cs
string error = "";
if(weight <= 0)
{
 error = ...;
}
else if(height <= 0)
{
  error += "";
}
if(string.IsNullOrEmpty(error))
{
  print(error);
  return -1;
}
else
{
  return bmi;
}
```

Almantask (changes requested):
A bit overly complex validation

vytautas-sablinskas:
Okay, so I redid it a bit. Now I added another function AskingInformation and called it in Demo function. 
Then I tried simplifying this thing, but not sure if I got it right. It's pretty hard to understand this, but I hope I can get better at it.

Almantask (approved):
Looks better 😄

PR #564, opened 2020-09-19
Chapter 2/Homework 3

Tailor-AI:
First commit for homework 3.

### DIFF
--- a/Src/BootCamp.Chapter/Computer/DesktopComputer.cs
+++ b/Src/BootCamp.Chapter/Computer/DesktopComputer.cs
@@ -2,6 +2,15 @@
 {
     public class DesktopComputer
     {
+        public DesktopComputer()
+        {
+            _body = new Body();
+            _ram = new Ram();
+            _cpu = new Cpu();
+            _gpu = new Gpu();
+            _hard = new HardDisk();
+            _motherboard = new Motherboard();
+        }
         private Body _body;
         public Body GetBody()
         {
@@ -36,6 +45,11 @@
         public Motherboard GetMotherboard()
         {
             return _motherboard;
+        }
+
+        public virtual string CompletedAssembly()
+        {
+            return "Assembly of computer complete.";
         }
     }
 }
--- a/Src/BootCamp.Chapter/Computer/DesktopComputer.cs
+++ b/Src/BootCamp.Chapter/Computer/DesktopComputer.cs
@@ -2,25 +2,26 @@
 {
     public class DesktopComputer
     {
-        public DesktopComputer()
-        {
-            _body = new Body();
-            _ram = new Ram();
-            _cpu = new Cpu();
-            _gpu = new Gpu();
-            _hard = new HardDisk();
-            _motherboard = new Motherboard();
-        }
         private Body _body;
         public Body GetBody()
         {
             return _body;
         }
         
+        public void SetBody(Body body)
+        {
+            _body = body;
+        }
+
         private Ram _ram;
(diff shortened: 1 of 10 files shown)


Src/BootCamp.Chapter/Computer/DesktopComputer.cs
```diff
@@ -2,6 +2,15 @@
 {
     public class DesktopComputer
     {
+        public DesktopComputer()
+        {
+            _body = new Body();
```
Almantask: Desktop computer has the same parts both for mac and for windows.
Each factory should be making branded parts and INSTALL those into a new computer.

Almantask (changes requested):

Src/BootCamp.Chapter/Computer/DesktopComputer.cs
Tailor-AI: I'm not sure what you mean by "making branded parts" but now I made new parts in each factory and "installed" them in the assemble function.

Src/BootCamp.Chapter/MacFactory.cs
```diff
@@ -1,26 +1,27 @@
...
 namespace BootCamp.Chapter
 {
     public class MacFactory : DesktopComputer
     {
-        public MacFactory()
-        {
-            Assemble();
-            Console.WriteLine(CompletedAssembly());
-        }
+        Body macBody = new Body();
```
Almantask: This means that all computers from this factory will have the same, shared Body, Ram, Cpu, etc. It looks the same, yes, but it's a DIFFERENT INSTANCE

Src/BootCamp.Chapter/MsFactory.cs
```diff
@@ -1,26 +1,27 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using BootCamp.Chapter.Computer;
+﻿using BootCamp.Chapter.Computer;
 
 namespace BootCamp.Chapter
 {
     public class MsFactory : DesktopComputer
```
Almantask: What dos Ms factory have different compared to Win factory? NOthing. Please make sure that Ms factory creates MsRam, MsBody, MsGpu, etc. And Mac creates MacRam, MacCpu, MacBody, etc.

Almantask (changes requested):

Src/BootCamp.Chapter/Computer/DesktopComputer.cs
Almantask: What dos Ms factory have different compared to Win factory? NOthing. Please make sure that Ms factory creates MsRam, MsBody, MsGpu, etc. And Mac creates MacRam, MacCpu, MacBody, etc.

Tailor-AI: Added Ms parts and Mac parts and moved the creation of the parts under the assemble function.

Src/BootCamp.Chapter/MacFactory.cs
Tailor-AI: Ok. Now when I moved these down to the assemble function there should be one instance for every computer?

Src/BootCamp.Chapter/MsFactory.cs
Tailor-AI: I removed the inheritance of DesktopComputer because these are not related I think?

Almantask (approved):
Looks good now

PR #566, opened 2020-09-20
Homework 1 and 2

EMonfort:
Added the code for homework 1 and 2.


Almantask (approved):
Look good.
Nit: names like BMI should still be camelCases (bmi).

PR #569, opened 2020-10-03
Chapter1/homework/7

Nothing-Too-Fancy:
Chapter 1 Homework 7 implementation

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -200,6 +200,7 @@ namespace BootCamp.Chapter
             CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
             var format = (NumberFormatInfo)CultureInfo.CurrentCulture.NumberFormat.Clone();
             format.CurrencyGroupSeparator = string.Empty;
+            format.CurrencyNegativePattern = 1;
             var valueStr = string.Format(format, "{0:C0}", value);
 
             return valueStr;
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -15,7 +15,7 @@ namespace BootCamp.Chapter
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            if (peopleAndBalances == null && peopleAndBalances.Length == 0) return defaultReturn;
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0) return defaultReturn;
 
             var highNames = new List<string>();
             int highBal = 0;
@@ -108,7 +108,7 @@ namespace BootCamp.Chapter
         /// </summary>
         public static string FindRichestPerson(string[] peopleAndBalances)
         {
-            if (peopleAndBalances == null && peopleAndBalances.Length == 0) return defaultReturn;
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0) return defaultReturn;
 
             var richNames = new List<string>();
             int biggestBalance = 0;
@@ -151,7 +151,7 @@ namespace BootCamp.Chapter
         /// </summary>
         public static string FindMostPoorPerson(string[] peopleAndBalances)
         {
-            if (peopleAndBalances == null && peopleAndBalances.Length == 0) return defaultReturn;
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0) return defaultReturn;
 
             var poorNames = new List<string>();
             int lowestBalance = int.MaxValue;
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -1,4 +1,9 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+
+namespace BootCamp.Chapter
 {
     public static class BalanceStats
     {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,233 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+
+namespace BootCamp.Chapter
 {
     public static class BalanceStats
     {
+        static string defaultReturn = "N/A.";
```
Almantask: instead of static, use const.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,233 @@
...
+                    else if (balance == highBal)
+                    {
+                        highNames.Add(name);
+                    }
+                }
+
+                StringBuilder sb = new StringBuilder();
+                var highBalStr = FormatValue(highBal);
+
+                BuildNamesString(sb, highNames);
```
Almantask: Slightly mixed responsibilities. Instead of passing a string builder inside that method- make SB just a detail. Return string instead. So:
instead of:
```cs
StringBuilder sb = new StringBuilder();
                var highBalStr = FormatValue(highBal);

                BuildNamesString(sb, highNames);
                sb.Append($" had the most money ever. {highBalStr}.");

                return sb.ToString();
```
```cs
return BuildNames(highBalstr, highNames);
```

Almantask: Move the selected part to BuildNamesString method, because that's what it does

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +38,51 @@ public static class TextTable
         /// </summary>
         public static string Build(string message, int padding)
         {
+            if (!string.IsNullOrEmpty(message))
```
Almantask: Prefer to state, rather than negate.
Prefer to exit/fail early. 
So instead:
```cs
if (string.IsNullOrEmpty(message)) return String.Empty;

                var sb = new StringBuilder();
                var messages = message.Split(Environment.NewLine);
                var longestString = "";
                for (int i = 0; i < messages.Length; i++)
                {
                    if (messages[i].Length > longestString.Length)
                        longestString = messages[i];
                }
                var whiteSpace = "";
                var length = longestString.Length + padding * 2;
                sb.Append($"+{whiteSpace.PadRight(length, '-')}+{Environment.NewLine}");
                BuildWhitespace(sb, whiteSpace, padding, length);
                BuildMessage(sb, messages, padding, longestString.Length);
                BuildWhitespace(sb, whiteSpace, padding, length);
                sb.Append($"+{whiteSpace.PadRight(length, '-')}+{Environment.NewLine}");

                return sb.ToString();
```

Almantask: It's better, because you no longer have code nested in an if statement, you take boring code and run it first, the interesting part is not mixed with the boring part.

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +38,51 @@ public static class TextTable
...
+                var messages = message.Split(Environment.NewLine);
+                var longestString = "";
+                for (int i = 0; i < messages.Length; i++)
+                {
+                    if (messages[i].Length > longestString.Length)
+                        longestString = messages[i];
+                }
+                var whiteSpace = "";
+                var length = longestString.Length + padding * 2;
+                sb.Append($"+{whiteSpace.PadRight(length, '-')}+{Environment.NewLine}");
```
Almantask: Use AppendLine instead, won't need NewLine.

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +38,51 @@ public static class TextTable
...
+            {
+                var sb = new StringBuilder();
+                var messages = message.Split(Environment.NewLine);
+                var longestString = "";
+                for (int i = 0; i < messages.Length; i++)
+                {
+                    if (messages[i].Length > longestString.Length)
+                        longestString = messages[i];
+                }
+                var whiteSpace = "";
```
Almantask: Not really a white space.

Src/BootCamp.Chapter/TextTable.cs
Almantask: use String.Empty instead your variable.

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +38,51 @@ public static class TextTable
...
+                var length = longestString.Length + padding * 2;
+                sb.Append($"+{whiteSpace.PadRight(length, '-')}+{Environment.NewLine}");
+                BuildWhitespace(sb, whiteSpace, padding, length);
```
Almantask: no need to pass whitespace

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,14 +1,21 @@
...
+
+namespace BootCamp.Chapter
 {
     class Program
     {
         static void Main(string[] args)
         {
+            
             // Print each of the statistical output using Text Table with padding 3:
             // - FindHighestBalanceEver
```
Almantask: Comments are no longer needed- code speaks for itself.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,233 @@
...
+            return defaultReturn;
         }
 
         /// <summary>
         /// Return name and loss of a person with a biggest loss (balance change negative).
         /// </summary>
         public static string FindPersonWithBiggestLoss(string[] peopleAndBalances)
         {
-            return "";
+            if (peopleAndBalances != null && peopleAndBalances.Length != 0)
```
Almantask: Invert if and return default here- will reduce nesting.

Almantask (changes requested):
Overall looks pretty good, but due to that, I added a few remarks to make it better. 
Three major points:
- instead of "" use string.Empty (more readable, the constant is there for you)
- fail/return eaarly- do validations FIRST, because it makes the reader focus to what matters instead of mixing it with validation
- be positive instead of negative. Prefer conditions that do not have not (!) in them. Easier to read.

Src/BootCamp.Chapter/BalanceStats.cs
Nothing-Too-Fancy: changed variable to const

Src/BootCamp.Chapter/TextTable.cs
Nothing-Too-Fancy: inverted the if statement to reflect this

Src/BootCamp.Chapter/TextTable.cs
Nothing-Too-Fancy: Changed to AppendLine

Src/BootCamp.Chapter/TextTable.cs
Nothing-Too-Fancy: I have removed this variable

Src/BootCamp.Chapter/TextTable.cs
Nothing-Too-Fancy: removed whiteSpace variable and have used String.Empty

Src/BootCamp.Chapter/TextTable.cs
Nothing-Too-Fancy: removed this parameter

Src/BootCamp.Chapter/Program.cs
Nothing-Too-Fancy: removed comments from main

Src/BootCamp.Chapter/BalanceStats.cs
Nothing-Too-Fancy: Inverted all if statements

Src/BootCamp.Chapter/BalanceStats.cs
Nothing-Too-Fancy: How can I account for the variable strings within the method? (i.e. "had the most money ever", "had the least money ever", etc). Should I pass this as another parameter? Same goes for the pluralOrSingle logic.

Almantask: Yes, pass them as in my example. It's better than passing StringBuilder. StringBuilder is a detail- it's optional for the overall working of the function. However the essence of that function is work with those two strings variables
\

Nothing-Too-Fancy: Ok. I've moved everything into the BuildNames method.

PR #571, opened 2020-10-04
Finishing first homework

chunggjo:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,39 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            var name = "Joe";
```
Almantask: Read input from conosle. Refer to `Console.ReadLine();`, int.Parse, float.Parse

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
chunggjo: I did not read instructions carefully, should be fixed now

Almantask (approved):
Looks good now

PR #572, opened 2020-10-04
Chapter4/homework/3

HellHunterMax:
https://leetcode.com/problems/max-increase-to-keep-city-skyline/
So I picked this problem to solve sounded fun and proved fun.
TDD is very hard for me, my thinking automatically goes the other way around. 
I don't have a problem making tests, but test first and then make something harder.

I tried to make use of SOLID as much as possible. If you could give some hints where to improve on this that would be great.
probably made some naming mistakes. should have used more interfaces. because right now its tightly coupled. But I think with such a small project its hard to see where to use it and where not.


Src/BootCamp.Chapter/Exceptions/BuildingException.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace BootCamp.Chapter.Exceptions
+{
+    [Serializable]
+    public class BuildingException : Exception
```
Almantask: Is Building an exception? There is nothing exceptional about the structure. please rename better :D

Src/BootCamp.Chapter/MaxCityRise.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace BootCamp.Chapter
+{
+    public static class MaxCityRise
+    {
+        public static int GetMaxCityRise(City city)
+        {
+            return SumHeightsTrueIsMaxHeightFalseIsOriginal(true, city) - SumHeightsTrueIsMaxHeightFalseIsOriginal(false, city);
+
+        }
+        private static int SumHeightsTrueIsMaxHeightFalseIsOriginal(bool isMax, City city)
```
Almantask: No idea what this function does. Please rename better.

Tests/BootCamp.Chapter.Tests/CityTest.cs
```diff
@@ -0,0 +1,45 @@
...
+using Moq;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class CityTest
+    {
+        [Theory]
+        [InlineData( 1, -1)]
+        [InlineData(1, 51)]
+        public void City_Given_BadBuildings_Throws_BuildingException(int citySize, int buildingHeight)
```
Almantask: What is a bad building?

Tests/BootCamp.Chapter.Tests/CityTest.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            int[][] buildings = Builders.BuildBuildingsArray(citySize, buildingHeight);
+            City city;
```
Almantask: No need

Tests/BootCamp.Chapter.Tests/CityTest.cs
```diff
@@ -0,0 +1,45 @@
...
+using BootCamp.Chapter.Exceptions;
+using BootCamp.Chapter.Tests.Public;
+using Moq;
+
+namespace BootCamp.Chapter.Tests
+{
+    public class CityTest
+    {
+        [Theory]
+        [InlineData( 1, -1)]
```
Almantask, line 13: Maybe underground height is ok?

Tests/BootCamp.Chapter.Tests/Public/Builders.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter.Tests.Public
+{
+    public static class Builders
+    {
+        public static int[][] BuildBuildingsArray(int citySize, int buildingHeight)
```
Almantask: BuildBuildings is enough. Appending the type to the name does not add value.

Almantask (changes requested):
I don't really understand the logic or tests. Please try to rename things following the patterns:
Verbs for methods
Nouns for variables/classes

Add some comments.

Almantask:
Also please describe what you did.

Tests/BootCamp.Chapter.Tests/CityTest.cs
HellHunterMax: Yea i agree but the assignment said:
All heights grid[i][j] are in the range [0, 100].
so that's why the test is there and that's why i added the exception.

Src/BootCamp.Chapter/MaxCityRise.cs
HellHunterMax: yea that was real bad naming >_<

HellHunterMax:
I know my Way of thinking is not TDD and its hard for me to think that way. Here is how i went through it:
I started off thinking i need a Object to hold Buildings and a SkyLine.
so i started creating a City and a SkyLine Model.
Then i finished the SkyLine model and following tests to see if it works.
Then the Calculation for Maximum a city can grow without changing the SkyLine followed by the tests.
Finally added the last class to calculate the maximum amount a city could grow followed by the tests.

So no TDD used TBH. I should have Created the tests sooner and i do see that now. I should have created the tests as soon as I knew what I was going to return. that's what I'll do next time.

Any comments on SOLID ?

Almantask:
> I know my Way of thinking is not TDD and its hard for me to think that way. Here is how i went through it:
> I started off thinking i need a Object to hold Buildings and a SkyLine.
> so i started creating a City and a SkyLine Model.
> Then i finished the SkyLine model and following tests to see if it works.
> Then the Calculation for Maximum a city can grow without changing the SkyLine followed by the tests.
> Finally added the last class to calculate the maximum amount a city could grow followed by the tests.
> 
> So no TDD used TBH. I should have Created the tests sooner and i do see that now. I should have created the tests as soon as I knew what I was going to return. that's what I'll do next time.
> 
> Any comments on SOLID ?

A better way of thinking would be to pretend you don't know anything and just:
I am required A so I will do write a test that would verify if it is met.
I know that building must be in a range of X and Y. I know that if a city has any invalid building, creating it fails.
What is a city?
- It's a collection of buildings
What is a building?
- It's just an int.
How do I feed buildings to city?
- I pass them via ctor.

Based on those answers:
```
int[] buildings = ...;

Action action = () => new City(building);

action.Should().Throw...
```

City does not exist yet. Proceed with creating a class City, a ctor that takes int[].
Run tests. 
Test will fail because there is no validation yet. Add validaiton.
Run tests.
Tests will pass.
Think about edge cases. Less than minum, more than minimum.
Two cases, same invocation- means you need to move it to parameters.
Add another case.
Run tests.
Fix or be done.
Refactor.
Run tests.
Fix or be done.

HellHunterMax:
Thanks for explaining! seems doable.

Almantask:
Done 👍

PR #575, opened 2020-10-05
Chapter 2/Homework 4

Tailor-AI:
First draft of homework 4.

### DIFF
--- a/Src/BootCamp.Chapter/GlobalSettings.cs
+++ b/Src/BootCamp.Chapter/GlobalSettings.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public static class GlobalSettings
+    {
+        public static string filePath = Directory.GetCurrentDirectory();
+
+        public static ILogger Logger(LogType logType)
+        {
+            switch (logType)
+            {
+                case LogType.ConsoleLog:
+                    return new ConsoleLogger();
+                case LogType.FileLog:
+                    return new FileLogger();
+                default:
+                    return new ConsoleLogger();
+            }
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


Src/BootCamp.Chapter/GlobalSettings.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public static class GlobalSettings
+    {
+        public static string filePath = Directory.GetCurrentDirectory();
+
+        public static ILogger Logger(LogType logType)
```
Almantask: Methods should be verbs. What do you do with logger?
- get logger?
- create logger?
- load logger?
?

Src/BootCamp.Chapter/Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.IO;
+
+namespace BootCamp.Chapter
+{
+    public class FileLogger : ILogger
+    {
+        public void LogMessage(DateTime dateTime, string message)
```
Almantask: DateTime is not really needed.
It doesn't make sense to log any other time than DateTime.Now.
If you want date to sometimes be there, sometimes not, then just pass it use another method or configuration. I imagine that you either want dates always or never, never a mix.

Almantask (changes requested):

Src/BootCamp.Chapter/GlobalSettings.cs
Tailor-AI: Correct, renamed this to CreateLog.

Src/BootCamp.Chapter/Logger/FileLogger.cs
Tailor-AI: Removed the parameter and put the DateTime.Now in the message itself.

Almantask (approved):

PR #576, opened 2020-10-09
Chapter1/homework/4

chunggjo:


Almantask (approved):
Looks good. 
Small nitpick: 
The constant values returned across a few different functions- you can save them into const int invalidNumber= -1; or similar.

PR #581, opened 2020-10-19
I am sending a failed version, could you, please, tell me, where I am…

JohnHNewland:
… wrong?

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -4,11 +4,68 @@ using System.IO;
 using Console = System.Console;
 
 namespace BootCamp.Chapter
-{
+{ 
     class Program
     {
         static void Main(string[] args)
         {
+            string nameMessage = "May I have your name?";
+            string surnameMessage = "May I have your surname?";
+            string ageMessage = "May I have your age?";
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -4,11 +4,68 @@
...
+        }
+        public static float CalculateBMI(float weight, float height)
+        {
+            float bmi = weight / (height * height);
+            return (float)Math.Round(bmi, 1);
+        }
+        public static int AskingAndConvertingINT(string message)
+        {
+            Console.WriteLine(message);
+            int integer = Int32.TryParse(Console.ReadLine(), 10);
```
Almantask, line 55: If you looked at the error messages:
image
It shows what's wrong.

And indeed, that's not how TryParse worse. Please note, that TryParse returns true or false (was parse successful).
It also returns the parsed value, but that is done using the `out` keyword.

What you actually need is this:
```
bool isNumber = Int32.TryParse(Console.ReadLine(), out int number);
```
or if you trust that it's always going to be a numberr, you can simply do this:
```
int number = int.Parse(Console.ReadLine());
```

Almantask (changes requested):

PR #582, opened 2020-10-19
Chapter1/homework/3

Abubakar-97:
I tried to solve this but I am not sure. Would love to find feedback on this.

another thing I wanted to ask is if there is a better way to implement this.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+
+        }
+
+        
+    }
+}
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static string NameofPerson()
```
Almantask: Functions should be verbs. You should specify an action in name. Get/Prompt/Read, ... and only then what- NameOfPerson.

Almantask (changes requested):
Please change the target branch. Should be homework/3, not master.
image

Src/BootCamp.Chapter/Lesson3.cs
Abubakar-97: I will do it right away.

Sincerely,
Memoire

Almantask (approved):

PR #589, opened 2020-11-01
Homework 1 & 2 solution

BNAndras:
Homework 1 & 2 solution


Almantask (approved):
Quite a unique approach, I like it. Well done splitting it into functions like so.

PR #591, opened 2020-11-02
Chapter1/homework/1 and2

littleNoodle:
homework for lesson 1 and lesson 2 in chapter 1


Almantask (changes requested):
Classes are a bit of an overkill here. Also, please don';t hardcode the person stats- it should all come from Console.Readline

Almantask (approved):
Looks good, but next time please read the task a bit more closely- you need to do the same for two people

PR #592, opened 2020-11-02
commit for chapter 1 lesson 3

littleNoodle:
homework for chapter1 lesson 3


Almantask (approved):
Looks good, but you should apply the same function twice (for 2 people)

PR #593, opened 2020-11-03
Chapter1/Homework/4

littleNoodle:


littleNoodle:
Hello, I feel a little confused about the test principle, just like the expected output is "Failed calculating BMI. Reason",and the actual output is "Failed calculating BMI.Reason", only one whitespace difference and then test result is NG.

Almantask (approved):
Looks good

Almantask:
> Hello, I feel a little confused about the test principle, just like the expected output is "Failed calculating BMI. Reason",and the actual output is "Failed calculating BMI.Reason", only one whitespace difference and then test result is NG.

Yeah, some tests are like that and it's pretty too strict on my end. You can ignore that test. Overall the code works and I think is good enough- carry on

PR #595, opened 2020-11-04
Lesson 4 Chapter 1

AdrianPaoletti:
I don't think is efficient the BMI if statements, I think there is a way to make it easier, shorter and with less code.
The way I sincronized the file I am not sure that I did it in the correct way.
Thank you in advance !

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -8,8 +8,7 @@ namespace BootCamp.Chapter
 {
     class Lesson3
     {
-        //Return Str Function
-        public static string printStr(string message, string variable)
+        public static string promptStr(string message, string variable)
         {
             Console.WriteLine(message);
             variable = Console.ReadLine();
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+            var isNumber = int.TryParse(input, out var age1);
+            if (!isNumber)
+            {
+                Console.WriteLine(input + " is not a valid number");
+                return -1;
+            }
+            return variable = age1;
+        }
+
+        //Return Float Function
```
Almantask: Such comments don't add any more information than the code already says. Can be removed.
It's okay to have such comments initially, as placeholders, but they should be pushed for publication

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Globalization;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        //Return Str Function
+        public static string printStr(string message, string variable)
```
Almantask: You don't print a string, you prompt to get a string.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (string.IsNullOrEmpty(variable)) 
+            {
+                Console.WriteLine("Name cannot be empty");
+                return "-";                  
+            }
+            return variable;
+        }
+
+        //Return Int Function
+        public static int printInt(string message, int variable)
```
Almantask: Prompt, not print

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+        //Return Float Function
+        public static float printFloat(string message, float variable)
+        {
+            Console.WriteLine(message);
+            variable = float.Parse(Console.ReadLine(), CultureInfo.InvariantCulture);
+            return variable;
+        }
+
+        //Return BMI Function
+        public static float printBMI(float var1, float var2)
```
Almantask: You don't print bmi, you calculate it

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,85 @@
...
+            string errorHeight = " Height cannot be equal or less than zero, but was " + var2;
+
+            bool error1 = var1 <= 0 && var2 <= 0;
+            if (error1) 
+            {
+                Console.WriteLine(always + errorWeight + errorHeight);
+                return -1;
+            }              
+                        
+            bool error2 = var1 <= 0;
```
Almantask: please name it better, what exactly is error 1-3?

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +9,22 @@ class Program
     {
         static void Main(string[] args)
         {
+            // Variables
+            string nameSurname = " ";
```
Almantask: don't declare all variables in one place- declare them as close as possible to where they are used

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,22 +9,18 @@ class Program
     {
         static void Main(string[] args)
         {
-            // Variables
             string nameSurname = " ";
+            nameSurname = Lesson3.promptStr("Enter your name", nameSurname);
             int age = 0;
```
Almantask: default values are not needed. Just assign them directly

Almantask (changes requested):

Almantask (approved):
Looks good now

PR #596, opened 2020-11-08
Chapter 1 Homework 3 Solution

BNAndras:
After seeing the note about using CultureInfo.InvariantCulture for reading floats, I also played with the float formatting on weight and height. Did my approach make sense?


Almantask (approved):
Yes, the approach is fine. If there were more of those formatting, you could have set a global culture and reset it back in the end.

PR #600, opened 2020-11-11
Chapter1/homework/4

BNAndras:


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +15,92 @@ namespace BootCamp.Chapter
     /// </summary>
     public static class Checks
     {
-        public static int PromptInt(string message)
+        public const int IntegerErrorCode = -1;
+        public const float FloatErrorCode = -1.0f;
+        public const string StringErrorCode = "-";
```
Almantask, line 20: these constants can be kep private.

Almantask: At least in this case.

Almantask (approved):
Good code, well done

PR #601, opened 2020-11-11
Chapter1/Homework/7

littleNoodle:
Commit


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +10,255 @@ public static class BalanceStats
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0)
+            {
+                return "N/A.";
```
Almantask, line 15: "N/A" is used in multiple places. What you did is considered hardcoding- because you used a magic value. Explain what value is used by using a named constant at thew class level

Almantask (approved):
Overall looks okay

PR #604, opened 2020-11-12
Chapter 3/Homework 4

Tailor-AI:
My first try on this homework. 
My understanding is that ISchool has both in and outputs so no in or out?
ITeacher has only output so this should have out?


Almantask:
> My first try on this homework.
> My understanding is that ISchool has both in and outputs so no in or out?
> ITeacher has only output so this should have out?

Yes, that is correct.

Src/BootCamp.Chapter/Teachers/UniversityTeacher.cs
```diff
@@ -0,0 +1,25 @@
+﻿using BootCamp.Chapter.Interface;
+using System;
+
+namespace BootCamp.Chapter.Teachers
+{
+    public class UniversityTeacher : ITeacher<ISubject>
+    {
+        ISubject subject;
```
Almantask, line 8: Add access modifier explicitly. Also make readonly

Almantask (approved):
Looks good overall 🛩️

PR #605, opened 2020-11-15
Chapter1/homework/8

littleNoodle:


Almantask:
This homework is tricky to check, because there is a to make it good, a lot of the features will come in Chapter 3. What you have, I think, looks good enough 😄

PR #606, opened 2020-11-15
Chapter1/homework/3

Camper94:
my homework 3


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,26 @@ public static class Checks
...
-            return 0;
+            return Lesson3.PromptInt(message);
+      
         }
 
         public static string PromptString(string message)
         {
             // To do: call your implementation. 
-            return "";
+            return Lesson3.PromptString(message);
```
James231, line 26: You could add the implementation here, you didn't need to put it in a separate class

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
```
James231, line 9: empty method - not needed

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            
+        }
+        public static int PromptInt(string message)
+        {
+            int Newmessage;
+            Console.WriteLine(message);
+            Newmessage = int.Parse(Console.ReadLine());
```
James231, line 17: use int.TryParse not int.Parse - just in case the user provides a string which is not a number

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+            
+        }
+        public static int PromptInt(string message)
+        {
+            int Newmessage;
```
James231, line 15: You can initialize and declare at the same time. Single line:
`int newMessage = int.TryParse( ...`

James231: And use camelCase for variable names. So it should be newMessage, not Newmessage

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
...
+            name = Console.ReadLine();
+            return name;
+        }
+
+
+        public static float PromptFloat(string message)
+        {
+            float height;
+            Console.WriteLine(message);
+            height = float.Parse(Console.ReadLine());
```
James231, line 34: again `float.TryParse` not `float.Parse`

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
...
+            return height;
+        }
+
+
+        public static float CalculateBmi(float weight, float height)
+        {
+            float Bmi;
+            Bmi = weight /(height*height);
```
James231, line 42: use consistent spacing between operators `weight / (height * height)`

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
```
James231, line 2: Unnecessary reference

Src/BootCamp.Chapter/Checks.cs
Camper94: this what was asked in the homework , create a class called Lesson3 and after that call it from check.cs can i insert a suggestion when i do the modification that u asked instead of pull request ?

James231: Tbh, I didn't read the homework.
In which case, what you have done here is fine 👍

Almantask:
Merging to a wrong branch.

PR #609, opened 2020-11-17
Chapter1/homework/10

littleNoodle:


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -11,12 +11,30 @@ public static class CaesarCipher
...
+            byte[] bytes = Encoding.ASCII.GetBytes(message);
+            for(int i = 0; i < bytes.Length; i++)
+            {
+                bytes[i] = Convert.ToByte((bytes[i] + shift)% 128);
+            }
+            string result = Encoding.ASCII.GetString(bytes);
+            return result;
         }
 
         public static string Decrypt(string message, byte shift)
```
Almantask, line 26: Decrypt should call encrypt, but with the opposite (negative) shift.

Almantask (changes requested):

PR #612, opened 2020-11-17
Chapter1/homework/3

Camper94:
try


Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +9,23 @@ class Program
     {
         static void Main(string[] args)
         {
+            // Variables
+            string nameSurname = " ";
+            int age = 0;
+            float weight = 0;
+            float height = 0;
```
Almantask, line 16: Define varaible right next to where they are used, not in one place.

Almantask (approved):

PR #614, opened 2020-11-19
Chapter2/homework/1

littleNoodle:
I refactor the constructors of InvalidBalancesException.


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -1,39 +1,409 @@
...
+                        name = tempArray[0];
+                    }
+                    else if (maxBalance - minBalance == diffBalance)
+                    {
+                        name += " " + tempArray[0];
+                    }
+                }
+                else
+                {
+                    return negative;
```
Almantask, line 173: What is negative? I think it would make more sense if you returned `Invalid`.

Almantask (approved):

PR #622, opened 2020-12-02
Chapter3/homework/1

littleNoodle:
use property and List<T>

### DIFF
--- a/Src/BootCamp.Chapter/Items/Chestpiece.cs
+++ b/Src/BootCamp.Chapter/Items/Chestpiece.cs
@@ -12,5 +12,27 @@
             _price = price;
             _weight = weight;
         }
+        public string Name
+        {
+            get => _name;
+            set => _name = value;
+        }
+        public decimal Price
+        {
+            get => _price;
+            set => _price = value;
+        }
+        public float Weight
+        {
+            get => _weight;
+            set => _weight = value;
+        }
+
+        private float _defense;
+        public float Defense
+        {
+            get => _defense;
+            set => _defense = value;
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


Src/BootCamp.Chapter/Items/Chestpiece.cs
```diff
@@ -6,11 +6,34 @@ public class Chestpiece
         private decimal _price;
         private float _weight;
 
-        public Chestpiece(string name, decimal price, float weight)
+        public Chestpiece(string name, decimal price, float weight, float defense)
```
norarrsgd, line 9: I think this class can be simplified further. Technically your approach is still correct. But I don't see any valid reason to keep it this way. You can try

```csharp
public class Chestpiece
{
    public string Name { get; set; }
    public decimal Price { get; set; }
    public float Weigh { get; set; }
    public float Defense { get; set; }
}
```

Then in your calling method/code you can instantiate it like so
```csharp
Chestpiece chestPiece = new Chestpiece
{
    Name = "Some Name",
    Price = 10.5,
    Weight = 4.5,
    Defense = 1.2
};
```

I know I only commented on this file, but I think this applies to most of your classes ✌🏻

LMK if you have further questions

norarrsgd (changes requested):
✌🏻

Src/BootCamp.Chapter/Items/Chestpiece.cs
littleNoodle: Thanks for your kind advice, I will modify the constructor function, O(∩_∩)O~

Almantask (approved):
Looks good!

PR #626, opened 2020-12-18
Chapter1/homework/1 and2 - Homework Assignment

Joshhortt:
1. Print to console all the info based on the example requested, using **concatenation**. 
    At first I declared some variables type: **double**, **int,** and **string**.

2. Body Mass Index Calculation is based on weight and height.  (BMI)=weight/(height*height), *10000
    where weight is in kilograms and height in centimeters. The program takes weight and height as 
    input and displays BMI using a method called **Math.Round** with an value to the specified 
    number of **decimal digits of 2**.

3. Print to console all the personal details requested based on any **users input**, using the method 
    **Convert.ToInt32** to convert specified values to signed integers. 
    Use of **concatenation** to print results to the console. Declared some type **double** variables.

3. Print to console all the personal details requested for Body Mass Index using the same calculations 
    in point number 2, but with request of **user input** using the method **double.Parse** that 
    converts the string representation of a number, to its double-precision floating-point number.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,7 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,7 +7,7 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
             Console.WriteLine("Hello World!");
-            Console.WriteLine("Hello World!");
+            Console.WriteLine("VS code is limited!");
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -7,6 +7,11 @@ namespace BootCamp.Chapter
         static void Main(string[] args)
         {
 			Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hello World!");
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,80 @@
...
+			   his weight is 50 kg and his height is 156.5 cm.
+			 */
+
+			double height = 156.5;
+			int weight = 50;
+			int age = 19;
+			string surName = "Jefferson";
+			string name = "Tom";
+
+			Console.WriteLine(name + " " + surName + " you are " + age + " years old, " +
```
HellHunterMax: When making a string a small tip would be to add $ i front of the string that way you dont have to + all the time or use numbers like you did later on.
so: Console.WriteLine($"{name} {surName} you are {age} years old, your weight is {weight} Kg and your height is {height} cm")

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,80 @@
...
+			 1.Print all the info based on the example message below:
+			   Tom Jefferson is 19 years old, 
+			   his weight is 50 kg and his height is 156.5 cm.
+			 */
+
+			double height = 156.5;
+			int weight = 50;
+			int age = 19;
+			string surName = "Jefferson";
+			string name = "Tom";
```
Almantask, line 24: Read input from console.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,80 @@
...
+
+			Console.Write("Enter your height in centimeters: ");
+			centimeters = double.Parse(Console.ReadLine());
+
+			Console.Write("Enter your weight in kilograms: ");
+			kilograms = double.Parse(Console.ReadLine());
+
+			BMI = kilograms / (centimeters * centimeters) * 10000;
+			Console.WriteLine("Your Body Mass index 'BMI' is: {0}", Math.Round(bmi, 2));
+			Console.Write("\n");
```
Almantask: Using writeline instead is more simple.

Almantask (changes requested):

Src/BootCamp.Chapter/Program.cs
Joshhortt: Thanks Max for the great Tip. Just changed those Lines. 
Mission Done!

PR #627, opened 2020-12-22
Chapter3/homework/3

littleNoodle:


Src/BootCamp.Chapter/Item.cs
```diff
@@ -1,26 +1,25 @@
...
-            _name = name;
-            _price = price;
-            _weight = weight;
+            Name = name ?? throw new ArgumentNullException("name");
+            Price = price;
+            Weight = weight;
+        }
+        public bool Equals(Item b)
+        {
+            if (this.Name == (b?.Name??throw new ArgumentNullException()) && this.Price ==( b?.Price??0) && this.Weight ==(b?.Weight??0))
```
Almantask: Very hard to read line. The argument null exception is better thrown in item constructor, to not let a null item in the first place. I would just compare name and price and weight and be done with it.

Almantask: You don't have to recheck for item being null. Just do it once. Also one null item can be equal to another null item- that's fine.

Src/BootCamp.Chapter/Items/Headpiece.cs
```diff
@@ -12,5 +12,27 @@ public Headpiece(string name, decimal price, float weight)
...
+            get => _weight;
+            set => _weight = value;
+        }
+
+        private float _defense;
+        public float Defense
+        {
+            get => _defense;
+            set => _defense = value;
+        }
```
Almantask: No need all the backing fields- just use `{get;set;}`

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,9 +1,47 @@
...
+            //WriteLine(itemBuy);
+
+
+            //var items = shop.Items;
+            //WriteLine(items.Count);
+
+            //foreach (Item i in items)
+            //{
+            //    WriteLine(i.Name);
+            //}
```
Almantask: Delete dead code.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -40,6 +42,18 @@ public void Add(Item item)
         /// <param name="name"></param>
         public void Remove(string name)
         {
+            if(name is string)
```
Almantask: This is confusing to read. What you are saying is- `if it's not null, do this`. Would be more readable if you were more explicit about it. Also, do validation first and the rest of the logic after.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -40,6 +42,18 @@ public void Add(Item item)
...
+            {
+                foreach (Item item in _inventory.Items)
+                {
+                    if (item.Name == name)
+                        _inventory.RemoveItem(item);
+                    if (_inventory.Items.Count == 0)
```
Almantask: validation first.

Src/BootCamp.Chapter/Shop.cs
```diff
@@ -64,7 +87,22 @@ public decimal Buy(Item item)
...
+                        _money += item1.Price;
+                        return item1;
+                    }
+                }
+                return null;
+            }
+            else
+                throw new ArgumentException();
+            
         }
```
Almantask: Awfully famiilair piece of code. Can't you reuse remove item function?

Almantask (changes requested):

Almantask (approved):

PR #628, opened 2020-12-22
Chapter1/homework/3 - Homework Assignment 

Joshhortt:
The Console Application is a refactored version of homework 1AND2. It uses **funcions and less code..**
[https://github.com/csinn/CSharp-From-Zero-To-Hero/tree/Chapter1/Homework/1And2](url)
All the functions should be placed in **Checks.cs**.
All Logic should be placed into **Lesson3.cs**. 
Call Logic from main function, with only one line of code:    **Lesson3.Demo();**
It should pass **all Run tests**, in Visual Studio.

### What it does:
1. Prompts for user input and converts it to string (print message for request, read console input and return input).

2. Prompts for user input and converts it to int (print message for request, read console input and return converted input to int), 
    using the method **Convert.ToInt32**.

3.  Prompt for input and converting it to float (print message for request, read console input and return converted input to 
    float); - To do this, I managed to pass the **InvariantCulture** object to a method that has a CultureInfo parameter.

4. Calculates the BMI of a user (weight in kg, height in meters), using the following formula: 
    **weight / (height * height)**.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,14 +1,13 @@
-﻿using System;
+﻿
+using System;
 using System.Collections.Generic;
 using System.Text;
 
 namespace BootCamp.Chapter
 {
-	class Lesson3
-	{
-		public static void Demo()
-		{
-			//Body
-		}
-	}
+    class Lesson3
+    {
+
+        // Body
+    }
 }
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,56 @@
...
+            string name = PromptString("Name: ");
+            string surName = PromptString("Surname: ");
+            int age = PromptInt("Age: ");
+            float weight = PromptFloat("Weight(Kg): ");
+            float height = PromptFloat("Height(m): ");
+
+            float heightM = ConvertToMeters(height);
+            float yourBMI = CalculateBMI(heightM, weight);
+
+            Console.Write("\n");
```
Almantask, line 21: Use WriteLine instead.

Almantask (approved):
Second person missing as per requirements. Otherwise looks good

PR #629, opened 2020-12-24
Chapter3/homework/4

littleNoodle:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,11 +1,26 @@
...
 {
     class Program
     {
         static void Main(string[] args)
         {
+            Teacher<Math> like = new Teacher<Math>();
+            like.ProduceMaterial(); //welcome to Math lesson
+
+            var yanan = new Student(123);
+            yanan.LearnFrom<Teacher<Math>,Math>(like); //student 123 learn from Math teacher
```
Almantask, line 15: Seems to be a bit redundant. I'd say it would be enough to just pass a teacher- after all the teacher knows only one subject- so it can provide it to the student itself. This would work though if teacher were teaching multiple subjects- in this case they are not.

Src/BootCamp.Chapter/Subject.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Subject
+    {
+        List<string> subjectArray = new List<string> { "Maths", "Art","PE","Music","English","Programming" };
```
Almantask, line 9: I find the need of this questionable

Src/BootCamp.Chapter/Subject.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Subject
```
Almantask, line 7: add explicit access modifier eerywhere. I think this class should be public.

Almantask, line 7: Instead, since it doesn't do anything- maybe it could be just an interface? These kind of interfaces have a special name- Maker interface.

Almantask (approved):
Overall looks good

PR #630, opened 2020-12-28
Chapter1/homework/4

Joshhortt:
The Console Application is a refactored version of homework 3. It uses funcions & IF ELSE statements.
[https://github.com/csinn/CSharp-From-Zero-To-Hero/tree/Chapter1/Homework/4](url)
All the functions should be placed in Checks.cs.
All Logic should be placed into Lesson4.cs.
Call Logic from main function,in program.cs
It should pass Run all tests successfully.

**What it does:**
1.Prompts for user input and converts into a  string (print message for request, read console input and return input).
      _It Validates name input if correct:_
        -If input is empty: print error message "Name cannot be empty." and return "-". Created a function to be returned.
        -If input is not empty, return input.

2.Prompts for user input and converts it to int (print message for request, read console input and return converted input to int),
      _Validates number input if correct:_
       - If input is not a number: print "{input} is not a valid number." and return -1. Created a function to be returned.
       - If input is valid: return parsed input.

3.Prompt for input and converting it to float (print message for request, read console input and return converted input to
   float); - To do this, Parsing Culture-Invariant Floating-Point Numbers was used.
     _Validates number input if correct:_
       -If input is not a number: print "{input} is not a valid number." and return -1. Created a function to be returned.
       - If input is valid: return parsed input.

4.Calculates the BMI of a user (weight in kg, height in meters), using the following formula:
  weight / height / height. Here I changed the formula compared with the previous formula used in Homework 3
    _Validates BMI input if correct:_
      -If height and/or weight being less or equal 0 print "Failed calculating BMI. Reason:" + all error messages and return -1.
      -Error messages: Used also Environment.NewLine to concanate this string inside WriteLine'.
      -Height cannot be equal or less than zero, but was {input}."
      -Weight cannot be equal or less than zero, but was {input}."

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -16,7 +16,6 @@ namespace BootCamp.Chapter
     /// </summary>
     public static class Checks
     {
-
         public const int IntegerError = -1;
         public const string StringError = "-";
         public const float FloatError = -1.0f;
@@ -30,7 +29,7 @@ namespace BootCamp.Chapter
             
             if (!num)
             {
-                Console.WriteLine($"{input} is not a valid number.");
+                Console.WriteLine($"\"{input}\" is not a valid number.");
                 return IntegerError;
             }
             return variable;
@@ -56,7 +55,7 @@ namespace BootCamp.Chapter
             bool num = float.TryParse(input, NumberStyles.Float,CultureInfo.InvariantCulture, out variable);
             if (!num)
             {
-                Console.WriteLine($"{input} is not a valid number.");
+                Console.WriteLine($"\"{input}\" is not a valid number.");
                 return FloatError;
             }
             return variable;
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -14,28 +16,64 @@ namespace BootCamp.Chapter
...
+        public const float FloatError = -1.0f;
+
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            string input = Console.ReadLine();
+            int variable;
+            bool num = int.TryParse(input, out variable);
```
Almantask, line 28: You can inline the variable declaration. Also please use more concrete names, variables can be literally said about anything.

```cs
bool isNumber = int.TryParse(input, out int number);
```

Almantask (approved):
Looks very good 😄

PR #632, opened 2020-12-30
Chapter3/homework/5

littleNoodle:


Src/BootCamp.Chapter/ToggleableGridJagged.cs
```diff
@@ -1,18 +1,52 @@
...
+                        sb.Append("■");
+                    }
+                    else
+                    {
+                        sb.Append(" ");
+                    }
+                }
+                sb.Append(Environment.NewLine);
+            }           
+            var str = sb.ToString().TrimEnd("\r\n".ToCharArray());
```
Almantask: Split to a function of its own

Src/BootCamp.Chapter/ToggleableGridJagged.cs
```diff
@@ -1,18 +1,52 @@
...
         public ToggleableGridJagged(bool[][] toggles, IGridClearer gridClearer)
         {
+            _toggle = toggles;
             _gridClearer = gridClearer;
         }
 
         public void Toggle(int x, int y)
         {
+            _toggle[x][y] = !_toggle[x][y];
+            StringBuilder sb = new StringBuilder();
```
Almantask: Move this close to where it's used- next to loop

Src/BootCamp.Chapter/ToggleableGridJagged.cs
Almantask: So the function will encapsulate sb- it's a low level detail

Almantask: Avoid mixing high level functions with implementation of other functions.

Almantask (changes requested):
Try splitting the rendering of a grid from a logical representation and triggering toggles. When you couple both, you have quite a lot of duplication

Almantask:
Please reply to remarks by telling how you resolved them.

littleNoodle:
> Please reply to remarks by telling how you resolved them.

Src/BootCamp.Chapter/ToggleableGridJagged.cs
littleNoodle: Sorry for late response.
For this issue, I regard 2D array as special jagged array and define DisplayGrid class to reuse.
Also spend some time to solve the null reference exception about how to make 2D array works.
Through this I could handle 2D array initialization

Src/BootCamp.Chapter/ToggleableGridJagged.cs
littleNoodle: I leave the line 19 at Toggle function and make Display function only judge element value( true/false) of array.
High/Low level function is a great strategy, I should focus on code reuse. 
Have a good day~

littleNoodle:
> Please reply to remarks by telling how you resolved them.
I add the reponse above.
Happy new year !

Almantask (approved):

PR #634, opened 2021-01-03
Chapter1/homework/7 - Assignment

Joshhortt:
**A console Aplication that prints to the console each operation below to TextTableBuilder**
**_[https://github.com/csinn/CSharp-From-Zero-To-Hero/compare/Chapter1/Homework/7...Joshhortt:Chapter1/Homework/7?expand=1](https://github.com/csinn/CSharp-From-Zero-To-Hero/compare/Chapter1/Homework/7...Joshhortt:Chapter1/Homework/7?expand=1l)_**

    - Each Table shows a line of a person's name with their balance history. 
    - The last balance in the array is their current amount of money they have.
    - Currency shown in the balances are unspecified.
    - A table for given message should have a padding = 3
    - The Table itself is made of symbols (+), (-), ( | ).
   
Given messages printed to the console should output the following: 
1. The name and balance of person who had the biggest historic balance. --> FindHighestBalanceEver
2. The name and loss of a person with a biggest loss (balance change negative). --> FindPersonWithBiggestLoss
3. The name and current money of the richest person. --> FindRichestPersson
4. The name and current money of the most poor person. -- > FindMostPoorPerson

There were some important private Functions that needed to be added to the program to work.
In C#, the string type is immutable. It means a string cannot be changed once created. Instead of creating a new string object on the memory, modifying an original string at the same memory address. This behavior would hinder the performance if the original string changed multiple times by replacing, appending, removing, or inserting new strings in the original string.

To solve this problem, C# introduced the StringBuilder. It doesn't create a new object in the memory but dynamically expands memory to accommodate the modified string. 

StringBuilder is mutable.
StringBuilder performs faster than only string when appending multiple string values.

**For this specific Aplication:**
I used StringBuilder to append more strings.
I used the Append() method to add or append strings to the StringBuilder object.
I used the ToString() method to retrieve a string from the StringBuilder object.

- Stringbuilder to calculate People's Balances
- Stringbuilder to compare names
- Stringbuilder to output name using split method.
- Stringbuilder that looks for people's cash balance.

**Also used:**
- String Comparison.
- CultureInfo.InvariantCulture to convert string Array to float values.
- CultureInfo.InvariantCulture to format currency.

### DIFF
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -1,4 +1,5 @@
-﻿namespace BootCamp.Chapter
+﻿
+namespace BootCamp.Chapter
 {
     public static class BalanceStats
     {
--- a/Src/BootCamp.Chapter/BalanceStats.cs
+++ b/Src/BootCamp.Chapter/BalanceStats.cs
@@ -16,9 +16,9 @@ namespace BootCamp.Chapter
 
                 return "N/A.";
 
-            var (peopleNames, cash) = CalculatePeoplesBalances(peoplesBalances, "max");
+            var (peopleNames, cashIndex) = CalculatePeoplesBalances(peoplesBalances, "max");
 
-            return $"{peopleNames} had the most money ever. {FormatCash(cash)}.";
+            return $"{peopleNames} had the most money ever. {FormatCash(cashIndex)}.";
         }
 
         /// <summary>
@@ -30,13 +30,13 @@ namespace BootCamp.Chapter
 
                 return "N/A.";
 
-            var (peopleNames, cash) = CalculatePeoplesBalances(peoplesBalances, "min");
+            var (peopleNames, cashIndex) = CalculatePeoplesBalances(peoplesBalances, "min");
 
-            if (cash >= 0)
+            if (cashIndex >= 0)
 
                 return "N/A.";
 
-            return $"{peopleNames} lost the most money. {FormatCash(cash)}.";
+            return $"{peopleNames} lost the most money. {FormatCash(cashIndex)}.";
         }
 
         /// <summary>
@@ -49,14 +49,14 @@ namespace BootCamp.Chapter
 
                 return "N/A.";
 
-            var (peopleNames, cash) = CalculatePeoplesBalances(peoplesBalances, "rich");
+            var (peopleNames, cashIndex) = CalculatePeoplesBalances(peoplesBalances, "rich");
             var subjects = PeoplePlus(peopleNames);
             var word = new string[]
             {
                 (subjects) ? "are" : "is",
                 (subjects) ? "people" : "person"
             };
-            return $"{peopleNames} {word[0]} the richest {word[1]}. {FormatCash(cash)}.";
+            return $"{peopleNames} {word[0]} the richest {word[1]}. {FormatCash(cashIndex)}.";
         }
 
         /// <summary>
@@ -68,14 +68,14 @@ namespace BootCamp.Chapter
 
                 return "N/A.";
 
-            var (peopleNames, cash) = CalculatePeoplesBalances(peoplesBalances, "poor");
+            var (peopleNames, cashIndex) = CalculatePeoplesBalances(peoplesBalances, "poor");
             var subjects = PeoplePlus(peopleNames);
             var word = new string[]
             {
                 (subjects) ? "have" : "has",
             };
 
-            return $"{peopleNames} {word[0]} the least money. {FormatCash(cash)}.";
+            return $"{peopleNames} {word[0]} the least money. {FormatCash(cashIndex)}.";
         }
 
         private static (StringBuilder peopleNames, float cash) CalculatePeoplesBalances(string[] peoplesBalances, string attribution)
@@ -87,7 +87,7 @@ namespace BootCamp.Chapter
 
             CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
             var peopleNames = new string[peoplesBalances.Length];
-            var cash = new float[peoplesBalances.Length];
+            var cashIndex = new float[peoplesBalances.Length];
             int i, j;
 
             for (i = 0; i < peoplesBalances.Length; i++)
@@ -99,7 +99,7 @@ namespace BootCamp.Chapter
 
                 if (attribution == "max")
                 {
-                    cash[i] = values.Max();
+                    cashIndex[i] = values.Max();
                 }
                 else if (attribution == "min")
                 {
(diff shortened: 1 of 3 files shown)


Almantask:
Just wanted to repeat- I love your PR descriptions. Very informative and on-point.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
         /// Return name and balance(current) of person who had the biggest historic balance.
-        /// </summary>
-        public static string FindHighestBalanceEver(string[] peopleAndBalances)
+        /// </summary>       
+        public static string FindHighestBalanceEver(string[] peoplesBalances)
         {
-            return "";
+            if (IsNullOrEmpty(peoplesBalances))
+
+                return "N/A.";
```
Almantask: N/A might change, but the concept of some invalid value being returned will not. In order to avoid changing the exact value in multiple places- please use a constant.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+
+            var (peopleNames, cash) = CalculatePB(peoplesBalances, "max");
```
Almantask: I don't know what PB is.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+            {
+                (subjects) ? "have" : "has",
+            };
+
+            return $"{peopleNames} {word[0]} the least money. {FormatCash(cash)}.";
+        }
+
+        private static (StringBuilder peopleNames, float cash) CalculatePB(string[] peoplesBalances, string attribution)
+        {
+            var validation = new string[]
```
Almantask: validation is very vague, please rename better.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+            { "max", "min", "rich", "poor" };
+
+            if (!validation.Contains(attribution)) return (new StringBuilder(), 0);
+
+            CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
+            var peopleNames = new string[peoplesBalances.Length];
+            var cash = new float[peoplesBalances.Length];
+            int iOne, iTwo;
```
Almantask: unclear names. Avoid suffix for type.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+                    {
+                        if (iTwo == values.Length - 1) continue;
+                        balance[iTwo] = values[iTwo + 1] - values[iTwo];
+                    }
+                    cash[iOne] = balance.Min();
+                }
+                else if (attribution == "rich" || attribution == "poor")
+                {
+                    cash[iOne] = values[^1];
+                }
```
Almantask: Try changing a design to split function into getting just max or min. Doing both at the same time is not convenient and is much harder to test and reason about. Also add a few comments to make it more clear what is going on.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+            }
+            var maximumCash = (attribution == "max" || attribution == "rich");
+            float cashCheck = (maximumCash) ? cash.Max() : cash.Min();
```
Almantask: not clear naming.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+        private static bool IsNullOrEmpty(string[] peoplesBalances)
+        {
+            return (peoplesBalances == null || peoplesBalances.Length == 0);
+        }
+
+        private static float[] NumPyArrayConversion(string[] numbers)
+        {
+            CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
+            var nullString = 0;
+            int iThree, iFour;
```
Almantask: ?

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+                else
+                {
+                    convertList[iFour] = numeric;
+                    iFour++;
+                }
+            }
+            return convertList;
+        }
+
+        private static StringBuilder ListNames(string nameLine)
```
Almantask, line 175: You're not listing names, you are formatting or something like that.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+                return output;
+
+            var convert = nameLine.Split(", ");
+            int iFive;
+
+            output.Append(convert[0]);
+
+            if (convert.Length > 2)
+            {
+                for (iFive = 1; iFive < convert.Length; iFive++)
```
Almantask: ?

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -1,37 +1,224 @@
...
+                {
+                    var andCom = (iFive == convert.Length - 1) ? " and" : ",";
```
Almantask: ?

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -24,17 +26,62 @@ public static class TextTable
...
+                textTable.Append(' ', textSplit[maxIndex].Length - textSplit[iOne].Length);
+                textTable.Append(' ', padding).AppendLine("|");
+            }
+            for (iOne = 0; iOne < padding; iOne++)
+            {
+                textTable.Append("|").Append(' ', textSplit[maxIndex].Length + padding * 2).AppendLine("|");
+            }
+            textTable.Append("+").Append('-', textSplit[maxIndex].Length + padding * 2).AppendLine("+");
+
+                return textTable.ToString();
```
Almantask, line 66: Split more into private functions so that individual steps are more highlighted. Right now I am burdened to read all until I can understand. Also, quite a lot of duplicate code, reuse it passing argument as a difference.

Almantask (changes requested):

Src/BootCamp.Chapter/TextTable.cs
Joshhortt: Oh Gosh! I think I am going to refactor step by step,  everything! Thanks to you I am learning a lot! Thanks Kai..

Src/BootCamp.Chapter/BalanceStats.cs
Joshhortt: Naming is hard!

Src/BootCamp.Chapter/BalanceStats.cs
Joshhortt: CalculatePeoplesBalances takes a bit more memory but I think it's required. You're totally right, who will understand my code.

Src/BootCamp.Chapter/BalanceStats.cs
Joshhortt: You mean instead of 'iOne' and 'iTwo' --- > I should be using something like 'i' and 'i2' or 'i 'and  'j' right?

Src/BootCamp.Chapter/BalanceStats.cs
Joshhortt: Naming is hard. Going to be more clear about it  for now on.

Src/BootCamp.Chapter/BalanceStats.cs
Joshhortt: Yes formatting --> formattingPeoplesNames would be better..I think!

Src/BootCamp.Chapter/BalanceStats.cs
Almantask: I mean personIndex, cashIndex or just i and j. Those are easier to understand. one and two makes no sense.

Joshhortt:
Hi Kai.  After requested changes, I would kindly ask you if you can review my Homework 7 again. Tried lot's of different approaches messing it up quite a bit. This is the best I can do righ now!

Almantask (approved):
Better 😄

PR #637, opened 2021-01-11
Chapter 1. Homework 1 and 2

LeonT27:
### Tasks completed

- [x] Read name, surname, age, weight (in kg) and height (in cm) from console.

- [x] Print all the info based on the example message below:

    `Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm. `

- [x] Calculate and print body-mass index (BMI)

### Approach

Create the function GetInput(string) to used it multiple times and save some lines of code (I think this DRY), same with the CalculateBodyMaxIndex in case this need to be reused.

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -2,11 +2,34 @@
 
 namespace BootCamp.Chapter
 {
-    class Program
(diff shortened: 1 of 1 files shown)


Almantask (approved):
Looks good

PR #638, opened 2021-01-12
Chapter 1 Homework 3

LeonT27:
### Task completed

- [x] Create Lesson3.cs and refactor code from homework 1/2 to implement DRY principle
- [x] Making sure all the tests pass

### Approach
Create multiple functions to group same functionality for reuse purpose.
Run the test to see what they were testing and the expected value, then change the functions that required for the tests to pass


Almantask (approved):
Looks good

PR #639, opened 2021-01-13
Chapter 1 Homework 4

LeonT27:
### Task completed

- [x] Validate numeric inputs
- [x] Validate string inputs
- [x] Validate BMI

### Approach

Using if statement to compare the input value and assure that it was a valid value given the criteria, if the value was not valid then a message was display and a default value was returned.

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace BootCamp.Chapter
 {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public static float GetInputFloat(string question)
+        {
+            Console.Write(question + " ");
+            var input = Console.ReadLine();
+
+            if (float.TryParse(input, out var number)) return number;
+
+            Console.WriteLine($"\"{input}\" is not a valid number.");
+            return -1;
```
Almantask, line 53: Can use a constant instead of repeating -1 in all places.

Almantask (approved):
Yup, this looks great 😄  Keep it up!

PR #641, opened 2021-01-21
Homework1

vmdelfin12:
Practice use of Console.ReadLine() and calculated of BMI


Almantask (approved):
Second person info missing. But it's good enough for this homework.

PR #642, opened 2021-01-22
Chapter 1 Homework 5

LeonT27:
### Task completed

- [x] Sort array in ascending order
- [x] Inverse array
- [x] Remove first element from array
- [x] Remove last element from array
- [x] Remove element at a given index from array
- [x] Insert element at the start of array
- [x] Insert element at the end of array
- [x] Insert element at specified index of array

### Approach

Using for and while statements to iterate through the array.

- Sort
Get the smaller number and it index to switch it for the current value and position in the array.
- Inverse
Using for statement to go descending and using another array for assigning  the value.
- Remove & Insert
Using while and continue when the index was identified and remove or insert the value.


Almantask (approved):
great :)

PR #643, opened 2021-01-26
Chapter1 Homework 3

vmdelfin12:
- Improved code by using functions.
- Used functions to divide tasks


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+ 
+        public static void Demo()
+        {
+            string name = "";
```
Almantask, line 12: declare where this is used- not all at the top.

Almantask (approved):

PR #644, opened 2021-01-27
Chapter 1 Homework 4

vmdelfin12:
- Renamed Lesson3 to Lesson4
- Changed class to static class
- Added string, integer, and float input validation
- Converted height cm to m

### DIFF
--- a/Src/BootCamp.Chapter/Lesson4.cs
+++ b/Src/BootCamp.Chapter/Lesson4.cs
@@ -1,0 +1,99 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    public static class Lesson4
+    {
+        const int integerError = -1;
+        const float floatError = -1.0f;
+        const string stringError = "-";
+
+        public static void Demo()
+        {
+            string name = PromptString("Name: ");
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,99 @@
...
+                return num;
+            }
+
+            Console.WriteLine($"\"{variable}\" is not a valid number.");
+            return floatError;
+        }
+
+        public static float CalculateBMI(float weight, float height)
+        {
+            float bmi;
```
Almantask, line 76: No need to delcare at the top, just return it right away at line 82.

Src/BootCamp.Chapter/Lesson4.cs
```diff
@@ -0,0 +1,99 @@
...
+                return num;
+            }
+           
+            Console.WriteLine($"\"{variable}\" is not a valid number.");
+            return integerError;
+        }
+
+        public static float PromptFloat(string message)
+        {
+            string variable;
```
Almantask, line 60: combine delcaration and assigning of value.

Almantask (approved):
The rest looks great

PR #645, opened 2021-01-31
Chapter1/homework/9  - Assigment

Joshhortt:
**A Console Aplication that prints to the console character keys (both capital and non capital) to print arrows**

Encoding is the process of transforming a set of Unicode characters into a sequence of bytes. Decoding is the process of transforming a sequence of encoded bytes into a set of Unicode characters.

The Unicode Standard assigns a code point (a number) to each character in every supported script. A Unicode Transformation Format (UTF) is a way to encode that code point. The Unicode Standard uses the following UTFs:

UTF-8, which represents each code point as a sequence of one to four bytes.

**This simple Console  Aplication does:**
-  2 conversion type modes: From decimal to binary From binary to decimal
-  By using the following indicator characters: W A S D (both capital and non capital) keys to print arrows to Console.

**Note that:** The user can change the character to print, by just changing the string value.
The Default value will print out the same  result than any character W.
W- ↥
A- ↤
S- ↧
D- ↦

Default - ↥

### DIFF
--- a/Src/BootCamp.Chapter/BinaryConverter.cs
+++ b/Src/BootCamp.Chapter/BinaryConverter.cs
@@ -8,12 +8,36 @@ namespace BootCamp.Chapter
     {
         public static long ToInteger(string binary)
         {
-            return 0;
+            if (binary == null || binary == string.Empty)
+                return 0;
+
+            else if (IsBinary(binary))
+            {
+                long num = Convert.ToInt32(binary, 2);
+                return num;
+            }
+            else
+            {
+                throw new InvalidBinaryNumberException(binary);
+            }
         }
 
         public static string ToBinary(long number)
         {
-            return "";
+            string str = Convert.ToString(number, 2);
+            return str;
+        }
+
+        private static bool IsBinary(string str)
+        {
+            bool result = false;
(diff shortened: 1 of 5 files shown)


Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -8,12 +8,36 @@ public static class BinaryConverter
     {
         public static long ToInteger(string binary)
         {
-            return 0;
+            if (binary == null || binary == string.Empty)
```
Almantask, line 11: Prefer to use `String.IsNullOrEmpty`

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -8,12 +8,36 @@ public static class BinaryConverter
...
+                return 0;
+
+            else if (IsBinary(binary))
+            {
+                long num = Convert.ToInt32(binary, 2);
```
Almantask, line 16: Nice, actually didn't know there was a premade function for that :)

Almantask (approved):
Looks good 🥇

PR #646, opened 2021-02-07
Chapter1/homework/10 - Assignment

Joshhortt:
**Ceasar Cipher console aplication that prints to console an encrypted message right-shifted by three characters and decrypts the encrypted message back left-shifted  by three.**

To understand all about Information Security is beyond the scope of tis assignment, but understanding information security, cryptography is crucial today. We are dealing with security. So, it is necessary to understand the basic terminology that we will be using in this challenging Assignment.

Plaintext - original message
Ciphertext - coded message
cipher - algorithm for transforming plaintext to ciphertext
key - info used in cipher known only to sender/receiver
encipher - (encrypt) converting plaintext to ciphertext
decipher - (decrypt) recovering ciphertext from plaintext
cryptography - study of encryption principles/methods

Fo this Assignment I used the Symmetric Cipher Model which really is very easy to understand.
It has a plaintext that is to be encrypted in Ciphertext via some encryption algorithm, and sent via a secure channel to the receiver. At the receiver part, this Ciphertext gets decrypted by some decryption algorithm, so that the receiver can read the message.

The Caesar Cipher involves replacing each letter of the alphabet with the letter – standing places down or up according to the key given.

Suppose “M” is the plaintext and the key is given as 3, then you get the Ciphertext as the letter “P”. In a similar way “A” is the plaintext then Ciphertext will be – “D”.

**Algorithm For :** 

P | Plaintext
C | Ciphertext
E | Encryption Algorithm
D | Decryption Algorithm
K | Key

**Encryption**

For each plaintext (P), Ciphertext (C) is.

C = E (3, P) = (P+3) mod 26

**For Decryption**
P = D (3, C) = (C-3) mod 26

In general  [C= E (K, P) = (P+K) mod 26]
[P = D (K, C) = (C-K) mod 26], where

balance change negative). --> FindPersonWithBiggestLoss
3. The name and current money of the richest person. --> FindRichestPersson
4. The name and current money of the most poor person. -- > FindMostPoorPerson

There were some important private Functions that needed to be added to the program to work.

### DIFF
--- a/Src/BootCamp.Chapter/CaesarCipher.cs
+++ b/Src/BootCamp.Chapter/CaesarCipher.cs
@@ -4,19 +4,54 @@ using System.Text;
 
 namespace BootCamp.Chapter
 {
+    // ToDo: Change code so it only allows letters from an alphabet.
+
     /// <summary>
     /// Caesar cipher which supports all ASCII character (256 in total)
     /// </summary>
     public static class CaesarCipher
     {
+        // Constant for the 26 ASCII character keys
+        private const int MOD26 = 26;
+
         public static string Encrypt(string message, byte shift)
         {
-            return message;
+            return CipherText(message, shift);
         }
 
         public static string Decrypt(string message, byte shift)
         {
-            return message;
+            return CipherText(message, MOD26 - shift);
+        }
+
+        private static string CipherText(string inputText, int shift)
+        {
+            if (string.IsNullOrEmpty(inputText)) return inputText;
+
+            var text = new StringBuilder();
+
+            foreach (var characterNumber in inputText)
+            {
+                text.Append(PlainChar(characterNumber, shift));
+            }
+            return text.ToString();
+        }
+
+        // Constant for Upper Alphabet keys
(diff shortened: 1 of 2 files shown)


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -4,19 +4,53 @@
...
 {
+    // ToDo: Change code so it only allows letters from an alphabet.
+
     /// <summary>
     /// Caesar cipher which supports all ASCII character (256 in total)
     /// </summary>
     public static class CaesarCipher
     {
+       
+        private const int MOD26 = 26;
```
Almantask, line 15: Not a clear name.

Almantask (approved):
Perfect :)

Src/BootCamp.Chapter/CaesarCipher.cs
Joshhortt: Already changed constant  name to 'NumOfKeys'.. Thanks

PR #647, opened 2021-02-08
Chapter1/homework/5

vmdelfin12:
- Sorted array
- Reverse array
- Insert element to beginning, middle, and end of array
- Remove element in beginning, middle, and end of array


Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -0,0 +1,197 @@
...
+    public static class ArrayOperations
+    {
+        /// <summary>
+        /// Checks if array is null or empty
+        /// </summary>
+        /// <param name="array"></param>
+        /// <returns>true or false</returns>
+        public static bool IsNullOrEmpty(int[] array)
+        {
+            if(array == null || array.Length == 0)
```
Almantask, line 12: You can directly `return  array == null || array.Length == 0` because it translate to true or false. Now it reads:
```
if(true) return true;
else return false;
```
which is redundant;

Almantask (approved):
image
PR is made to the wrong branch

PR #648, opened 2021-02-14
Chapter3, Homework 6

Tailor-AI:
Delegate/Events homework.

Task1.
Parsing with ReadAllLines and function StringToPerson to populate the personlist _people.

Task 2.
In the Filter function I return a new list which takes the elements from the _people list which fulfills the Predicate parameter.

Task 3.
Using ReadKey to get the pressed key and firing event if valid (wanted) keys are pressed. Using switch/case to apply choosen filter.
I was not sure how the third task should be implemented with user input so it may be a bit cumbersome.

### DIFF
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -12,24 +12,28 @@ namespace BootCamp.Chapter
 
         public ContactsCenter(string peopleFile)
         {
-            string[] people;
-
             try
             {
-                people = File.ReadAllLines(peopleFile);
+                _people = File.ReadAllLines(peopleFile).Select(s => StringToPerson(s)).ToList();
             }
-            catch
+            catch (FileNotFoundException ex)
             {
-                throw new FileNotFoundException();
+                throw new FileNotFoundException($"File MOCK_DATA.csv was not found. Details: {ex}");
             }
-
-            foreach (var line in people)
+            catch (DirectoryNotFoundException ex)
             {
-                _people.Add(StringToPerson(line));
+                throw new FileNotFoundException($"Directory with the file MOCK_DATA.csv was not found. Details: {ex}");
             }
 
             // Remove header text.
-            _people.RemoveAt(0);
+            if (_people.Count != 0)
+            {
+                _people.RemoveAt(0);
+            }
+            else
+            {
+                throw new Exception("File was empty.");
+            }
         }
 
         /// <summary>
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -18,7 +18,7 @@ namespace BootCamp.Chapter
             {
                 people = File.ReadAllLines(peopleFile);
             }
-            catch (Exception)
+            catch
             {
                 throw new FileNotFoundException();
             }
@@ -48,18 +48,12 @@ namespace BootCamp.Chapter
             Person person = new Person();
 
             person.Name = personData[0];
-
             person.SureName = personData[1];
-
             DateTime.TryParse(personData[2], out DateTime birthday);
             person.Birthday = birthday;
-
             person.Gender = personData[3] == "Male" ? Enums.Gender.Male : Enums.Gender.Female;
-
             person.Country = personData[4];
-
             person.Email = personData[5];
-
             person.StreetAdress = personData[6];
 
             return person;
--- a/Src/BootCamp.Chapter/ContactsCenter.cs
+++ b/Src/BootCamp.Chapter/ContactsCenter.cs
@@ -1,16 +1,35 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace BootCamp.Chapter
 {
     public class ContactsCenter
     {
-        private readonly List<Person> _people;
+        private readonly List<Person> _people = new List<Person>();
 
         public ContactsCenter(string peopleFile)
         {
-            // load people
+            string[] people;
+
+            try
+            {
+                people = File.ReadAllLines(peopleFile);
+            }
+            catch (Exception)
+            {
+                throw new FileNotFoundException();
+            }
+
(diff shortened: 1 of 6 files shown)


Src/BootCamp.Chapter/Demo.cs
```diff
@@ -0,0 +1,133 @@
...
+                case 'b':
+                    foreach (var person in contactsCenter.Filter(PeoplePredicates.IsB))
+                    {
+                        Console.WriteLine(person);
+                    }
+
+                    PrintMenuChoice();
+                    break;
+                case 'c':
+                    foreach (var person in contactsCenter.Filter(PeoplePredicates.IsC))
```
Almantask: Repetitive- you can extract a function out of this and just feed a filter inside that function. That way you won't repeat 10 lines of code.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -19,9 +38,31 @@ public ContactsCenter(string peopleFile)
...
         {
-            var people = new List<Person>();
-            // ToDo: implement applying filter.
-            return people;
+            return _people.Where(x => predicate(x)).ToList();
+        }
+
+        private static Person StringToPerson(string line)
+        {
+            var personData = line.Split(',');
```
Almantask, line 50: Weird formatting. Too many empty lines. Combine related blocks of code by not adding an empty line.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,35 @@
...
         public ContactsCenter(string peopleFile)
         {
-            // load people
+            string[] people;
+
+            try
+            {
+                people = File.ReadAllLines(peopleFile);
+            }
+            catch (Exception)
```
Almantask: No need to specify anything if you catch all.

Almantask (changes requested):

Src/BootCamp.Chapter/ContactsCenter.cs
Tailor-AI: Removed (Exception) after catch.

Src/BootCamp.Chapter/ContactsCenter.cs
Tailor-AI: Removed empty lines between personData assignment.

Src/BootCamp.Chapter/Demo.cs
Tailor-AI: Made a new function PrintPeople and renamed PrintPeopleFilter to ChoosePeopleFilter which use the PrintPeople function.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,35 @@
...
     {
-        private readonly List<Person> _people;
+        private readonly List<Person> _people = new List<Person>();
 
         public ContactsCenter(string peopleFile)
         {
-            // load people
+            string[] people;
+
+            try
```
Almantask, line 15: This will throw a similar exception anyways. What if the error is elsewhere? You just hide it that way. Also, when throwing an error, try to be as detailed as possible, because when the problem occurs, we want all possible hints to track it down.

Src/BootCamp.Chapter/ContactsCenter.cs
```diff
@@ -1,16 +1,35 @@
...
+            {
+                people = File.ReadAllLines(peopleFile);
+            }
+            catch
+            {
+                throw new FileNotFoundException();
+            }
+
+            foreach (var line in people)
```
Almantask: Instead, you can: 
```cs
_people = File.ReadAllLines(peopleFile)
                       .Select(s => StringToPerson(s));
                       .ToList();
```

Src/BootCamp.Chapter/Demo.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public bool DemoRunning
+        {
+            get { return _demoRunning; }
+            set
+            {
+                if (_demoRunning != value)
+                {
+                    _demoRunning = value;
+                    Console.WriteLine(_demoRunning == true ? "Demo started." : "Demo Ended.");
```
Almantask: It might look conveninent, but it might be annoying. It's okay in this case, but please try to avoid such side-affect for properties. They should be simple.

Src/BootCamp.Chapter/Demo.cs
```diff
@@ -0,0 +1,128 @@
...
+                {
+                    DemoRunning = true;
+                    OnKeyPressed?.Invoke(this, null);
+                }
+            } while (DemoRunning);
+        }
+
+        private void RunDemoMenu(object sender, EventArgs eventArgs)
+        {
+            OnKeyPressed -= RunDemoMenu;
```
Almantask, line 46: Pretty cool to see you try and switch behavior like this dynamically.

Almantask (changes requested):

Src/BootCamp.Chapter/ContactsCenter.cs
Tailor-AI: Ok. If I understand this correctly I need to catch two exceptions (File and DirectoryNotFound) which both throws FileNotFound (this because DirectoryNotFound expects FileNotFound in the tests). 

I throw a general exception when file is empty (also expected in the tests).

Src/BootCamp.Chapter/ContactsCenter.cs
Tailor-AI: Nice! Even more cleaner.

Src/BootCamp.Chapter/Demo.cs
Tailor-AI: Agree. I removed the property and use only a private field to read the status of the demo. I could extract the functionality to a seperate function but it feels like overkill?

Src/BootCamp.Chapter/Demo.cs
Tailor-AI: Thanks!

PR #653, opened 2021-02-25
Solution for homework 1 and 2.

dt1307:
- created the user variables with their specific data type and collected the data from the user
- formatted the output to include all the user information
- added the BMI formula
- formatted the BMI output (rounded to 2 decimals) and displayed with the correct result for each user
- created a for loop to fulfil the last homework task ("Do it for 2 people (repeat the same thing twice)")


Almantask (approved):
Looks clean :) Good enough for the first homework.

PR #656, opened 2021-03-16
Homework 1 first solution

cyberiumcodes:
homework for c sharp bootcamp, chaper1, homework 1 and 2


Almantask (approved):
Please don't declare variables all in one place. Declare them next to where you use them. Anyways, this looks good enough for the first homework.

PR #659, opened 2021-03-19
Homework 3

illayarajah:
Completion of Homework 3

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+
+
+           
+            string firstName = returnString("Enter First Name");
+            string lastName = returnString("Enter Last Name");
+            int age = returnInt("Enter Your Age");
+            float weight = returnFloat("Enter Your Weight in KG");
+            float height = returnFloat("Enter Your Height in M");
+            float bmi = calculateBMI(weight, height);
+            Console.WriteLine("Your BMI is " + bmi);
+            
+            
+
+
+
+
+         
+        }
+
+
+        public static int returnInt(string message)
+        {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,57 @@
...
+            float height = returnFloat("Enter Your Height in M");
+            float bmi = calculateBMI(weight, height);
+            Console.WriteLine("Your BMI is " + bmi);
+            
+            
+
+
+
+
+
```
Almantask: What's up with so many blank lines?

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,57 @@
...
+            
+
+
+
+
+         
+        }
+
+
+        public static int returnInt(string message)
```
Almantask: PascalCase your method names

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Console.WriteLine(message);
+            return Convert.ToInt32(Console.ReadLine());
+        }
+
+        public static string returnString(string message)
```
Almantask: You don't just return a string. You `PromptString`. Please rename. Don't use a word `return` in a name ever, it does not add any value.

Almantask (changes requested):

Src/BootCamp.Chapter/Lesson3.cs
illayarajah: Removed

Src/BootCamp.Chapter/Lesson3.cs
illayarajah: Thank you, fixed as well.

Src/BootCamp.Chapter/Lesson3.cs
illayarajah: done, thanks for reviewing the work.

Almantask:
Looks good

PR #660, opened 2021-03-22
Chapter1/homework/4

illayarajah:
Homework 4

### DIFF
--- a/Src/BootCamp.Chapter/Lesson3.cs
+++ b/Src/BootCamp.Chapter/Lesson3.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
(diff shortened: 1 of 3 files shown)


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Globalization;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+                       
+            string firstName = returnString("Enter First Name");
```
Almantask: Second person missing

Almantask (changes requested):
Refer to other remarks from previous PR.

Src/BootCamp.Chapter/Lesson3.cs
illayarajah: I actually thought that we just do it twice to make sure the information is correct. 
i will add the 2nd person in.

Almantask:
Looks good

PR #663, opened 2021-03-30
Homework 7

illayarajah:
Review Homework 7

### DIFF
--- a/Src/BootCamp.Chapter/TextTable.cs
+++ b/Src/BootCamp.Chapter/TextTable.cs
@@ -1,4 +1,7 @@
-﻿namespace BootCamp.Chapter
+﻿using System;
+using System.Text;
+
+namespace BootCamp.Chapter
 {
     /// <summary>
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +37,81 @@ public static class TextTable
...
+
+                sb.Append(BuildPadding(paddingWidth, padding));
+                sb.Append(BuildFirstOrLast(paddingWidth));
+
+                //Console.WriteLine(sb.ToString());
+
+                return sb.ToString(); ;
+            }
+            else
+                return "";
```
Almantask, line 71: Could be split into smaller functions

Almantask, line 71: Either inline both or add brackets.
Also return String.Empty instead- more readable.

Almantask (approved):
Overall still looks okay

PR #666, opened 2021-04-06
Split code into functions

deempa:


Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,59 @@
...
+
+                float weight = PrintAndReadFloat("Weight: ");
+
+                float height = PrintAndReadFloat("Height: ");
+
+                var bmi = BMICalc(weight, height);
+
+                Console.WriteLine($"{name} {sureName} is {age} years old, his weight is {weight} kg and his height is {height} cm.");
+
+                Console.WriteLine($"Your BMI - {bmi:N2}");
```
Almantask, line 27: Instead of a loop, consider moving this to a function and calling it twice.

Almantask (changes requested):

PR #671, opened 2021-04-09
Encoding HW

illayarajah:


Src/BootCamp.Chapter/ArrowMovement.cs
```diff
@@ -18,7 +18,30 @@ public static class ArrowMovement
         /// <returns>One of the arrow characters. '↥' by default.</returns>
         public static char GetIndicator(char symbol)
         {
-            return '-';
+            char result = Convert.ToChar('\u21A5');
+            string acceptedChars = "wasdWASD";
```
Almantask, line 22: Could be const

Src/BootCamp.Chapter/ArrowMovement.cs
```diff
@@ -18,7 +18,30 @@ public static class ArrowMovement
         /// <returns>One of the arrow characters. '↥' by default.</returns>
         public static char GetIndicator(char symbol)
         {
-            return '-';
+            char result = Convert.ToChar('\u21A5');
```
Almantask, line 21: convert it to lowercase. In that case won't need half of the checks.

Src/BootCamp.Chapter/BinaryConverter.cs
```diff
@@ -2,18 +2,68 @@
...
 using System.Text;
 
+
 namespace BootCamp.Chapter
 {
     public static class BinaryConverter
     {
         public static long ToInteger(string binary)
         {
+            if (!(binary==null || binary == ""))
```
Almantask, line 12: can use `string.IsnullorEmpty()`

Almantask (changes requested):

PR #672, opened 2021-04-09
Homework 10

illayarajah:


Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -11,12 +11,48 @@ public static class CaesarCipher
...
+            try
+            {
+                byte[] byteMessage = Encoding.GetEncoding("iso-8859-1").GetBytes(message);
+                for (var i = 0; i < byteMessage.Length; i++)
+                {
+                    
+                    byteMessage[i] = (byte)(byteMessage[i] + shift);
+                    
+                }
+                message = System.Text.Encoding.GetEncoding("iso-8859-1").GetString(byteMessage);
```
Almantask, line 23: You don;'t need to get the encoding every time. Get it once and reuse it.

Src/BootCamp.Chapter/CaesarCipher.cs
```diff
@@ -11,12 +11,48 @@ public static class CaesarCipher
...
+                return null;
+            }
+            
+
+           
+            
             return message;
         }
 
         public static string Decrypt(string message, byte shift)
```
Almantask, line 36: Decrypt can reuse encrypt (not quite, but almost). They both work exactly the same- only the shift is different.

Almantask (changes requested):

PR #673, opened 2021-04-10
Chapter2/homework/3

deempa:


Almantask (approved):
I will approve this, because I think essentially the idea is understood.
However, what I expected was that Mac and Ms will use different hierarchies of parts (Mac and Ms). o instead of body, you would install MacBody or WinBody which both inherit body. Up to you to make amends or not.

deempa:
So I should create class for each Mac&Ms part of computer?

בתאריך יום א׳, 11 באפר׳ 2021, 12:37, מאת Kaisinel ***@***.***

PR #674, opened 2021-04-10
Chapter2/homework/4

deempa:


Almantask (approved):
Main could be split into a few function, as you don't want implementation in main, just calls to other higher level functions. Other than that, looks great.

PR #681, opened 2021-07-09
Homework1&2

halloffame0793:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+            get
+            {
+                return _height;
+            }
+            set
+            {
+                _height = value;
+            }
+
+        }
```
Almantask, line 60: This can be auto properties.
Instead of:
```cs
        public string Name {
            get
            {
                return _name;
            }
            set
            {
                _name = value;
            } 

        }
```
you can:
```cs
        public string Name{get;set;}
```

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+
+        public Person(string name, string age, string weight, string height)
+        {
+            _name = name;
+            _age = age;
+            _weight = weight;
+            _height = height;
+        }
+
+        public double BMIcalc()
```
Almantask, line 70: BMICalc is a noun. Methods normally should be verbs. So a better name would be CalculateBmi. Also avoid using abreviations like that, it is indeed less code, but it reads worse.

Also, in C# the convention is to name methods PascalCase, even if it's an acronym.
Actual name should be:
`CalculateBmi()`

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+        {
+            var inMeters = double.Parse(_height) * 100;
+            var metersSquared = Math.Pow(inMeters,2);
+            var BMI = int.Parse(_weight) / metersSquared;
```
Almantask, line 74: local variables should be non-capital, regardless of acronym.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+                return _name;
+            }
+            set
+            {
+                _name = value;
+            } 
+        
+        }
+
+        public string Age
```
Almantask, line 25: Age should be years- so int.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+                return _age;
+            }
+            set
+            {
+                _age = value;
+            }
+
+        }
+
+        public string Weight {
```
Almantask, line 38: weight should be a number with a floating point (double or float)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+            {
+                return _weight;
+            }
+            set
+            {
+                _weight = value;
+            }
+        }
+
+        public string Height
```
Almantask, line 49: Should be float or double

Src/BootCamp.Chapter/Program.cs
```diff
@@ -2,11 +2,99 @@
...
+        }
+
+    }
     class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            var Person1 = new Person("Tom Jefferson", "19", "50", "156.5");
+            var Person2 = new Person("Mike Miller", "28", "75", "160.2");
```
Almantask, line 85: Should be read from console.

Almantask (changes requested):

PR #682, opened 2021-07-10
Homework 3

halloffame0793:
Writing the values to the console probably isn't needed but it fails with or without that.

### DIFF
--- a/Src/BootCamp.Chapter/Checks.cs
+++ b/Src/BootCamp.Chapter/Checks.cs
@@ -14,29 +14,36 @@ namespace BootCamp.Chapter
     /// </summary>
     public static class Checks
     {
-        public static int PromptInt(string message)
+        public static int PromptInt()
         {
-            var Age = message;
-            var ageConverted = int.Parse(Age);
+            Console.WriteLine("Please enter your age:");
+            var age = Console.ReadLine();
+            var ageConverted = int.Parse(age);
 
-            Console.WriteLine(ageConverted);
 
             return ageConverted;
         }
 
-        public static string PromptString(string message)
+
+        public static string PromptString()
         {
-            var Name = message;
-  
-            return Name;
+            Console.WriteLine("Please enter your name:");
+            var name = Console.ReadLine();
+
+            return name;
         }
+
+
 
         public static float PromptFloat(string message)
         {
-            var convertedFloat = float.Parse(message);
-            
-            return convertedFloat;
+        var input = message;
+        var convertedFloat = float.Parse(input);
+
+        return convertedFloat;
         }
+
+
 
         public static float CalculateBmi(float weight, float height)
         {
@@ -44,9 +51,8 @@ namespace BootCamp.Chapter
             var metersSquared = Math.Pow(inMeters, 2);
             var BMI = weight / metersSquared;
 
-            Console.WriteLine($"Your BMI is {BMI}");
-
             return (float)BMI;
         }
+
     }
 }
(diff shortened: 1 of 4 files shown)


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,37 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            var Age = message;
```
Almantask: use consistent variable names. For local variables- camelCase (starts with non-capital)

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,37 @@ public static class Checks
...
+            var ageConverted = int.Parse(Age);
+
+            Console.WriteLine(ageConverted);
```
Almantask: If it doesn't ask to print the age, don't print it.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,37 @@ public static class Checks
...
 
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            var inMeters = height * 100;
+            var metersSquared = Math.Pow(inMeters, 2);
+            var BMI = weight / metersSquared;
+
+            Console.WriteLine($"Your BMI is {BMI}");
```
Almantask: Calculating is not the same as printing. It should go to another line.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -9,6 +9,25 @@ class Program
     {
         static void Main(string[] args)
         {
+            Console.WriteLine("Please enter your name:");
+            var Name = Console.ReadLine();
```
Almantask: Each of these 2 lines should go to the prompt methods.

Almantask: Because prompt means ask for. You ask for it in console, then read the user input form console.

halloffame0793: yup I refactored to include those lines in the methods. Doing it this way makes it so these prompt methods do not need to accept parameters, that may be a issue as the unit tests required these methods to accept params.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BootCamp.Chapter
+{
+    class Lesson3
+    {
+        public static void Demo()
+        {
+
```
Almantask, line 13: nitpick: no need an empty line right after a function's opening bracket.

Src/BootCamp.Chapter/Lesson3.cs
```diff
@@ -0,0 +1,29 @@
...
+            Console.WriteLine("Please enter your height(cm):");
+            var height = Console.ReadLine();
+            var cHeight = Checks.PromptFloat(height);
+
+            Console.WriteLine("Please enter your weight(km):");
+            var weight = Console.ReadLine();
+            var cWeight = Checks.PromptFloat(weight);
+
+            Checks.CalculateBmi(cWeight, cHeight);
+
```
Almantask, line 26: nitpick: same here.

Almantask (approved):
Looks good :)

Almantask:
Looks much better now!

PR #684, opened 2021-07-14
Chapter1/homework/4

halloffame0793:


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,114 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            Console.WriteLine(message);
+            var age = Console.ReadLine();
```
Almantask, line 21: There is a pattern for prompting something. You could make a method for those 2 lines:

```cs
public static string Prompt(string message)
{
                Console.WriteLine(message);
                return Console.ReadLine();
} 
```

Almantask: You can also include a null/empty string validation

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,114 @@ public static class Checks
...
+
         public static float CalculateBmi(float weight, float height)
         {
-            // To do: call your implementation. 
-            return 0;
+            if ((weight > 0) && (height > 0))
+            {
+                if (weight > 0)
+                {
+                    if (height > 0)
```
Almantask, line 99: Overlapping if

Almantask: So do this validation first, outside not combined with other validations.

Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +17,114 @@ public static class Checks
...
+            {
+                return -1;
+            }
+
+ 
         }
 
+
+
         public static float CalculateBmi(float weight, float height)
```
Almantask, line 93: Try to simplify, too many ifs (nested)

Almantask (changes requested):

PR #686, opened 2021-07-23
Chapter1/Homework/1And2

Anequit:


Src/BootCamp.Chapter/IPerson.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace BootCamp.Chapter
+{
+    public interface IPerson
```
Almantask: It's very rare, that data needs an interface. In this case, it certainly does not, because it's just a single implementation of a person with a getter and setter.

Src/BootCamp.Chapter/Person.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string FirstName { get => _firstName; set => _firstName = value; }
+        public string SurName { get => _surName; set => _surName = value; }
+        public int Age { get => _age; set => _age = value; }
+        public float Weight { get => _weight; set => _weight = value; }
+        public float Height { get => _height; set => _height = value; }
+        public float BMI { get => (float)Math.Round(CalculateBMI(_height, _weight), 2); }
+
+        public void Introduce()
+        {
+            Console.WriteLine($"{_firstName} {_surName} is {_age} years old, his weight is {_weight} kg and his height is {_height} cm.");
```
Almantask: You couple person with console. It's better to just return an introduction as a string.

Almantask (changes requested):

PR #688, opened 2021-08-09
Chapter1/homework/1

iamzumie:
homework 1 completed

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,55 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string fullName = askFullName();
+            int age = askAge();
+            int weight = askWeight();
+            double height = askHeight();
+
+            Console.WriteLine($"{fullName} is {age} years old, his weight is {weight} kg and his height is {height:F1} cm.");
+
+            CalculateBMI(weight, height);
+        }
+
+        static string askFullName()
+        {
+            Console.WriteLine("What is your first name?");
+            string firstname = Console.ReadLine();
+            Console.WriteLine("What is your family name?");
+            string surname = Console.ReadLine();
+
+            return firstname + " " + surname;
+        }
+
+        static int askAge()
+        {
+            Console.WriteLine("What is your age?");
+            int age = Convert.ToInt32(Console.ReadLine());
+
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,55 @@ class Program
...
+            int age = askAge();
+            int weight = askWeight();
+            double height = askHeight();
+
+            Console.WriteLine($"{fullName} is {age} years old, his weight is {weight} kg and his height is {height:F1} cm.");
+
+            CalculateBMI(weight, height);
+        }
+
+        static string askFullName()
```
Almantask: functions in C# start with capital letter (PascalCase)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,55 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string fullName = askFullName();
+            int age = askAge();
+            int weight = askWeight();
+            double height = askHeight();
+
+            Console.WriteLine($"{fullName} is {age} years old, his weight is {weight} kg and his height is {height:F1} cm.");
```
Almantask: doing the same for a second person is missing

Almantask (approved):

Src/BootCamp.Chapter/Program.cs
iamzumie: I wrote a loop around it so it will run twice.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,58 @@ class Program
...
+
+        static double AskHeight()
+        {
+            Console.WriteLine("How tall are you?");
+            double height = Convert.ToDouble(Console.ReadLine());
+
+            return height;
+        }
+
+        static void CalculateBmi(int weight, double height)
```
aduckinpants, line 56: This is less 'CalculateBmi' and more 'ConsoleWriteBmi'. It would make more sense to return the calculated value and have the caller handler the write so that clean separation of responsibilities is maintained.

aduckinpants (approved):

Src/BootCamp.Chapter/Program.cs
iamzumie: Got it, naming variables & methods is def something i need to work on.

PR #695, opened 2021-08-30
Chapter1/homework/7

iamzumie:


Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
         /// </summary>
         public static string FindHighestBalanceEver(string[] peopleAndBalances)
         {
-            return "";
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0) return "N/A.";
+
+            int highestBalance = 0;
```
aduckinpants: With the function dealing with currency amounts is int the most appropriate type to represent an amount of currency?

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+            List<String> listOfPeople = new List<String>();
```
aduckinpants: This is going to be a bit pedantic: `listOfPeople` is not a great variable name. We know it's a list, it's right there on the left and right of it. `people` would work just fine. There's no need to emphasize mechanism over intent.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+
+            foreach (var values in peopleAndBalances)
+            {
+                // Splitting the values
+                var array = values.Split(", ");
```
aduckinpants: This may work in the unit test data set but this is going to be extremely fragile in any other situation as you are encoding whitespace requirements into the input data that may not map to any such restriction in data generation. In other words it could just as easily be `10,11,12` as `10, 11, 12`.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
         /// </summary>
         public static string FindPersonWithBiggestLoss(string[] peopleAndBalances)
-        {
-            return "";
+        { 
+            if (peopleAndBalances == null || peopleAndBalances.Length == 0) return "N/A.";
+
+            CultureInfo.CurrentCulture = new CultureInfo("en-US"); // because numbers are with . instead of ,
+            List<String> listOfPeople = new List<String>();
+            double biggestLoser = 0;
```
aduckinpants: This is better then `int` but is double the best way to represent currency?

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+                var array = values.Split(", ");
+
+                // if the current person doesn't have at least 3 balances
+                if (array.Length <= 2) return "N/A.";
+
+                // Check what the biggest loss if for each person between current & next balance
+                for (int i = 1; i < array.Length - 1; i++)
+                {
+                    bool succes = double.TryParse(array[i], out double currentBalance);
+                    bool succes2 = double.TryParse(array[i+1], out double NextBalance);
```
aduckinpants: `succes` / `succes2` are not good examples of variable naming.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+                return $"{people[0]}";
+            }
+            else if (people.Count == 2)
+            {
+                return $"{people[0]}, {people[1]}";
+            }
+            else
+            {
+                string listpeople = "";
+                for (int i = 0; i < people.Count - 2; i++)
```
aduckinpants: It's possible to do this without the extra string allocs in a tight loop. Take a look at `string.Join`

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+
+            if (listOfPeople.Count == 1)
+            {
+                return $"{persons} has the least money. {amount}.";
+            }
+            return $"{persons} have the least money. {amount}.";
+        }
+
+        // Return the name of the people
+        public static string CountPeople(List<string> people)
```
aduckinpants: This function claims to 'Count people' but does not return a count.

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -7,31 +12,198 @@ public static class BalanceStats
...
+                {
+                    listOfPeople.Add(array[0]);
+                }
+            }
+            // Get the list of people
+            string persons = CountPeople(listOfPeople);
+
+            // Get a negative amount back with the sign after the minus sign
+            string amount = "";
+            if (currentHigh < 0) amount = $"-¤{-currentHigh}";
```
aduckinpants: The formatting of the if/else branch is nonstandard and difficult to read.

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -34,7 +37,54 @@ public static class TextTable
...
+            int maxWidth = maxCharacters + (padding * 2);
+            string minus = new string('-', maxWidth);
+            string emptyLine = new string(' ', maxWidth);
+            string emptySpace = new string(' ', padding);
+
+            // TopLine
+            textTable.Append($"+{minus}+\r\n");
+
+            // NewLine if padding
+            for (int i = 0; i < padding; i++)
```
aduckinpants: Fyi sometimes loops like this can be avoided by re-framing things just a little like so:

`textTable.Append(string.Join(Environment.NewLine, Enumerable.Repeat($"|{emptyLine}|", padding)));`

aduckinpants (changes requested):

Src/BootCamp.Chapter/BalanceStats.cs
iamzumie: For currency amount it's better to use decimal

Src/BootCamp.Chapter/BalanceStats.cs
iamzumie: Used a split & then a Trim afterwards

Src/BootCamp.Chapter/BalanceStats.cs
iamzumie: Renamed it to ReturnPeople (variable naming is difficult)

Src/BootCamp.Chapter/TextTable.cs
iamzumie: I had to add \r\n because Environment.NewLine didn't do anything?

Src/BootCamp.Chapter/BalanceStats.cs
```diff
@@ -112,30 +113,30 @@ public static string FindRichestPerson(string[] peopleAndBalances)
...
+            List<String> people = new List<String>();
 
             foreach (var values in peopleAndBalances)
             {
-                var array = values.Split(", ");
+                var array = values.Split(',').Select(a => a.Trim()).ToArray();
 
                 // Parse the last value of each person
-                bool succes = int.TryParse(array[^1], out int parsed);
+                bool parsedAmount = decimal.TryParse(array[^1], out decimal parsed);
```
aduckinpants: This name is better then `succes` but has a new problem! `parsedAmount`, unless you are looking at the exact line it's declared and set on looks like it would be holding a numeric value, an amount of some kind. In fact it holds the boolean result of the parsing attempt. Add an 'is' ir 'was' in there to make it clear that this variable holds the results of some other action. Something like `amountWasParsed` or similar along those lines would go a long way towards improving code clarity.

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -62,10 +63,7 @@ public static string Build(string message, int padding)
             textTable.Append($"+{minus}+\r\n");
 
             // NewLine if padding
-            for (int i = 0; i < padding; i++)
-            {
-                textTable.Append($"|{emptyLine}|\r\n");
-            }
+            textTable.Append(string.Join(Environment.NewLine, Enumerable.Repeat($"|{emptyLine}|\r\n", padding)));
```
aduckinpants, line 66: For what it's worth the loop change I mentioned here was more of a food for thought comment rather then a requested change. I'm not exactly sure how to note that in this UI.

PR #750, opened 2022-12-15
Chapter1/Homework/1And2 (needs review)

mcworkaholic:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,26 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            for (int i = 0; i < 2; i++)
```
Almantask, line 9: Looks good but yuo could have used a function and cal it twice instead of a loop. Would read better.

Almantask (approved):

PR #751, opened 2023-01-05
Add homework 1

kungfuphil:


Almantask (commented):
In tge requirements it asks the same to be done for 2 people. You did only for 1. Tip: use a function not to duplicate code

kungfuphil:
Understood. I've moved that code into a function and called it twice now. Thanks!

Almantask (approved):
You could also rename the function to PromptBmi or similar, but this is good enough. Good job!

PR #754, opened 2023-01-10
Add homework 5

kungfuphil:


Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -9,7 +9,24 @@ public static class ArrayOperations
         /// <param name="array">Input array in a random order.</param>
         public static void Sort(int[] array)
         {
-            // ToDo: implement.
+            if (array == null || array.Length == 0)
```
Almantask: Didn't teach yet, but you can use array?.Length > 0. It doesn both the null check and the comparison

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -29,8 +58,19 @@ public static void Reverse(int[] array)
         /// <returns>A new array with the last element removed. If an array is empty or null, returns input array.</returns>
         public static int[] RemoveLast(int[] array)
```
Almantask, line 59: When you have RemoveAt, can't you reuse it when implementing RemoveFirst/RemoveAt?

Src/BootCamp.Chapter1/ArrayOperations.cs
```diff
@@ -63,8 +137,17 @@ public static int[] RemoveAt(int[] array, int index)
...
+
+            for (int i = 1; i <= newLength - 1; i++)
+            {
+                newArray[i] = array[i - 1];
+            }
+
+            return newArray;
         }
 
         /// <summary>
```
Almantask, line 130: Same with insert. Can you reuse something?

Almantask (changes requested):

kungfuphil:
Thanks for the suggestions!

PR #758, opened 2023-03-04
Chapter1/homework/1 and2

pradoPedroPrado:
Hi there. I made the first homework. Hope there is some reviewers still working on this. Cheers

### DIFF
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,7 +6,30 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            GetPersonData();
+            GetPersonData();
+
+
+            void GetPersonData()
+            {
+                Console.WriteLine("Enter person name:");
+                string name = Console.ReadLine();
+                Console.WriteLine("Enter person surename:");
+                string surename = Console.ReadLine();
+                Console.WriteLine("Enter person age:");
+                int age = Convert.ToInt32(Console.ReadLine());
+                Console.WriteLine("Enter person weight in kg:");
+                double weight = Convert.ToDouble(Console.ReadLine());
+                Console.WriteLine("Enter person height in cm:");
+                double height = Convert.ToDouble(Console.ReadLine());
+                double heightMt = height / 100.0;
+                Console.WriteLine($"{name} {surename} is {age}, weight is {weight} kg and his height is {height} cm.");
+                var BMI = weight / (heightMt * heightMt);
+                Console.WriteLine($"BMI is {BMI:f2}");
+
+            }
+
+
         }
     }
 }
--- a/Src/BootCamp.Chapter/Program.cs
+++ b/Src/BootCamp.Chapter/Program.cs
@@ -6,30 +6,45 @@ namespace BootCamp.Chapter
     {
         static void Main(string[] args)
         {
+            string name;
+            string surename;
+            int age;
+            double weight;
+            double height;
+
             GetPersonData();
-            GetPersonData();
+
+            double BMI;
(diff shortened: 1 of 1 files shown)


Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,45 @@ class Program
     {
         static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            string name;
+            string surename;
+            int age;
+            double weight;
+            double height;
```
Almantask: Prefer to declare a variable just before you use it.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,45 @@ class Program
...
+            double BMI;
+            CalculateBMI(weight, height);
+
+            Console.WriteLine($"{name} {surename} is {age}, weight is {weight} kg and his height is {height} cm.");
+            Console.WriteLine($"BMI is {BMI:f2}");
+
+
+
+
+            void GetPersonData()
```
Almantask: What you have here is a function within a function. It works, but only in the scope of this function. Avoid nested functions (function within function). You should move this outside of the main method

Almantask: For now, I recommend breaking this down into many smaller functions each returning a different thing: name, surename, age, weight, height.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,45 @@ class Program
...
+                surename = Console.ReadLine();
+                Console.WriteLine("Enter person age:");
+                age = Convert.ToInt32(Console.ReadLine());
+                Console.WriteLine("Enter person weight in kg:");
+                weight = Convert.ToDouble(Console.ReadLine());
+                Console.WriteLine("Enter person height in cm:");
+                height = Convert.ToDouble(Console.ReadLine());
+            }
+
+            void CalculateBMI(double weight, double height)
```
Almantask: A function that calculates something should return something (and not be void)

Src/BootCamp.Chapter/Program.cs
```diff
@@ -6,7 +6,45 @@ class Program
...
+
+            GetPersonData();
+
+            double BMI;
+            CalculateBMI(weight, height);
+
+            Console.WriteLine($"{name} {surename} is {age}, weight is {weight} kg and his height is {height} cm.");
+            Console.WriteLine($"BMI is {BMI:f2}");
+
+
```
Almantask, line 29: Too many random spaces

Almantask (changes requested):

pradoPedroPrado:
Thank you so much Kai. 
I made the corrections you suggested. I don't know if you can already see them or if I need to make a new pull request.
Sorry, it's my first time using GitHub.

Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,12 +1,43 @@
...
+            surename = GetPersonData("surename");
+            age = int.Parse(GetPersonData("age"));
+            weight = double.Parse(GetPersonData("weight"));
+            height = double.Parse(GetPersonData("height in cm"));
+            BMI = CalculateBMI(weight, height);
+            Console.WriteLine($"{name} {surename} is {age}, weight is {weight} kg and height is {height} cm.");
+            Console.WriteLine($"BMI is {BMI:f2}");
+        }
+
+        public static string GetPersonData(string text)
```
Almantask, line 30: nit: can be private, because you only use it within this class. The intention is not to use it outside of the class, but rather to just group lines of code in a readable way that you use internally within this function.

Almantask (approved):
Look good enough now. Well done:)

PR #770, opened 2023-03-29
Chapter2/homework/1

pradoPedroPrado:


Src/BootCamp.Chapter/Program.cs
```diff
@@ -1,16 +1,363 @@
...
+            {
+                newArray[i] = items[i];
+            }
+            newArray[^1] = item;
+            return newArray;
+
+        }
+
     }
 }
```
Almantask: Should this belong in Program.cs? Or should it be moved to its own components? 360 lines - is that many lines of code for one file?

Src/BootCamp.Chapter/TextTable.cs
```diff
@@ -0,0 +1,40 @@
...
+         */
+
+        /// <summary>
+        /// Build a table for given message with given padding.
+        /// Padding means how many spaces will a message be wrapped with.
+        /// Table itself is made of: "+-" symbols. 
+        /// </summary>
+        public static string Build(string message, int padding)
+        {
+            return Program.Frame(message, padding);
```
Almantask: This function isn't just for the looks or code to be testable. Things were supposed to be the other way round - implementation to be in here. In high levels (like the Program.cs class) you should only call other things, start the app and have nothing more. In other classes (Like this one) you should implement the meissing functionality.

Almantask (changes requested):

PR #784, opened 2023-06-13
Submit Ch1 HW 7

TMason11095:
- FindHighestBalanceEver- return name and balance(current) of person who had the biggest historic balance.
-FindPersonWithBiggestLoss- return name and loss of a person with a biggest loss (balance change negative).
-FindRichestPerson- return name and current money of the richest person.
-FindMostPoorPerson- return name and current money of the most poor person.


TMason11095:
Added calls to Program.cs to display the newly added function calls.

Added "99980" to PeoplesBalances.Balances for Bill's 2nd balance as it was blank and made the string array not valid.

PR #812, opened 2023-10-14
Homework 1 and 2

Dead-World-Operator:
1. Print all the info based on the example message below:
Tom Jefferson is 19 years old, his weight is 50 kg and his height is 156.5 cm.
2. Calculate and print body-mass index (BMI)
3. Do it for 2 people (repeat the same thing twice)


arbabmalik46 (approved):
you could also while loop. that whenever choice is made to store the data then you can enter data

PR #822, opened 2024-04-17
Chapter1/homework/4 - Completed

ggan-au:
Chapter 1 - Homework 4 Completed. All tests passing.


Src/BootCamp.Chapter/Checks.cs
```diff
@@ -16,26 +16,30 @@ public static class Checks
     {
         public static int PromptInt(string message)
         {
-            // To do: call your implementation. 
-            return 0;
+            // To do: call your implementation.
+            int age = Lesson4.promptInt(message);
```
dishitd, line 20: small suggestion: You can probably return - 
`return Lesson4.promptInt(message)`
This will save you 1 line of code in each of the methods.

